Issue #2, opened 2021-05-20
workflow for releasing library [and schema]


yarikoptic:
Continuing on https://github.com/dandi/dandi-cli/issues/638#issuecomment-845224089 
Since we ended up with a library, it should have its own versioning from schema version.  But it somewhat brings us a full circle to the motivation behind https://github.com/dandi/dandi-cli/issues/638 -- automate and harmonize (have staging etc) schema releases...

observation: **any new schema release should trigger library release** (since it would provide all the pydantic "interface" to that schema), **but not wise versa** (we can release library patches and big changes to API etc without changing anything to schema).

I am trying to wrap my head around what should be our workflow here. I guess we can 

- use `auto` labels etc for regular library releasing (all the `minor`, `major` etc labels)
- schema version change will require manual change of `DANDI_SCHEMA_VERSION`
- we dedicate tags with `schema-{DANDI_SCHEMA_VERSION}` to annotate releases which change the schema
- both client and schema (if current `{DANDI_SCHEMA_VERSION}` was not released) will be released upon merge of the PR marked for `release`:
  - library **can** be released without making a new release of schema
  - schema **cannot** be released without release of the library
  - PR marked with `release` label should get extra test to ensure that `DANDI_SCHEMA_VERSION` schema was not yet released (e.g. by testing for presence of `schema-{DANDI_SCHEMA_VERSION}` tag and/or looking into exported schemas, and fail otherwise.
    - note: `DANDI_SCHEMA_VERSION` might be incremented in some prior PR which was not marked for `release`
    - we need to re-trigger tests if labels change on a PR
    - we need to protect `master` branch to not allow untested/failing PRs (yet to see if feasible given possible codecov "complaints")
- no `release` of library should be allowed if it changes already released schema: extra check to be added to `release` workflow for paranoid @yarikoptic.
WDYT @satra and @jwodder ?

satra:
i agree with observation/assertion.

i think github-actions can trigger a re-check if a label is changed (added/removed). this is how conda-forge does a rerender.

yarikoptic:
Recheck might be insufficient if I just quickly click merge after adding release label. But I guess it also shouldn't hurt , so let's do that too.

yarikoptic:
no better ideas came to mind, so let's proceed with the outlined here plan

jwodder:
@yarikoptic 

> we dedicate tags with schema-{DANDI_SCHEMA_VERSION} to annotate releases which change the schema

Would this be done manually?

[...]

PR #6, opened 2021-05-21
ENH: support migration, validation, schema improvements

satra:
These schema updates are being used both to fix missing elements in the schema (like schemaKey) but also to handle asset metadata being discussed in this PR: https://github.com/dandi/dandi-cli/pull/644

- Fix: missing schemaKey and autogenerate them
- Enh: Migration from older schema
- Enh: Validation of dandiset
- RF: Move datacite support out of models
- RF: move datacite terms into their own namespace
- FIX: relatedResource now supports doi's and URLs
- TST: improve test coverage
- RF: JSON data
- RF: make pydantic validation primary


satra:
@yarikoptic - Equipment is added as an entity that was used in a Session rather than as an Agent. This will get more complicated with self-driving equipment, but for the moment most equipment are passive and therefore have no agency.

satra:
@yarikoptic - this should be good to go for now. how do i create a release?

yarikoptic:
If all works, just add release and minor or major or patch label and click merge. I added labels, you click merge? ;-)

yarikoptic:
The question though - do we want to release or stage more changes into it?

satra:
> do we want to release or stage more changes into it? 

this will help use it in the API server. also this schema change will require schema version updates on both the UI and the API. 

we will likely need to do a mini-migration for some dandiset metadata. actually, let me add a migrate function to this. probably will need refinement over time.

yarikoptic:
I wouldn't bother with migration ATM. We still should be able to reextract and upload (since they don't mutate ATM ;-))

satra:
realized that i wasn't paying attention to the dandi namespace, so had conflicts between roletypes and top level items (Software, EthicsApprovals). also changed Open to OpenAccess and added a dandiset metadata migration function.

for assets we would mint new metadata, but dandiset metadata also needs to be migrated alongside schema change on the server. so after this is merged, we do need to do the updates simultaneously.

wouldn't mind some eyes looking at the changes before we merge.

dandischema/model_types.py
```diff
@@ -422,169 +422,169 @@
             "rdfs:subClassOf": {"@id": "schema:Enumeration"},
         },
         {
-            "@id": "dandi:Author",
+            "@id": "dandirole:Author",
```
yarikoptic, line 425: so it is `dandi:RestrictedAccess` (thus `Access` is folded into value) but `dandirole:Author` and not e.g. `dandi:AuthorRole` ... ideally should be consistent

dandischema/tests/test_datacite.py
```diff
@@ -0,0 +1,277 @@
...
+from jsonschema import Draft6Validator
+import pytest
+import requests
+
+from ..datacite import to_datacite
+from ..models import LicenseType, PublishedDandisetMeta, RoleType
+
+
+def datacite_post(datacite, doi):
+    """ posting the datacite object and checking the status of the requests"""
```
yarikoptic, line 16:
```suggestion
    """Post the datacite object and check the status of the request"""
```
I think imperative form is generally better and that is what we use primarily: it is shorter, `-ing` suggests some extended in time action (content manager? generator? nope...), and I do not see a point in leading space.

dandischema/tests/test_datacite.py
```diff
@@ -0,0 +1,277 @@
...
+    # checking if i'm able to get the url
+    rg = requests.get(url=f"https://api.test.datacite.org/dois/{doi}/activities")
+    rg.raise_for_status()
+
+    # cleaning url
+    _clean_doi(doi)
+
+
+def _clean_doi(doi):
+    """removing doi, ignoring the status code"""
```
yarikoptic, line 39:
```suggestion
    """Remove doi. Status code is ignored"""
```

dandischema/tests/test_datacite.py
```diff
@@ -0,0 +1,277 @@
...
+    sr.raise_for_status()
+    schema = sr.json()
+    return schema
+
+
+def _basic_publishmeta(dandi_id, version="v.0", prefix="10.80507"):
+    """
+    adding basic info required by PublishedDandisetMeta in addition to
+    fields required by DandisetMeta
+    """
```
yarikoptic, line 61:
```suggestion
    """Return extra metadata required by PublishedDandisetMeta 
    
    Returned fields are additional to fields required by DandisetMeta
    """
```

dandischema/metadata.py
```diff
@@ -90,3 +95,30 @@ def validate_asset_json(data, schema_dir):
...
+    if schema_key == "Dandiset":
+        schema_key = "DandisetMeta"
+    elif schema_key == "Asset":
+        schema_key = "AssetMeta"
+    schema_version = schema_version or obj.get("schemaVersion") or DANDI_SCHEMA_VERSION
+    klass = getattr(models, schema_key)
+    klass(**obj)
+
+
+def migrate_001(obj):
```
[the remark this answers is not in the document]
yarikoptic: also how is this one interfaced, or nohow yet? altogether, ideally, I would have kept it for another PR which would have e.g. added support for migrations and allow to validate older schemas upon automagic migration.

dandischema/tests/test_metadata.py
```diff
@@ -0,0 +1,238 @@
...
+
+@pytest.fixture(scope="module")
+def schema_dir(tmp_path_factory):
+    return publish_model_schemata(tmp_path_factory.mktemp("schema_dir"))
+
+
+def test_asset(schema_dir):
+    with (METADATA_DIR / "asset_001.json").open() as fp:
+        data_as_dict = json.load(fp)
+    data_as_dict["schemaVersion"] = DANDI_SCHEMA_VERSION
```
yarikoptic, line 28: with migrations added, I think such overwrites of schemaVersion should not be needed and not desired.  But I guess for now it is ok, but let's make that explicit

```suggestion
    # overload (here and below) schemaVersion until we support automagic schema migrations
    # under assumption that the 0.3.2 schema would be forward compatible
    data_as_dict["schemaVersion"] = DANDI_SCHEMA_VERSION
```

yarikoptic (commented):
Well, can't make verdict one way or another ATM, but suggested some changes and asked about clarification for `dandi:` -> `dandirole:` necessity .

Migrations -- might need a more thorough thinking through on how/if we could support those.

dandischema/metadata.py
```diff
@@ -3,17 +3,19 @@
...
 
 
 def generate_context():
     import pydantic
 
-    fields = {
+    field_preamble = {
         "@version": 1.1,
         "dandi": "http://schema.dandiarchive.org/",
+        "dandirole": "http://schema.dandiarchive.org/roles/",
```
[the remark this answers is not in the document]
satra: to make the uris address different concepts. this is the same thing as: `dandi:roles/Software` but i was using the prefix to parse certain things.

dandischema/metadata.py
```diff
@@ -90,3 +95,30 @@ def validate_asset_json(data, schema_dir):
...
+    elif schema_key == "Asset":
+        schema_key = "AssetMeta"
+    schema_version = schema_version or obj.get("schemaVersion") or DANDI_SCHEMA_VERSION
+    klass = getattr(models, schema_key)
+    klass(**obj)
+
+
+def migrate_001(obj):
+    schema_version = obj.get("schemaVersion")
+    if schema_version == DANDI_SCHEMA_VERSION:
```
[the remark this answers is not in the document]
satra: this should work as `DANDI_SCHEMA_VERSION` is always the latest. so if something conforms to that, don't make any changes. it's a trivial check, but i think it does the right thing.

dandischema/model_types.py
satra: let me think about this. it is indeed our namespace so we can control.

dandischema/metadata.py
```diff
@@ -90,3 +95,50 @@ def validate_asset_json(data, schema_dir):
     with Path(schema_dir, "asset.json").open() as fp:
         schema = json.load(fp)
     jsonschema.validate(data, schema)
+
+
+def validate(obj, schema_version=None, schema_key=None):
```
jwodder, line 100: This docstring isn't terminated.

satra: yup - will be fixed in the update. got sidetracked on improving models.

dandischema/model_types.py
satra: i moved everything for datacite to a `dcite` prefix. unfortunately they don't have a url so i just used a dandi namespace for it.

dandischema/metadata.py
```diff
@@ -90,3 +95,30 @@ def validate_asset_json(data, schema_dir):
...
+    schema_key = schema_key or obj.get("schemaKey")
+    if schema_key is None:
+        raise ValueError("Provided object has no known schemaKey")
+    if schema_key == "Dandiset":
+        schema_key = "DandisetMeta"
+    elif schema_key == "Asset":
+        schema_key = "AssetMeta"
```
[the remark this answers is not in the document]
satra: done

satra:
@yarikoptic - this should be ready now.

satra:
@jwodder, @yarikoptic  - i created a new access token for our bot user for this repo and set the secret with it.

yarikoptic:
lost in comments:  I did add `release` label awhile back, but then I thought I saw a comment from you @satra that we might not want to release just right away so I disabled releasing schemata with #11 in favor of @jwodder working out a proper release support according to #2 .  So -- what is your desire -- to have it released asap upon merge?

.github/workflows/test.yml
```diff
@@ -3,12 +3,13 @@ name: Tests
 on:
   push:
   pull_request:
-  schedule:
-    - cron: '0 6 * * *'
```
yarikoptic, line 7: why?   It is good to have things tested regularly to make sure that it still all works eg. when new pydantic comes out etc.
 
I haven't spotted any commit message summary among 22 (so didn't go one by one) to see if more details would be given.

NB filing as a separate comment to not await for my full review

satra: it seems to be solving a different problem. it would be good to use less electrons when possible. once we get into lazy mode with schema, we can turn cron job back on. or if we did this once a week or month. every 6 hours is total overkill.

satra:
yes this releases schema 0.4.0 so should happen on release of this library. so merge -> release library -> release schema. the last two go hand in hand.

.github/workflows/test.yml
jwodder: That's not "every 6 hours"; it's "once a day at 6 AM".

satra: that is true.

dandischema/models.py
```diff
@@ -1088,16 +905,18 @@ class BareAssetMeta(CommonModel):
...
+
     dateModified: Optional[datetime] = Field(
         nskey="schema", title="Asset (file or metadata) modification date and time"
     )
     blobDateModified: Optional[datetime] = Field(
         nskey="dandi", title="Asset file modification date and time"
     )
 
-    path: str = Field(None, nskey="dandi")
-
```
yarikoptic, line 1100: is order of specification here reflects itself in metadata records dumps?  then we might as well harmonize as to place all mandatory first (may be sort them alphabetically) and then all optional next, also alphabetically.

Special cases -- read-only identifier first?  the rest of read-only fields last.

Might be feasible to do automagically in that metaclass I proposed above. if done prior feeding the list of attributes to super class (so then `schemaKey` would be last, as it should ;)?)

dandischema/models.py
```diff
@@ -1133,42 +952,57 @@ class BareAssetMeta(CommonModel):
         "nskey": "dandi",
     }
 
+    @validator("digest")
+    def digest_etag(cls, values):
+        print(values)
```
yarikoptic, line 957:
```suggestion
```
I guess a left over

dandischema/models.py
```diff
@@ -1133,42 +952,57 @@ class BareAssetMeta(CommonModel):
...
+        try:
+            if len(values[DigestType.dandi_etag]) != 32:
+                raise ValueError
+        except (KeyError, ValueError):
+            raise ValueError("Digest must have an appropriate dandi-etag value.")
```
yarikoptic, line 962:
```suggestion
            raise ValueError(f"Digest must have an appropriate dandi-etag value. Got {values[DigestType.dandi_etag]}")
```
what can I say -- I am a friend of mortals ;-) (note: might need linting)

I assume that the `@validator` would wrap those nicely into `pydantic.ValidationError`s?

dandischema/models.py
```diff
@@ -1133,42 +952,57 @@ class BareAssetMeta(CommonModel):
...
-class PublishedAssetMeta(AssetMeta, Publishable):
-    pass
+class PublishedAsset(Asset, Publishable):
+    @validator("digest")
+    def digest_bothhashes(cls, values):
+        try:
+            if len(values[DigestType.dandi_etag] + values[DigestType.sha2_256]) != 96:
+                raise ValueError
+        except (KeyError, ValueError):
+            raise ValueError("Digest must have both dandi-etag and sha2-256.")
```
yarikoptic, line 1004:
```suggestion
            raise ValueError("Digest must have both valid dandi-etag and sha2-256.")
```
I would have loved to stick actual values but decided not to bother yet - would need assignment to temp vars above

dandischema/models.py
```diff
@@ -315,6 +96,7 @@ def encode(self, o):
 
 class DandiBaseModel(BaseModel):
     id: Optional[str] = Field(description="Uniform resource identifier", readOnly=True)
+    schemaKey: Literal["DandiBaseModel"] = Field("DandiBaseModel", readOnly=True)
```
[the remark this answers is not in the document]
yarikoptic: I will submit this review now and quickly try to furnish a PR against this PR branch to introduce such metaclass to see if feasible.

yarikoptic (changes requested):
- minor suggestions or questions (e.g. re `Type` classes) here and there to strengthen code and unify naming
- `metaclass` idea: to avoid manual schemaKey entry causing duplication etc.  Will furnish a PR quickly to see if it is worth it
- etc ;-)

.github/workflows/test.yml
yarikoptic: pushed bringing back daily run - the change was OOS for this PR ;)

dandischema/datacite.py
```diff
@@ -0,0 +1,190 @@
...
+            "name": contr_el.name,
+            "contributorName": contr_el.name,
+            "schemeURI": "orcid.org",
+        }
+        if isinstance(contr_el, Person):
+            contr_dict["nameType"] = "Personal"
+            if len(contr_el.name.split(",")) == 2:
+                contr_dict["familyName"], contr_dict["givenName"] = contr_el.name.split(
+                    ","
+                )
```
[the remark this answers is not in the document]
satra: i can't remember if i put a regex in the name field for person, but will check.

dandischema/datacite.py
```diff
@@ -0,0 +1,190 @@
...
+                continue
+
+        if getattr(contr_el, "roleName"):
+            contr_all = [
+                el.name for el in contr_el.roleName if el.name in DATACITE_CONTRTYPE
+            ]
+            if contr_all:
+                contr_dict["contributorType"] = contr_all[0]
+            else:
+                contr_dict["contributorType"] = "Other"
```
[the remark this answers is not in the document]
satra: there are many things to do, also not all roles we have are in datacite yet. they will eventually get there, but not yet. so i don't want to cripple our system to work just for datacite.

dandischema/datacite.py
```diff
@@ -0,0 +1,190 @@
...
+        creators[0]["creatorName"] = creators[0].pop("contributorName")
+        creators[0].pop("contributorType")
+
+    attributes["contributors"] = contributors
+    attributes["creators"] = creators
+
+    if getattr(meta, "relatedResource"):
+        attributes["relatedIdentifiers"] = []
+        for rel_el in meta.relatedResource:
+            ident = rel_el.identifier.split(":")
```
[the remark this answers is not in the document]
satra: this is true. that split won't work for the pattern we have.

dandischema/metadata.py
```diff
@@ -3,17 +3,20 @@
...
 
 
 def generate_context():
     import pydantic
 
-    fields = {
+    field_preamble = {
         "@version": 1.1,
         "dandi": "http://schema.dandiarchive.org/",
+        "dcite": "http://schema.dandiarchive.org/datacite/",
```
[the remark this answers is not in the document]
satra: thank you.

dandischema/metadata.py
```diff
@@ -3,17 +3,20 @@
...
         "dandiasset": "http://iri.dandiarchive.org/",
```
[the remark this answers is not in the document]
satra: yes, the schema.dandiarchive.org are for terms/classes/properties not identifiers. i will change to this last form.

dandischema/metadata.py
```diff
@@ -61,32 +65,89 @@ def generate_context():
...
 
-def validate_dandiset_json(data, schema_dir):
+def validate_dandiset_json(data: dict, schema_dir: str) -> None:
     with Path(schema_dir, "dandiset.json").open() as fp:
         schema = json.load(fp)
     jsonschema.validate(data, schema)
 
 
-def validate_asset_json(data, schema_dir):
+def validate_asset_json(data: dict, schema_dir: str) -> None:
```
satra, line 95:
```suggestion
def _validate_asset_json(data: dict, schema_dir: str) -> None:
```

dandischema/metadata.py
```diff
@@ -61,32 +65,89 @@ def generate_context():
...
+    schema_version = schema_version or obj.get("schemaVersion") or DANDI_SCHEMA_VERSION
+    klass = getattr(models, schema_key)
+    klass(**obj)
+
+
+def migrate(obj, to_version=DANDI_SCHEMA_VERSION):
+    if to_version not in ALLOWED_TARGET_SCHEMAS:
+        raise ValueError(f"Current target schemas: {ALLOWED_TARGET_SCHEMAS}.")
+    schema_version = obj.get("schemaVersion")
+    if schema_version == DANDI_SCHEMA_VERSION:
```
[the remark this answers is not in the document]
satra: yes

dandischema/metadata.py
```diff
@@ -61,32 +65,89 @@ def generate_context():
...
+        for contrib in obj["contributor"]:
+            contrib["roleName"] = [
+                val.replace("dandi:", "dcite:") for val in contrib["roleName"]
+            ]
+        for contrib in obj["relatedResource"]:
+            contrib["relation"] = contrib["relation"].replace("dandi:", "dcite:")
+        for access in obj["access"]:
+            access["status"] = "dandi:OpenAccess"
+    obj["schemaVersion"] = to_version
+    return obj
```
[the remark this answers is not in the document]
satra: because validation may still fail. the only time validation passes will be when required fields are in place. migrate can be called on non-publishable metadata.

dandischema/models.py
satra: i wouldn't rely on ordering here i just moved it to be visually closer to mandatory fields

dandischema/models.py
```diff
@@ -1052,13 +869,13 @@ def check_data(cls, values):
...
-    citation: TempOptional[str] = Field(readOnly=True, nskey="schema")
+    citation: str = Field(readOnly=True, nskey="schema")
 
     # From assets
-    assetsSummary: TempOptional[AssetsSummary] = Field(readOnly=True, nskey="dandi")
+    assetsSummary: AssetsSummary = Field(readOnly=True, nskey="dandi")
 
     # From server (requested by users even for drafts)
-    manifestLocation: TempOptional[List[HttpUrl]] = Field(readOnly=True, nskey="dandi")
+    manifestLocation: List[HttpUrl] = Field(readOnly=True, nskey="dandi")
```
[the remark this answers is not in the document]
satra: metadata saved does not have to be valid until publish. however, while `assetsSummary` can change and should be updated based on asset related triggers, `manifestLocation` does not change once a dandiset is created and/or published. so it will be considered invalid till those fields are populated. but that's a state a dandiset metadata can be in without much effort (for example no ContactPerson entered or removed).

dandischema/metadata.py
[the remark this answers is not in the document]
yarikoptic: filed dedicated https://github.com/dandi/dandischema/issues/14 so we do not forget

dandischema/models.py
```diff
@@ -380,10 +176,13 @@ def schema_extra(schema: Dict[str, Any], model) -> None:
                 # In pydantic 1.8+ all Literals are mapped on to enum
                 # This presently breaks the schema editor UI. Revert
                 # to const when generating the schema.
+                # Note: this no longer happens with custom metaclass
                 if prop == "schemaKey":
-                    if len(value["enum"]) == 1:
+                    if "enum" in value and len(value["enum"]) == 1:
                         value["const"] = value["enum"][0]
```
yarikoptic, line 182: absent coverage here confirms that it doesn't happen across CIs either after moving to custom metaclass, hm... good or bad? ;)

dandischema/models.py
```diff
@@ -425,47 +208,47 @@ class TypeModel(DandiBaseModel):
...
 
-class SpeciesType(TypeModel):
+class SpeciesType(BaseType):
     """Identifier for species of the sample"""
 
     schemaKey: Literal["SpeciesType"] = Field("SpeciesType", readOnly=True)
 
 
-class Disorder(TypeModel):
+class Disorder(BaseType):
```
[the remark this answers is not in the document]
yarikoptic: filed dedicated https://github.com/dandi/dandischema/issues/15

dandischema/datacite.py
```diff
@@ -0,0 +1,192 @@
...
+    ]
+
+    attributes["doi"] = meta.doi
+    attributes["titles"] = [{"title": meta.name}]
+    attributes["descriptions"] = [
+        {"description": meta.description, "descriptionType": "Abstract"}
+    ]
+    attributes["publisher"] = "DANDI Archive"
+    attributes["publicationYear"] = str(meta.datePublished.year)
+    # not sure about it dandi-api had "resourceTypeGeneral": "NWB"
```
yarikoptic, line 81: note: lost track of comments... I was wondering there about where we could get specific  data standards -- we would have them in `AssetsSummary. dataStandard: Optional[List[StandardsType]] `, but not sure where we would stick them yet in datacite

dandischema/datacite.py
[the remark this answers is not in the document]
yarikoptic: filed dedicated https://github.com/dandi/dandischema/issues/16

yarikoptic (approved):
To not hold progress -- filed dedicated issues on comments which weren't addressed yet as part of this PR. I think we would be in greater shape with this magnificent change (thanks @satra for doing the exciting metadata schema stuff!) so let's proceed. Personally I would have preferred to massage it more to get "well done" given the number of components reliant on schema, but of cause we can iterate.

satra:
closes #14, #16

PR #8, opened 2021-05-25
Include CHANGELOG.md and tox.ini in sdists

jwodder:


yarikoptic:
actually -- nevermind, I will submit a crude PR to suspend temporarily autopublishing jsonschemas in https://github.com/dandi/dandischema/pull/8 following #2 discussion

PR #13, opened 2021-05-25
Fix/schemakey metaclass

yarikoptic:
Please see individual commit messages for some extra metadata.

Supplement to #6

May be we should drop that `enum` special handling altogether and then assignment of that `const`?

edit: FWIW tried to sort `dct` in metaclass for no effect on produced jsonschema


yarikoptic:
argh, we need precommit here!

satra:
we are abusing schemaKey a bit. while we are making a distinction in the code between publishedasset and publisheddandiset, they are both of type dandiset and asset. i.e. the publishing metadata does not change the type.

PR #19, opened 2021-05-26
Prefix library versions with `v`

jwodder:
See https://github.com/dandi/dandischema/issues/2#issuecomment-848900472.

In addition to this PR, the tag for at least the most recent library release (0.2.0) must be renamed to have a `v` prefix, and the corresponding GitHub release must be updated to match.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -70,7 +70,7 @@ VCS = git
 style = pep440
 versionfile_source = dandischema/_version.py
 versionfile_build = dandischema/_version.py
-tag_prefix =
+tag_prefix = v
 parentdir_prefix =
 
 [codespell]
--- a/.autorc
+++ b/.autorc
@@ -2,7 +2,7 @@
     "onlyPublishWithReleaseLabel": true,
     "baseBranch": "master",
     "author": "auto <auto@nil>",
-    "noVersionPrefix": true,
+    "noVersionPrefix": false,
     "plugins": [
         "git-tag",
         [
--- a/dandischema/_version.py
+++ b/dandischema/_version.py
@@ -42,9 +42,9 @@ def get_config():
     cfg = VersioneerConfig()
     cfg.VCS = "git"
     cfg.style = "pep440"
-    cfg.tag_prefix = ""
+    cfg.tag_prefix = "v"
     cfg.parentdir_prefix = ""
-    cfg.versionfile_source = "nibabel/_version.py"
+    cfg.versionfile_source = "dandischema/_version.py"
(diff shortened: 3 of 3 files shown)


setup.cfg
```diff
@@ -70,7 +70,7 @@ VCS = git
 style = pep440
 versionfile_source = dandischema/_version.py
 versionfile_build = dandischema/_version.py
-tag_prefix =
+tag_prefix = v
```
yarikoptic, line 73: Try
```suggestion
tag_prefix = [0-9]
```

yarikoptic: If doesnt work, make it 0 for now

jwodder: That doesn't work, likely due to:

https://github.com/dandi/dandischema/blob/4d9bfbc7943f5607480dee84a6556393cffd20f3/versioneer.py#L630

jwodder: @yarikoptic Setting the prefix to `0` causes the leading `0` to be stripped, resulting in an invalid version string of ".2.0".

yarikoptic: Oh, no joy... grr...
ok -- let's do a workaround within workflow -- remove all tags which aren't starting with `[0-9]+\.` upon checkout ;-) and I will ask you to fixup versioneer so we could avoid that going forward.

jwodder: I really think it'd be easier to just go with this PR and rename the tags.

yarikoptic: I dislike all the `v` prefixing in general since taints `git describe` to not be semver right away. Well -- here we would get that even with versioneer fix since we have `schema-` tags, but at least I can `--match` quickly.

But I also dislike  my proposed workaround since it would nohow solve the problem locally.

Awaiting for fix in versioneer is "suboptimal" as well, but here is my thinking "how": https://github.com/python-versioneer/python-versioneer/issues/255 .

Another workaround would be to patch versioneer.py we have to make it work as we like for now at least:  I think the least intrusive would be to keep no prefix, but hardcode `--exclude '[^0-9]*'` or alike to `git-describe` call.  Then whenever versioneer supports it , we would just add `tag_exclude` and happily upgrade our copy of versioneer.  WDYT @jwodder ?

jwodder: Patching our local versioneer to add `--exclude` seems like a workable solution.

yarikoptic: great, please proceed that way then; here or in #17 so it could turn that CI green ;-)

jwodder:
Closing in favor of #20.

PR #20, opened 2021-05-26
Patch versioneer to exclude tags that don't start with a digit

jwodder:


yarikoptic:
wow -- 3 places!? didn't know... let's proceed. Thanks!

PR #24, opened 2021-05-28
fix: dandi-etag carries extra component beyond md5

satra:
The check for digest assumed an md5sum. however, the digest has an extra component and are of the form: `<md5sum>-<chunks>`

### DIFF
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -937,7 +937,8 @@ class BareAsset(CommonModel):
     @validator("digest")
     def digest_etag(cls, values):
         try:
-            if len(values[DigestType.dandi_etag]) != 32:
+            digest = values[DigestType.dandi_etag]
+            if "-" not in digest or len(digest.split("-")[0]) != 32:
                 raise ValueError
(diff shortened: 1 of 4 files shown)


dandischema/models.py
```diff
@@ -988,10 +989,14 @@ class PublishedAsset(Asset, Publishable):
     @validator("digest")
     def digest_bothhashes(cls, values):
         try:
-            if len(values[DigestType.dandi_etag] + values[DigestType.sha2_256]) != 96:
-                raise ValueError
-        except (KeyError, ValueError):
-            raise ValueError("Digest must have both valid dandi-etag and sha2-256.")
+            digest = values[DigestType.dandi_etag]
+            if "-" not in digest or len(digest.split("-")[0]) != 32:
```
yarikoptic, line 993: I think 
```suggestion
            if len(digest.split("-")[0]) != 32:
```
would be sufficient but ok either way

yarikoptic:
Left minor non-consequential comment which might be wrong, so let's proceed

PR #25, opened 2021-05-28
Fix release workflow

jwodder:


yarikoptic:
FWIW -- removed `release` from #24 to avoid any side effects. will merge whenever tests turn green. Thank you @jwodder

yarikoptic:
green enough! ;)

PR #28, opened 2021-06-08
fix: remove readOnly from constants

satra:
a fix to make us compatible with the vjsf component for meditor.


satra:
@yarikoptic - this does not change the models and only the schema. should we overwrite the schema or increase schema as well? the only thing this impacts is the meditor component.

satra:
bumped schema version anyway.

waxlamp:
> bumped schema version anyway.

I believe the patch level bump was appropriate and required in this case.

yarikoptic (approved):
I am glad to have similar minded zealots of versioning on the team @waxlamp ! ;-)

In general approve, but shouldn't we have just patch label on library then, instead of minor?

satra:
yes, should be patch.

PR #30, opened 2021-06-09
fix: update migration code to account for missing fields

satra:
I just ran through all the dandiset metadata and this code wasn't taking badly formed metadata into account.

### DIFF
--- a/dandischema/metadata.py
+++ b/dandischema/metadata.py
@@ -76,15 +76,21 @@ def publish_model_schemata(releasedir: str) -> Path:
     version = models.get_schema_version()
     vdir = Path(releasedir, version)
     vdir.mkdir(exist_ok=True, parents=True)
-    (vdir / "dandiset.json").write_text(models.Dandiset.schema_json(indent=2))
-    (vdir / "asset.json").write_text(models.Asset.schema_json(indent=2))
+    (vdir / "dandiset.json").write_text(
+        models.Dandiset.schema_json(indent=2).rstrip("\n") + "\n"
+    )
+    (vdir / "asset.json").write_text(
+        models.Asset.schema_json(indent=2).rstrip("\n") + "\n"
+    )
     (vdir / "published-dandiset.json").write_text(
-        models.PublishedDandiset.schema_json(indent=2)
+        models.PublishedDandiset.schema_json(indent=2).rstrip("\n") + "\n"
     )
     (vdir / "published-asset.json").write_text(
-        models.PublishedAsset.schema_json(indent=2)
+        models.PublishedAsset.schema_json(indent=2).rstrip("\n") + "\n"
     )
-    (vdir / "context.json").write_text(json.dumps(generate_context(), indent=2))
+    (vdir / "context.json").write_text(
+        json.dumps(generate_context(), indent=2).rstrip("\n") + "\n"
+    )
     return vdir
 
 
--- a/dandischema/metadata.py
+++ b/dandischema/metadata.py
@@ -165,11 +165,10 @@ def migrate(
         if not id.startswith("DANDI:"):
             obj["id"] = f'DANDI:{obj["id"]}'
         for contrib in obj.get("contributor", []):
-            if not contrib.get("roleName"):
-                continue
-            contrib["roleName"] = [
-                val.replace("dandi:", "dcite:") for val in contrib["roleName"]
-            ]
+            if contrib.get("roleName"):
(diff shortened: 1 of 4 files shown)


satra:
@yarikoptic -  a quick review of this would be much appreciated.

regarding adding newlines. they were added to the schema from the schema precommit action when i pushed manually. i'm just adding that here as a default mechanism when we write out the schemas.

dandischema/metadata.py
```diff
@@ -76,15 +76,21 @@ def publish_model_schemata(releasedir: str) -> Path:
     version = models.get_schema_version()
     vdir = Path(releasedir, version)
     vdir.mkdir(exist_ok=True, parents=True)
-    (vdir / "dandiset.json").write_text(models.Dandiset.schema_json(indent=2))
-    (vdir / "asset.json").write_text(models.Asset.schema_json(indent=2))
+    (vdir / "dandiset.json").write_text(
+        models.Dandiset.schema_json(indent=2).rstrip("\n") + "\n"
```
yarikoptic, line 80: absent trailing new line was an issue?

we don't care about windows here, or python would magically convert line breaks to windows ones?

I would have probably came up with a helper `ensure_trailing_space` or alike to avoid duplication.

dandischema/metadata.py
```diff
@@ -158,16 +164,20 @@ def migrate(
         id = obj.get("id")
         if not id.startswith("DANDI:"):
             obj["id"] = f'DANDI:{obj["id"]}'
-        for contrib in obj["contributor"]:
+        for contrib in obj.get("contributor", []):
+            if not contrib.get("roleName"):
+                continue
             contrib["roleName"] = [
                 val.replace("dandi:", "dcite:") for val in contrib["roleName"]
             ]
```
yarikoptic, line 172:
```suggestion
            if contrib.get("roleName"):
                contrib["roleName"] = [
                    val.replace("dandi:", "dcite:") for val in contrib["roleName"]
                ]
```
since otherwise we would not add schemaKey for affiliation below for contribs without role

dandischema/metadata.py
```diff
@@ -158,16 +164,20 @@ def migrate(
...
             for affiliation in contrib.get("affiliation", []):
                 affiliation["schemaKey"] = "Affiliation"
-        for contrib in obj["relatedResource"]:
+        for contrib in obj.get("relatedResource", []):
             contrib["relation"] = contrib["relation"].replace("dandi:", "dcite:")
-        for access in obj["access"]:
+        for access in obj.get("access", []):
             access["status"] = "dandi:OpenAccess"
+        else:
+            obj["access"] = [{"status": "dandi:OpenAccess"}]
```
yarikoptic, line 180: this Python pattern of `for ... else` was always confusing to me since it is pretty much `for ... if_I_did_not_break` and not  `for ... if_I_had_nothing_to_loop_through`, and I am afraid you might have been confused by it as well since 
```
$> cat /tmp/1.py   
for i in [1]:
   print(i)
else:
   print("or else")

$> python3 /tmp/1.py
1
or else
```
so you would end up just overwriting what you looped through first.  So you do need an explicit `if` for absent `access` instead.

satra: indeed. thanks for that. i don't use it much, but i guess when i use it, i use it incorrectly!

dandischema/metadata.py
yarikoptic: or better, as I have suggested before, I think we should overload `schema_json` and centralize all our tuneups (indent and trailing line) in there

satra: python magically converts or rather both `open`/`write_text` does as far as i could tell.

satra: i'm still a little hesitant on overwriting `schema_json` as then we would need to ensure that everything we have is derived from a common `BaseModel` that just overrides `schema_json`. doable, but hesitant.

satra:
can i merge this now?

dandischema/metadata.py
yarikoptic: isn't everything relevant here is a subclass of `DandiBaseModel` or otherwise we would have other issues (since `DandiBaseModel` does already to_dictrepr and unvalidated and we would fail if it didn't?).  Anyways -- could be harmonized after this PR.  But the question just keeps coming, so we better deal with it.... I guess next time ;)

yarikoptic:
Sure ;-) Let's see how release goes (no new release of schema but new release of library)

PR #33, opened 2021-06-10
Allow location fields to have localhost URLs

dchiquito:
Development/test environments use an object store that is generally
served at `localhost:9000`. URLs that point to objects in this store
obviously have the form `http://localhost:9000/{...}`. The pydantic
field `HttpUrl` requires a TLD, which excludes `localhost`. Instead, we
should use `AnyHttpUrl` which allows `localhost`.

### DIFF
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -5,7 +5,16 @@ import json
 import sys
 from typing import Any, Dict, List, Optional, Type, Union
 
-from pydantic import UUID4, BaseModel, ByteSize, EmailStr, Field, HttpUrl, validator
+from pydantic import (
+    UUID4,
+    AnyHttpUrl,
+    BaseModel,
+    ByteSize,
+    EmailStr,
+    Field,
+    HttpUrl,
+    validator,
+)
 from pydantic.main import ModelMetaclass
 
 from .consts import DANDI_SCHEMA_VERSION
@@ -860,7 +869,7 @@ class Dandiset(CommonModel):
     assetsSummary: AssetsSummary = Field(readOnly=True, nskey="dandi")
 
     # From server (requested by users even for drafts)
-    manifestLocation: List[HttpUrl] = Field(readOnly=True, min_items=1, nskey="dandi")
+    manifestLocation: List[AnyHttpUrl] = Field(readOnly=True, min_items=1, nskey="dandi")
 
     version: str = Field(readOnly=True, nskey="schema")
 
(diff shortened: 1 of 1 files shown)


satra:
@dchiquito - so this seems like a dev only hack rather than what should be in the schema, right? so perhaps we can use an environment variable and make HttpUrl map to AnyHttpUrl, as the official validation should not allow a non TLD ?

dchiquito:
My thought was that all of these fields are calculated by the API server, so it would be acceptable for these fields to accept any possible calculated value from any possible deployment. Providing a separate schema that is only used for dev/testing deployments sounds very difficult and much more complicated.

satra:
all this would take is something like this:

```python
if os.environ["DEV_ENV"]:
    HttpUrl = AnyHttpUrl
```

this is only a pydantic dataclass issue as in the json schema there is no distinction of the URI types.

dchiquito:
Changing the result of importing a module based on environment variables sounds to me like it would cause a lot of headaches down the line, but if that is your preferred solution I can do that.

dandischema/models.py
```diff
@@ -25,6 +35,12 @@
 else:
     from typing import Literal
 
+# Local or test deployments of dandi-api will insert URLs into the schema that refer to the domain
+# localhost, which is not a valid TLD. To make the metadata valid in those contexts, setting this
+# environment variable will use a less restrictive pydantic field that allows localhost.
+if os.environ["DANDI_ALLOW_LOCALHOST_URLS"]:
```
satra:
```suggestion
if os.environ.get("DANDI_ALLOW_LOCALHOST_URLS"):
```

satra:
@dchiquito - would you like a release to go alongside this? in which case please label this with release, patch.

dandischema/models.py
```diff
@@ -25,6 +35,12 @@
 else:
     from typing import Literal
 
+# Local or test deployments of dandi-api will insert URLs into the schema that refer to the domain
+# localhost, which is not a valid TLD. To make the metadata valid in those contexts, setting this
+# environment variable will use a less restrictive pydantic field that allows localhost.
+if "DANDI_ALLOW_LOCALHOST_URLS" in os.environ:
```
yarikoptic, line 41:
```suggestion
if "DANDISCHEMA_ALLOW_LOCALHOST_URLS" in os.environ:
```
or
```suggestion
if "DANDI_SCHEMA_ALLOW_LOCALHOST_URLS" in os.environ:
```
to not conflate with `dandi-cli` variables, or shouldn't bother?

I guess if we hit more of similar tune ups needed, the whole schema might indeed need to become "configurable".  I agree with @dchiquito that import-time tune up like this makes it rather suboptimal since e.g. we wouldn't be able to do validation against "proper" schema for the assets which aren't coming from localhost while some others would be for localhost one.

dchiquito: We also use `DANDI_` variables for dandi-api, I think it's fine to have that as a global prefix for everything in the DANDI ecosystem.

yarikoptic: sure, but `DANDI_ALLOW_LOCALHOST_URLS` sounds like something we might end up adding at some point to another component which also uses `dandischema` and then it might be an undesired effect.  anyways -- now it is 3 brains which will remember about it and warn each other if see someone adding it somewhere ;)  so no strong desires on this one

PR #34, opened 2021-06-16
enh: implement metadata aggregator for asset summary

satra:
This PR implements the computation of asset summary for a dandiset to this library.

- slight changes to regexes required pattern updates (so new schema will be released)

addresses part 1 of https://github.com/dandi/dandi-api/issues/337

will file a PR on API to use this.


yarikoptic:
Let's proceed!

PR #40, opened 2021-06-17
fix: allow migration for versions greater than 0.3.2

satra:
Allow migrating from versions greater 0.3.1


dchiquito (commented):
It seems like we'd only like to run most of that block of code when migrating from < 0.3.2. What about adding this instead:
```
elif version2tuple(schema_version) < version2tuple(DANDI_SCHEMA_VERSION):
    obj["schemaVersion"] = to_version
```

dchiquito (approved):

PR #41, opened 2021-06-18
Fix: Make the library more robust for publishing

satra:
1. updates datacite constructor to ensure
  - contributorType
  - support validation against jsonschema from datacite (this may not be authoritative and should be used with care)
2. ensure that round tripping of schemaKey creates the same object or raises a validation error

closes https://github.com/dandi/dandi-api/issues/347

### DIFF
--- a/dandischema/datacite.py
+++ b/dandischema/datacite.py
@@ -90,7 +90,7 @@ def to_datacite(
         "resourceTypeGeneral": "Dataset",
     }
     # meta has also attribute url, but it often empty
-    attributes["url"] = str(meta.url)
+    attributes["url"] = str(meta.url or "")
     # assuming that all licenses are from SPDX?
     attributes["rightsList"] = [
         {
--- a/dandischema/datacite.py
+++ b/dandischema/datacite.py
@@ -2,6 +2,9 @@ from copy import deepcopy
 import re
 import typing as ty
 
+from jsonschema import Draft6Validator
+import requests
+
 from .models import NAME_PATTERN, Organization, Person, PublishedDandiset, RoleType
 
 DATACITE_CONTRTYPE = {
@@ -53,7 +56,9 @@ DATACITE_IDENTYPE = {
 DATACITE_MAP = {el.lower(): el for el in DATACITE_IDENTYPE}
 
 
-def to_datacite(meta: ty.Union[dict, PublishedDandiset]) -> dict:
+def to_datacite(
+    meta: ty.Union[dict, PublishedDandiset], validate: bool = False
+) -> dict:
     """Convert published Dandiset metadata to Datacite"""
     if not isinstance(meta, PublishedDandiset):
         meta = PublishedDandiset(**meta)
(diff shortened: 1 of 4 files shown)


dandischema/datacite.py
```diff
@@ -85,7 +90,7 @@ def to_datacite(meta: ty.Union[dict, PublishedDandiset]) -> dict:
         "resourceTypeGeneral": "Dataset",
     }
     # meta has also attribute url, but it often empty
-    attributes["url"] = meta.url
+    attributes["url"] = str(meta.url)
```
yarikoptic, line 94: could it be `None`? then if to make robust may be:
```suggestion
    attributes["url"] = str(meta.url or "")
```

dandischema/datacite.py
```diff
@@ -190,4 +196,25 @@ def to_datacite(meta: ty.Union[dict, PublishedDandiset]) -> dict:
...
 
     datacite_dict = {"data": {"id": meta.doi, "type": "dois", "attributes": attributes}}
+
+    if validate:
+        validate_datacite(datacite_dict)
+
     return datacite_dict
+
+
+def _get_datacite_schema():
```
yarikoptic, line 207: I know that it is just a code move.

But this better be cached or (even without any analysis of how often this is called) we might end up with all kinds fails upon some intermittent connection issues etc.
Or might be better to just include that file right here in the dandischema sources/distribution -- who knows what RFing they would decide to do in datacite etc.  Then this file would always be with us and reliably available -- would make it more robust.

dandischema/models.py
```diff
@@ -133,6 +133,19 @@ def json_dict(self):
         """
         return json.loads(self.json(exclude_none=True, cls=HandleKeyEnumEncoder))
 
+    @validator("schemaKey", always=True)
+    def ensure_schemakey(cls, val):
+        tempval = val
+        if "Published" in cls.__name__:
+            tempval = "Published" + tempval
+        if "BareAsset" == cls.__name__:
+            tempval = "Bare" + tempval
```
yarikoptic, line 142: so it could make it `BarePublished<val>`?  shouldn't `Published` be added later so it would be `PublishedBare<val>`?

yarikoptic (commented):
some minor comments from cursory review

dandischema/datacite.py
satra: it's a required field, therefore cannot be `None`

dandischema/models.py
satra: that should not happen here. the second if is an equality check, in which case the only value tempval should have is `Asset`, thus turning it into `BareAsset`. anything else it should throw an error in the next few lines.

dandischema/datacite.py
yarikoptic: Ok, then just remember and tell everyone that if we relax that some time (allowing empty but not allowing None is fragile) - we need to fix code here ;-)

dandischema/datacite.py
satra: that's why i have left that by default as `False`. we should have some trust no?

people can retry on http errors for anyone using the code. i will at least make it a permalink.

satra:
@yarikoptic - feel free to merge this tomorrow if you have no other changes to make.

yarikoptic (approved):
If we talk about "more robust": I still think we should cache online fetching, and if someone is to retry -- it is for a fetcher . But that could be done later, so I am ok to proceed

PR #44, opened 2021-06-24
release.yml: Use DANDI_GITHUB_TOKEN to push to repo

### LINKED ISSUE #43: add branch protection rules and update auto config
`auto` does work with branch protection rules, it requires two things:

1. see here for the repo setup: https://github.com/neuronets/nobrainer/blob/master/.github/workflows/release.yml#L20
2. requires a personal access token instead of GITHUB_TOKEN

the one thing that i have not implemented in the nobrainer  yet is that the twine upload should only work if auto creates a new release, but not otherwise. i'll be looking into that soon.

jwodder:
Closes #43.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,6 +48,7 @@ jobs:
         with:
           fetch-depth: 0
           path: dandischema
+          token: ${{ secrets.DANDI_GITHUB_TOKEN }}
 
       - name: Delete schema-* tags so as not to confuse auto
         run: |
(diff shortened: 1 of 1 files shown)


satra:
i think this may also need the git checkout and git ref update that was in the linked yaml

satra:
or we can test this by giving it a release + patch label first.

jwodder:
@satra Do you mean the `git fetch --unshallow --tags`?  That's equivalent to passing `with: {fetch-depth: 0}` to the checkout action.

satra:
the section after that.

jwodder:
@satra The `git config --local --unset` should be taken care of by specifying `token: ${{ secrets.DANDI_GITHUB_TOKEN }}` when checking out.

satra:
ok let's check it out. i added the release + patch labels. merging now.

jwodder:
@satra It failed because the 0.2.9 tag already exists from the last failed release.

satra:
just deleted it and triggering a new release.

PR #46, opened 2021-06-25
fix: change id to indentifier for software

satra:
an RRID can be associated with different versions, therefore cannot be used as the id of an object. this fixes this in a test.


waxlamp:
@satra, who is able to review this change? Could you nominate someone in the sidebar?

satra:
@waxlamp - i added @dchiquito just because this is also related to the api PR #364

this is a rather innocuous change for this repo. so not releasing a new version at this point.

yarikoptic:
@satra we still need this PR to be merged right? (hard to assess the necessity since tests seemed to be ok without it and with it here ;-)) your PR in dandi-api was superseded by https://github.com/dandi/dandi-api/pull/372 (merged already)

satra:
yes, this is more of a clarity PR than a test PR. i just merged with master. we can merge when tests pass.

PR #48, opened 2021-06-25
enh: allow checking to see data are matched but allow for missing fields

satra:
adding a mechanism to allow checking that a metadata blob for dandiset or asset has proper values given the schema, but may have missing fields. these metadata objects should be ok for posting but not for publishing.

@dchiquito and @jwodder - would this be useful?

### DIFF
--- a/dandischema/metadata.py
+++ b/dandischema/metadata.py
@@ -158,12 +158,12 @@ def validate(obj, schema_version=None, schema_key=None, missing_ok=False):
             raise
         reraise = False
         messages = []
-        for el in exc.value.errors():
+        for el in exc.errors():
             if el["msg"] != "field required":
                 reraise = True
                 messages.append(el["msg"])
         if reraise:
-            raise ValueError(messages)
+            ValueError(messages)
 
 
 def migrate(
--- a/dandischema/metadata.py
+++ b/dandischema/metadata.py
@@ -121,7 +121,9 @@ def validate(obj, schema_version=None, schema_key=None, for_post=False):
       Name of the schema key to be used, if not specified, `schemaKey` of the
       object will be consulted
     for_post: bool, optional
-      Checks if the metadata is suitable for posting to the api server
+      Checks if the metadata is suitable for posting to the api server. If fails validation,
+      but ValidationError only includes "field required" errors, a `ValueError` is raised with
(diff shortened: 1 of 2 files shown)


jwodder:
@satra I think it would be more useful to give `Dandiset` a base class akin to `BareAsset` that lacks the fields that the server sets.

dandischema/metadata.py
```diff
@@ -119,6 +120,8 @@ def validate(obj, schema_version=None, schema_key=None):
     schema_key: str, optional
       Name of the schema key to be used, if not specified, `schemaKey` of the
       object will be consulted
+    for_post: bool, optional
+      Checks if the metadata is suitable for posting to the api server
```
yarikoptic, line 124: not clear from this description "and then what?"
```suggestion
      Checks if the metadata is suitable for posting to the api server. If fails validation,
      but ValidationError only includes "field required" errors, a `ValueError` is raised with
      the list of all errors.
```

and then `Raises` below to be adjusted to describe that it might be raised also in `for_post=True` mode.

yarikoptic (changes requested):
If decided to proceed this way (I think @jwodder 's comment on a dedicated base class might be a more "kosher"), left a comment/suggestiong on needed adjustment to docstring

satra:
what this checks is that fields are valid when provided, not if they are required. in fact this explicitly does not care about required. for dandiset this is important at the API level so that metadata can be updated through the UI, and that a user could POST dandiset meta through the API.

i don't think a base class buys us anything in this use case. further fields required can change across schema versions, and hence any base class will also have to support this. a validation function can take that into account a lot more easily than attributes in a class.

yarikoptic:
> what this checks is that fields are valid when provided, not if they are required

oh, so it sounds like `missing_ok: bool =False` flag !?  `for_post`  was really not descriptive to me so I have missed the intent of the proposed change even though it was described ;)

would need some rudimentary test.

longer term: adding such a flag we are getting into the territory of flags/settings on what to ignore, and might end up breeding other boolean flags later on such as `invalid_ok` -- just to check if nothing is missing while allowing for errors in the value specification.
When we get to BIDS and if we decide to channel bids validator's errors similarly, such "bool flags" wouldn't be sufficient really.

PR #50, opened 2021-06-28
adding AgeReferenceType

djarecka:
I'm trying to split [this older PR from dandi](https://github.com/dandi/dandi-cli/pull/450), and I believe I should update the schema before adding my test


dandischema/model_types.py
```diff
@@ -597,3 +597,27 @@
         },
     ]
 }
+
+AgeReferenceTypeDict = {
```
satra, line 601: this should be added to dandischema.

satra, line 601: can you add this to test_models.py?

PR #51, opened 2021-06-29
adding Optional[AgeReferenceType] to valueReference in Property Value

djarecka:
small addition to #50


satra:
this change would make it deviate from schema.org. perhaps you can keep the original and make the valuereference added through a propertyvalue's property id.

PR #55, opened 2021-07-08
Switch to versioningit

jwodder:


jwodder:
I believe the current test failures are because Python is importing the copy of `dandischema` in the repository rather than the one installed by tox.  The standard solution to this is the `src/` layout, which I will apply in the next commit.

jwodder:
FYI, all the tests are failing because of [this issue with dandi-api](https://github.com/dandi/dandi-api/issues/393#issuecomment-874890385).

yarikoptic:
decided to go with non-src layout in #56

PR #56, opened 2021-07-09
Switch to versioningit (non-src-layout)

jwodder:
This is an alternative to #55 if you don't want to switch to a `src/` layout.


pyproject.toml
```diff
@@ -1,10 +1,25 @@
...
+default-version = "0+unknown"
+
+[tool.versioningit.vcs]
+exclude = ["schema-*"]
+
+[tool.versioningit.format]
+# Same format as versioneer
+distance = "{version}+{distance}.{vcs}{rev}"
+dirty = "{version}+{distance}.{vcs}{rev}.dirty"
+distance-dirty = "{version}+{distance}.{vcs}{rev}.dirty"
```
yarikoptic, line 16: since it would be a common usecase AFAIK -- why `versiongit` doesn't just default to it (and also for above `default-version` which sounds like something to be within the same config section IMHO) and thus not require  all this boilerplate in every project?

pyproject.toml
```diff
@@ -1,10 +1,25 @@
 [build-system]
 # Setuptools version should match setup.py; wheel because pip will insert it noisily
-requires = ["setuptools >= 38.3.0", "wheel"]
+requires = ["setuptools >= 42.0.0", "versioningit ~= 0.1.0", "wheel"]
```
yarikoptic, line 3: FWIW, didn't try yet, but looks really neat in that it indeed seems to avoid needing all the copied code etc as versioneer does.  That though then makes projects rely on the 'stability' of versiongit but indeed `~=` (or even `==` for paranoids) pinning should provide remedy

pyproject.toml
jwodder: versioningit already has a default set of formats (listed at the end of [this section](https://github.com/jwodder/versioningit#the-tool-versioningit-format-subtable).  I just gave different ones here so that the versioning would continue to be consistent with versioneer.  As for `default-version`, that's for recovering from errors, which I believe the default response to should be the installer failing.

yarikoptic:
@satra WDYT -- should we give it a try?

satra:
sure - let's try it out.

yarikoptic:
sorry , this one fell off the radar. @jwodder - could you please rebase so we get a fresh sweep of testing, and then lets merge?

yarikoptic:
I am sorry -- radar was still "off".  Once more @jwodder and feel free to merge it if no CI errors can be coming from changes in the PR

yarikoptic:
thank you @jwodder , all green, let's give it a shot!

PR #58, opened 2021-07-09
Copy dandietag code from dandi-cli

### LINKED ISSUE #57: Move dandi-etag related logic from dand-cli to dandischema
Per @satra 's thinking and my approval nodding -- rationale is

- to lighten up dandi-api dependencies (ATM depends on dandi-cli) for PartGenerator . 
  - Not sure if that is the only component which should be moved or entire `dandietag` -- most likely entire
  - dandi-api already depends on dandischema
- to avoid deadlock of dandi-api wanting a newer dandischema and now needing to wait for newer dandi-cli which would support it
- kinda makes sense since `dandischema` is the one desiring dandi-etag

Functionality should be first copied to dandischema and removed only later, whenever dandi-cli depends on this newer dandischema released version

jwodder:
Closes #57.

### DIFF
--- a/dandischema/digests/dandietag.py
+++ b/dandischema/digests/dandietag.py
@@ -1,0 +1,202 @@
+# Derived from <https://github.com/girder/django-s3-file-field/blob/master/
+# s3_file_field/_multipart.py>, copyright Kitware, Inc. <kitware@kitware.com>
+# under the Apache 2.0 license
+
+from dataclasses import dataclass
+from hashlib import md5
+import math
+import os
+from typing import Iterator, List, NamedTuple, Optional, Union
+
+
+def mb(bytes_size: int) -> int:
+    return bytes_size * 2 ** 20
+
+
+def gb(bytes_size: int) -> int:
+    return bytes_size * 2 ** 30
+
+
+def tb(bytes_size: int) -> int:
+    return bytes_size * 2 ** 40
+
+
+class Part(NamedTuple):
+    number: int
+    offset: int
+    size: int
+
+
+@dataclass
+class PartGenerator:
+    part_qty: int
+    initial_part_size: int
+    final_part_size: int
+
+    # S3 multipart limits: https://docs.aws.amazon.com/AmazonS3/latest/dev/qfacts.html
+    # 10k is the maximum number of allowed parts allowed by S3
+    MAX_PARTS = 10_000
+    # 5MB is the minimum part size allowed by S3
+    MIN_PART_SIZE = mb(5)
+    # 5GB is the maximum part size allowed by S3
+    MAX_PART_SIZE = gb(5)
+
+    @classmethod
+    def for_file_size(cls, file_size: int) -> "PartGenerator":
+        """Method to calculate sequential part sizes given a file size"""
+        if file_size == 0:
+            return cls(0, 0, 0)
+
+        part_size = mb(64)
+
+        if file_size > tb(5):
+            raise ValueError("File is larger than the S3 maximum object size.")
+
+        if math.ceil(file_size / part_size) >= cls.MAX_PARTS:
+            part_size = math.ceil(file_size / cls.MAX_PARTS)
+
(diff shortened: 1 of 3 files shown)


dandischema/digests/dandietag.py
```diff
@@ -0,0 +1,202 @@
...
+
+    @classmethod
+    def for_file_size(cls, file_size: int) -> "PartGenerator":
+        """Method to calculate sequential part sizes given a file size"""
+        if file_size == 0:
+            return cls(0, 0, 0)
+
+        part_size = mb(64)
+
+        if file_size > tb(5):
```
satra, line 52: just an fyi that on amazon this is terabytes not tebibytes.

dandischema/digests/dandietag.py
```diff
@@ -0,0 +1,202 @@
...
+def mb(bytes_size: int) -> int:
+    return bytes_size * 2 ** 20
+
+
+def gb(bytes_size: int) -> int:
+    return bytes_size * 2 ** 30
+
+
+def tb(bytes_size: int) -> int:
+    return bytes_size * 2 ** 40
```
satra, line 21: these are mebi, gibi, and tebi - bytes. perhaps its a good time to consolidate these options in relation to aws.

dandischema/digests/dandietag.py
```diff
@@ -0,0 +1,202 @@
...
+    initial_part_size: int
+    final_part_size: int
+
+    # S3 multipart limits: https://docs.aws.amazon.com/AmazonS3/latest/dev/qfacts.html
+    # 10k is the maximum number of allowed parts allowed by S3
+    MAX_PARTS = 10_000
+    # 5MB is the minimum part size allowed by S3
+    MIN_PART_SIZE = mb(5)
+    # 5GB is the maximum part size allowed by S3
+    MAX_PART_SIZE = gb(5)
```
satra, line 42: this is the one piece of info that gives me pause in terms of what amazon is accepting, we should check on our large files what the size of each part is.  because 5 GiB > 5 GB, so if 5GB is the limit then i would assume 5 GiB would not be accepted.

dandischema/digests/dandietag.py
satra: having said that, it's possible that amazon's info is not consistent across services. but they are generally clear about TB/TiB nomenclature.

dandischema/digests/dandietag.py
yarikoptic: would indeed be nice to clear it up, but I would keep it *as is* for the purpose of this PR to provide as smooth (no side-effects) transition from dandi-cli to dandischema

dandischema/digests/dandietag.py
```diff
@@ -0,0 +1,202 @@
...
+        while p is not None and p.size <= len(self._partial_blob):
+            next_block = self._partial_blob[: p.size]
+            self._partial_blob = self._partial_blob[p.size :]
+            self._add_next_digest(md5(next_block).digest())
+            p = self.get_next_part()
+        if p is None and self._partial_blob:
+            raise ValueError("Partial update extended past end of file")
+
+
+class ETagHashlike:
```
yarikoptic, line 192: this class (and used `.partial_update`) seems to lack tests code coverage -- we never had it for this portion or it was triggered by dandi-cli other functionality somehow?  should be easy to add a dedicated unittest to ensure a more complete test coverage

yarikoptic (changes requested):
needs a tiny bit of extra tests

dandischema/digests/dandietag.py
satra: i'll file an issue. this can be merged without these changes.

dandischema/digests/dandietag.py
jwodder: Test for `ETagHashlike` added.

yarikoptic:
ok, let's get this merged -- test coverage is improved although not 100% yet, and dandi-cli errors are due to newer metadata shema version which is not yet supported by dandi-api

PR #60, opened 2021-07-15
improve help options

satra:
This PR improves several of the items that show up in the metadata editor and elsewhere.

### DIFF
--- a/dandischema/consts.py
+++ b/dandischema/consts.py
@@ -5,6 +5,7 @@ ALLOWED_INPUT_SCHEMAS = ["0.3.0", "0.3.1", "0.4.0", "0.4.1", "0.4.2", "0.4.3", "
 # migrate has a guard now for this since it cannot migrate to anything but current
 # version
 ALLOWED_TARGET_SCHEMAS = [DANDI_SCHEMA_VERSION]
+ALLOWED_VALID_SCHEMAS = ALLOWED_TARGET_SCHEMAS + ["0.4.4"]
 
 if DANDI_SCHEMA_VERSION not in ALLOWED_INPUT_SCHEMAS:
     ALLOWED_INPUT_SCHEMAS.append(DANDI_SCHEMA_VERSION)
--- a/dandischema/consts.py
+++ b/dandischema/consts.py
@@ -5,7 +5,9 @@ ALLOWED_INPUT_SCHEMAS = ["0.3.0", "0.3.1", "0.4.0", "0.4.1", "0.4.2", "0.4.3", "
 # migrate has a guard now for this since it cannot migrate to anything but current
 # version
 ALLOWED_TARGET_SCHEMAS = [DANDI_SCHEMA_VERSION]
-ALLOWED_VALID_SCHEMAS = ALLOWED_TARGET_SCHEMAS + ["0.4.4"]
+# This allows multiple schemas for validation, whereas target schemas focus on
+# migration.
+ALLOWED_VALIDATION_SCHEMAS = ALLOWED_TARGET_SCHEMAS + ["0.4.4"]
 
 if DANDI_SCHEMA_VERSION not in ALLOWED_INPUT_SCHEMAS:
     ALLOWED_INPUT_SCHEMAS.append(DANDI_SCHEMA_VERSION)
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -272,17 +272,29 @@ DANDIURL = str
 class BaseType(DandiBaseModel):
     """Base class for enumerated types"""
 
-    identifier: Optional[str] = Field(
+    identifier: Optional[Union[HttpUrl, str]] = Field(
         None,
         description="The identifier can be any url or a compact URI, preferably"
         " supported by identifiers.org.",
-        regex=r"(^(https?)://)|(^[a-zA-Z0-9]+:[a-zA-Z0-9-/\._]+$)",
+        regex=r"^[a-zA-Z0-9]+:[a-zA-Z0-9-/\._]+$",
         nskey="schema",
     )
     name: Optional[str] = Field(
         None, description="The name of the item.", max_length=150, nskey="schema"
     )
     _ldmeta = {"rdfs:subClassOf": ["prov:Entity", "schema:Thing"], "nskey": "dandi"}
+
+    class Config:
+        @staticmethod
+        def schema_extra(schema: Dict[str, Any], model: Type["BaseType"]) -> None:
+            for prop, value in schema.get("properties", {}).items():
+                if prop == "identifier":
+                    options = value.pop("anyOf", None)
+                    if "format" in options[0]:
+                        value.update(**options[0])
+                    else:
+                        value.update(**options[1])
+                    value["maxLength"] = 1000
 
 
 class AssayType(BaseType):
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -200,7 +200,7 @@ class DandiBaseModel(BaseModel, metaclass=DandiBaseModelMetaclass):
                     value["title"] = name2title(prop)
                 if value.get("format", None) == "uri":
                     value["maxLength"] = 1000
-                if value.get("pattern", "").startswith("^http"):
+                if re.match("https?://", value.get("pattern", "")):
                     value["format"] = "uri"
                 allOf = value.get("allOf")
                 anyOf = value.get("anyOf")
@@ -1048,6 +1048,7 @@ class BareAsset(CommonModel):
     blobDateModified: Optional[datetime] = Field(
         None, nskey="dandi", title="Asset file modification date and time"
     )
+    # overload to restrict with max_items=1
     access: List[AccessRequirements] = Field(
         title="Access information",
         default_factory=lambda: [AccessRequirements(status=AccessType.OpenAccess)],
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -579,7 +579,7 @@ class AccessRequirements(DandiBaseModel):
         if status == AccessType.EmbargoedAccess and embargoed is None:
             raise ValueError("An embargo date. For NIH supported awards, this "
             "must comply with NIH policy. For other datasets, this will be at "
-            "most a year from dandiset creation.")
+            "most a year from the dandiset creation.")
         return values
     """
 
@@ -724,16 +724,16 @@ class RelatedParticipant(DandiBaseModel):
     )
     relation: ParticipantRelationType = Field(
         title="Participant or subject relation",
-        description="Indicates how the current participant or subject is related"
-        " to the other participant or subject  This relation should "
-        "satisfy: Participant/Subject <relation> relatedParticipant/Subject",
+        description="Indicates how the current participant or subject is related "
+        "to the other participant or subject. This relation should "
+        "satisfy: Participant/Subject <relation> relatedParticipant/Subject.",
         nskey="dandi",
     )
 
     _ldmeta = {
         "rdfs:subClassOf": ["schema:CreativeWork", "prov:Entity"],
         "rdfs:comment": "Another participant or subject related to the current "
-        "participant or subject (e.g., another parent, sibling, child)",
+        "participant or subject (e.g., another parent, sibling, child).",
         "nskey": "dandi",
     }
 
@@ -881,20 +881,20 @@ class CommonModel(DandiBaseModel):
     )
     studyTarget: Optional[List[str]] = Field(
         None,
-        description="What the objectives or specific questions of the study are.",
+        description="Objectives or specific questions of the study.",
         nskey="dandi",
     )
     license: Optional[List[LicenseType]] = Field(
         None,
         description="Licenses associated with the item. DANDI only supports a "
         "subset of Creative Commons Licenses (creativecommons.org) "
-        "relevant to datasets.",
+        "applicable to datasets.",
         nskey="schema",
     )
     protocol: Optional[List[HttpUrl]] = Field(
         None,
         description="A list of persistent URLs describing the protocol (e.g. "
-        "protocols.io, or other dois).",
+        "protocols.io, or other DOIs).",
         nskey="dandi",
     )
     ethicsApproval: Optional[List[EthicsApproval]] = Field(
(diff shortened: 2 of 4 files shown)


dandischema/models.py
```diff
@@ -709,16 +736,21 @@ class RelatedParticipant(DandiBaseModel):
...
+    of the Participant refers to information at the timepoint when the Participant
+    engaged in the data being described.
+    """
 
     identifier: Identifier = Field(nskey="schema")
     altName: Optional[List[Identifier]] = Field(None, nskey="dandi")
 
     strain: Optional[StrainType] = Field(
-        None, description="Identifier for the strain of the sample", nskey="dandi"
+        None, description="Identifier for the strain of the participant.", nskey="dandi"
```
bendichter, line 750: I know the neuroimaging community prefers the word "participant," but for non-human subjects (which will always be the case when talking about strains) "participant" is odd, as it implies that involvement with the study is voluntary, and "subject" is better.

dandischema/models.py
```diff
@@ -709,16 +736,21 @@ class RelatedParticipant(DandiBaseModel):
...
     )
     cellLine: Optional[Identifier] = Field(
-        None, description="Cell line associated with the sample", nskey="dandi"
+        None, description="Cell line associated with the participant.", nskey="dandi"
```
bendichter, line 753: I would use "participant or subject." in cases where it could be human or non-human

dandischema/models.py
```diff
@@ -820,26 +869,36 @@ class CommonModel(DandiBaseModel):
...
+        None,
+        description="What the objectives or specific questions of the study are.",
+        nskey="dandi",
     )
     license: Optional[List[LicenseType]] = Field(
-        None, description="Licenses associated with the item.", nskey="schema"
+        None,
+        description="Licenses associated with the item. DANDI only supports a "
+        "subset of Creative Commons Licenses (creativecommons.org) "
+        "relevant to datasets.",
```
bendichter, line 880: The issue that kicked off this ticket was that licenses have abbreviations that I do not understand and I doubt other users would either. Could you spell out what these options are and how they are different from each other?

dandischema/models.py
```diff
@@ -979,12 +1042,12 @@ class BareAsset(CommonModel):
 
     wasDerivedFrom: Optional[List[BioSample]] = Field(None, nskey="prov")
     wasAttributedTo: Optional[List[Participant]] = Field(
-        None, description="Participant(s) to which this file belongs to", nskey="prov"
+        None, description="Participant(s) to which this file belongs to.", nskey="prov"
```
bendichter, line 1045: "belongs to" is odd here. The participant owns the file? maybe "participant associated with this file."

bendichter (commented):
see comments above

dandischema/models.py
satra: i can do that.

dandischema/models.py
satra: we may keep changing the list and i did not want to describe all of them here. simply a place to go to, to look. providing clear guidance on choices would be in the domain of the handbook or some tutorial, not this descriptor i think.

dandischema/models.py
```diff
@@ -381,7 +394,7 @@ class Organization(Contributor):
     identifier: Optional[RORID] = Field(
         None,
         title="A ror.org identifier",
-        description="Use an ror.org identifier for institutions",
+        description="Use an ror.org identifier for institutions.",
```
yarikoptic, line 397: FWIW, I generally dislike unnecessary trailing punctuation - just waste of bytes IMHO. If desired, makes it also easy to just add `.` consistently "in code"

dandischema/models.py
```diff
@@ -550,14 +563,26 @@ class AccessRequirements(DandiBaseModel):
...
 
+    """
+    # TODO: Enable when embargoed access is supported.
+    @root_validator
+    def open_or_embargoed(cls, values):
+        status, embargoed = values.get("status"), values.get("embargoedUntil")
+        if status == AccessType.EmbargoedAccess and embargoed is None:
+            raise ValueError("An embargo date. For NIH supported awards, this "
+            "must comply with NIH policy. For other datasets, this will be at "
+            "most a year from dandiset creation.")
```
yarikoptic, line 582:
```suggestion
            "most a year from the dandiset creation.")
```
since of a specific dandiset creation time.

dandischema/models.py
```diff
@@ -665,60 +688,76 @@ class PublishActivity(Activity):
...
+        None, title="URL of the related participant or subject", nskey="schema"
     )
     relation: ParticipantRelationType = Field(
-        title="Participant relation",
-        description="Indicates how the current participant is related to the other participant "
-        "This relation should satisfy: Participant <relation> relatedParticipant",
+        title="Participant or subject relation",
+        description="Indicates how the current participant or subject is related"
+        " to the other participant or subject  This relation should "
+        "satisfy: Participant/Subject <relation> relatedParticipant/Subject",
```
yarikoptic, line 729:
```suggestion
        description="Indicates how the current participant or subject is related "
        "to the other participant or subject. This relation should "
        "satisfy: Participant/Subject <relation> relatedParticipant/Subject.",
```
for consistency with the rest

dandischema/models.py
```diff
@@ -665,60 +688,76 @@ class PublishActivity(Activity):
...
         nskey="dandi",
     )
 
     _ldmeta = {
         "rdfs:subClassOf": ["schema:CreativeWork", "prov:Entity"],
-        "rdfs:comment": "Another participant related to the participant (e.g., another "
-        "parent, sibling, child)",
+        "rdfs:comment": "Another participant or subject related to the current "
+        "participant or subject (e.g., another parent, sibling, child)",
```
yarikoptic, line 736:
```suggestion
        "participant or subject (e.g., another parent, sibling, child).",
```

yarikoptic: and there is more missing trailing `.` below.  I would just harmonize to not have any ;-)

dandischema/models.py
```diff
@@ -820,26 +880,36 @@ class CommonModel(DandiBaseModel):
         nskey="schema",
     )
     studyTarget: Optional[List[str]] = Field(
-        None, description="What the study is related to", nskey="dandi"
+        None,
+        description="What the objectives or specific questions of the study are.",
```
yarikoptic, line 884:
```suggestion
        description="Objectives or specific questions of the study.",
```

dandischema/models.py
```diff
@@ -820,26 +880,36 @@ class CommonModel(DandiBaseModel):
...
+        nskey="dandi",
     )
     license: Optional[List[LicenseType]] = Field(
-        None, description="Licenses associated with the item.", nskey="schema"
+        None,
+        description="Licenses associated with the item. DANDI only supports a "
+        "subset of Creative Commons Licenses (creativecommons.org) "
+        "relevant to datasets.",
```
yarikoptic, line 891:
```suggestion
        "applicable to datasets.",
```

dandischema/models.py
```diff
@@ -820,26 +880,36 @@ class CommonModel(DandiBaseModel):
...
+        nskey="schema",
     )
     protocol: Optional[List[HttpUrl]] = Field(
-        None, description="A list of protocol.io URLs", nskey="dandi"
+        None,
+        description="A list of persistent URLs describing the protocol (e.g. "
+        "protocols.io, or other dois).",
```
yarikoptic, line 897:
```suggestion
        "protocols.io, or other DOIs).",
```
for consistent use of DOI abbreviation (and above `URLs`)

dandischema/models.py
```diff
@@ -903,14 +973,18 @@ def contributor_musthave_contact(cls, values):
...
-        None, nskey="schema", title="Last modification date and time", readOnly=True
+        None, nskey="schema", title="Last modification date and time.", readOnly=True
     )
 
     license: List[LicenseType] = Field(
-        min_items=1, description="Licenses associated with the item.", nskey="schema"
+        min_items=1,
+        description="Licenses associated with the item. DANDI only supports a "
+        "subset of Creative Commons Licenses (creativecommons.org) "
+        "relevant to datasets.",
```
yarikoptic, line 986:
```suggestion
        "applicable to datasets.",
```

no way to borrow it from another class/item?

dandischema/models.py
```diff
@@ -926,7 +1000,7 @@ def contributor_musthave_contact(cls, values):
     wasGeneratedBy: Optional[List[Project]] = Field(
         None,
         title="Associated projects",
-        description="Describe the project(s) that generated this Dandiset",
+        description="Describe the project(s) that generated this Dandiset.",
```
yarikoptic, line 1003:
```suggestion
        description="Project(s) that generated this Dandiset.",
```

since in general non-imperative (shorter) form is used throughout

dandischema/consts.py
```diff
@@ -5,6 +5,7 @@
 # migrate has a guard now for this since it cannot migrate to anything but current
 # version
 ALLOWED_TARGET_SCHEMAS = [DANDI_SCHEMA_VERSION]
+ALLOWED_VALID_SCHEMAS = ALLOWED_TARGET_SCHEMAS + ["0.4.4"]
```
yarikoptic, line 8: what does it mean to be a "VALID" schema?  I thought they are all "valid" at the moment of their existence in the history.  Needs some kind of a comment here at the least since separation between TARGET and VALID is not clear.

satra: changed name and added comment.

yarikoptic (changes requested):
some comments/summary from discussion

satra:
@yarikoptic - this should be ok now even for the gui. take a look at the last commit and the overall merge. we need to remove the patch/release from PR #65

dandischema/models.py
```diff
@@ -279,6 +284,18 @@ class BaseType(DandiBaseModel):
...
+    class Config:
+        @staticmethod
+        def schema_extra(schema: Dict[str, Any], model: Type["BaseType"]) -> None:
+            for prop, value in schema.get("properties", {}).items():
+                if prop == "identifier":
+                    options = value.pop("anyOf", None)
+                    if "format" in options[0]:
+                        value.update(**options[0])
+                    else:
+                        value.update(**options[1])
```
yarikoptic, line 296: not sure what `if` is about, and why choosing the first one if has `format` but 2nd one (what if it is not there and just anyOf of 1 item?). Note that `else:` path isn't unit-test covered so YMMV ;-)

dandischema/models.py
```diff
@@ -200,6 +200,8 @@ def schema_extra(schema: Dict[str, Any], model) -> None:
                     value["title"] = name2title(prop)
                 if value.get("format", None) == "uri":
                     value["maxLength"] = 1000
+                if value.get("pattern", "").startswith("^http"):
```
yarikoptic, line 203: `"pattern"` here a string, not regex,  then `^` wouldn't work. also some smartypants could have "httplike". Lets do

```suggestion
                if re.match("https?://", value.get("pattern", "")):
```
but needs `import re` on top (no there yet).

dandischema/models.py
```diff
@@ -200,6 +200,8 @@ def schema_extra(schema: Dict[str, Any], model) -> None:
...
+                    value["format"] = "uri"
```
yarikoptic, line 204: shouldn't then this come before preceding `if value.get("format", None) == "uri":` so it gets maxLength set too?

dandischema/models.py
```diff
@@ -960,6 +1048,12 @@ class BareAsset(CommonModel):
     blobDateModified: Optional[datetime] = Field(
         None, nskey="dandi", title="Asset file modification date and time"
     )
+    access: List[AccessRequirements] = Field(
```
yarikoptic, line 1051: we overload this one just to add `max_item=1` for the `BareAsset`? might be worth a comment

```suggestion
    # overload to restrict with max_items=1
    access: List[AccessRequirements] = Field(
```
even thought we might have not followed the same for other fields in similar situations yet (I think we need a helper to make it more obvious etc... not here though ;-))

yarikoptic (changes requested):
Left some suggestions and primarily a question about `schema_extra` -- some documentation + comments and ideally some kind of a test seems would be very benefitial.

dandischema/models.py
satra: since the field is a union it should generate two items for anyOf. so this if is really there just in case the order is not deterministic. i have always seen thus far the order to be the same as what is specified in the field definition.

dandischema/models.py
satra: pattern is a regex, so the match will work. if you look at all the regex' which translate into "pattern" for schema generation they all have the start of string character. but i like this change. i'll commit this and then add re.

dandischema/models.py
```diff
@@ -960,6 +1050,13 @@ class BareAsset(CommonModel):
     blobDateModified: Optional[datetime] = Field(
         None, nskey="dandi", title="Asset file modification date and time"
```
yarikoptic, line 1051:
```suggestion
        None, nskey="dandi", title="Asset file modification date and time."
```
since you like them

dandischema/models.py
```diff
@@ -979,12 +1076,14 @@ class BareAsset(CommonModel):
 
     wasDerivedFrom: Optional[List[BioSample]] = Field(None, nskey="prov")
     wasAttributedTo: Optional[List[Participant]] = Field(
-        None, description="Participant(s) to which this file belongs to", nskey="prov"
+        None,
+        description="Participant(s) or subject(s) associated with this file.",
```
yarikoptic, line 1080:
```suggestion
        description="Participant(s) or subject(s) associated with this asset.",
```
who knows where the life brings us, and there is `file`, `asset` and `asset file` used interchangeably through descriptions.  Should we unify it all?  could as well be just
```suggestion
        description="Associated participant(s) or subject(s).",
```
thus avoiding mentioning one of the three (and change the rest to follow)

yarikoptic: BioSample has it
```
 description="Participant(s) or Subject(s) associated with this sample.",
```
so IMHO could be unified between the two to just that latter form in the comment above

yarikoptic (commented):
just nit picking on descriptions - should we unify to avoid the "subject of the matter" or at least harmonize the "subject" in *Asset's?

satra:
looks like all tests are passing. the two failing checks are on released cli which does not include the docker-compose fix merged today.

yarikoptic:
eh, I hoped to release dandi-cli with https://github.com/dandi/dandi-cli/pull/743 but testing just got stuck there -- never saw such to happen.  but i guess we should just delay release until we also boost for new dandischema as soon that is what dandi-api starts to use.
Anyways -- I guess click Merge if satisfied ;-)

PR #64, opened 2021-07-21
RF(CI): run dandi-cli tests only against 3.8 (but all OSes)

yarikoptic:
ATM we really overwhelm dandischema PRs with reports from dandi-cli.
I think it will be sufficient to just test against a single python version
at least for now


jwodder:
@yarikoptic You may want to also change the `python: 3.7` on lines 34 and 38 to `python: 3.8`.

yarikoptic:
thank you @jwodder -- force-pushed.

jwodder (approved):

PR #52, opened 2021-07-07
fix: clean up optional components of the schema

satra:
This PR cleans up different elements of the schema that were a mismatch between expectations and validation mostly around the use of None.

PR #65, opened 2021-07-26
Use a generic DOI site instead of 10.80507

dchiquito:
10.80507 is our test DOI site, we need to support both that and our
production site, 10.48324.

### DIFF
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -49,7 +49,7 @@ UUID_PATTERN = (
     "[a-f0-9]{8}[-]*[a-f0-9]{4}[-]*" "[a-f0-9]{4}[-]*[a-f0-9]{4}[-]*[a-f0-9]{12}$"
 )
 ASSET_UUID_PATTERN = r"^dandiasset:" + UUID_PATTERN
-DANDI_DOI_PATTERN = r"^10.80507/dandi\.\d{6}/\d+\.\d+\.\d+"
+DANDI_DOI_PATTERN = r"^10.[0-9]{5}/dandi\.\d{6}/\d+\.\d+\.\d+"
 DANDI_PUBID_PATTERN = r"^DANDI:\d{6}/\d+\.\d+\.\d+"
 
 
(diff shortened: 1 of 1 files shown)


satra:
@yarikoptic - yup - that's why i wanted to chat.

yarikoptic:
What about:
- we initiate `schema-next` (so matches `schema-<version>` tags) branch off current master and push it here
- change base for #60 to be `schema-next`
- hard reset `master ` to 4f6e708f8816fb37ff88c83074715c54174857ba
- re-merge (some? all?) PRs some PRs merged since then but skip #52 which boosted version (do only local testing)
- force-push reset `master`
- re-initiate tests for this PR to just make sure we are in clear
- merge PR/release without changes to schema version

gotchas might come in changelog generation after we do merge `schema-next` into master since we are per se not interested in that merge but rather in merges/PRs into that branch.  I remember @jwodder investigated that abit awhile ago but I forgot what was the verdict.

satra:
this PR will also require a schema version change, so before we implement a full blown refactor, we may try to cut a release to 0.5.0 with the other PR in tow before we fully implement validation against multiple schemas.

dandischema/models.py
```diff
@@ -49,7 +49,7 @@
     "[a-f0-9]{8}[-]*[a-f0-9]{4}[-]*" "[a-f0-9]{4}[-]*[a-f0-9]{4}[-]*[a-f0-9]{12}$"
 )
 ASSET_UUID_PATTERN = r"^dandiasset:" + UUID_PATTERN
-DANDI_DOI_PATTERN = r"^10.80507/dandi\.\d{6}/\d+\.\d+\.\d+"
+DANDI_DOI_PATTERN = r"^10.[0-9]{5}/dandi\.\d{6}/\d+\.\d+\.\d+"
```
yarikoptic, line 52: @satra -- do we want to limit may be to the only two we have? in the future we should relax it anyways to allow for truly arbitrary DOI prefix I guess

satra: i see no reason not too at this point.

satra, line 52:
```suggestion
DANDI_DOI_PATTERN = r"^10.(48324|80507)/dandi\.\d{6}/\d+\.\d+\.\d+"
```

PR #66, opened 2021-07-27
Set DANDI_ALLOW_LOCALHOST_URLS when running dandi-cli tests

jwodder:


.github/workflows/test-dandi-cli.yml
```diff
@@ -13,6 +13,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     env:
       NO_ET: 1
+      DANDI_ALLOW_LOCALHOST_URLS: 1
```
yarikoptic, line 16: how this would help since IIRC this is used by dandi-api, which runs in docker and not within this environment?

jwodder: The variable is already set for dandi-api via the `docker-compose.yml`.  This is for a dandi-cli test that tries to validate an asset metadata object, which contains a localhost URL.

yarikoptic: but I don't see that variable being "used" in dandi-cli and just set/hardcoded (also not clear why in tox.ini and test.yml):

```shell
(git)lena:~/proj/dandi/dandi-cli-master[master]git
$> git grep -i LOCALHOST.URLS
.github/workflows/test.yml:      DANDI_ALLOW_LOCALHOST_URLS: 1
dandi/tests/data/dandiarchive-docker/docker-compose.yml:      DANDI_ALLOW_LOCALHOST_URLS: "1"
dandi/tests/data/dandiarchive-docker/docker-compose.yml:      DANDI_ALLOW_LOCALHOST_URLS: "1"
tox.ini:    DANDI_ALLOW_LOCALHOST_URLS=1
```

jwodder: It's required implicitly by [this line](https://github.com/dandi/dandi-cli/blob/7cc845d24def7025426f5d17283ec26793ed1f02/dandi/tests/test_dandiapi.py#L511).  We need to set it outside of the code instead of, say, via monkeypatching because the envvar needs to be set when dandischema is imported.

yarikoptic: The line is
```
asset.get_metadata()
```
I am still not getting it why needed, or why/how it effects anything

yarikoptic: Ah found it here

https://github.com/dandi/dandischema/blob/b736094324a5ccb084172d532964089f6c7fc8a5/dandischema/models.py#L43

jwodder: `get_metadata()` constructs an `Asset` instance from raw metadata, which involves pydantic validation.  Because the API instance is a Docker container with a localhost address, one of the contentUrls has a domain of "localhost", which dandischema only accepts if the environment variable in question is set on import.

yarikoptic:
Let's proceed with this, thank you @jwodder.  For longer term solution, filed https://github.com/dandi/dandischema/issues/67

PR #70, opened 2021-07-29
Datacite publish argument

dchiquito:
This is only partially tested, since the data isn't actually sent off to the test server. If we were to do that in a test, we would accumulate findable and undeletable(?) DOIs.

@yarikoptic I invite followup commits with more comprehensive testing.

### DIFF
--- a/dandischema/datacite.py
+++ b/dandischema/datacite.py
@@ -57,13 +57,18 @@ DATACITE_MAP = {el.lower(): el for el in DATACITE_IDENTYPE}
 
 
 def to_datacite(
-    meta: ty.Union[dict, PublishedDandiset], validate: bool = False
+    meta: ty.Union[dict, PublishedDandiset],
+    validate: bool = False,
+    publish: bool = False,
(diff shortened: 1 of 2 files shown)


yarikoptic:
Added labels, all green now. I would've preferred avoiding copy pasting data structures across tests and come up with reusable scenarios (otherwise harder to spot relevant code/data) but that I guess could be done later. Let's proceed and release library so could be put to use

yarikoptic:
> If we were to do that in a test, we would accumulate findable and undeletable(?) DOIs. 

FWIW at least in webui indeed "Delete" button in the menu on the left is not available for "Findable" DOIs like https://doi.test.datacite.org/dois/10.80507%2F000003%2F0.210401.2020, but present for "Draft", like https://doi.test.datacite.org/dois/10.80507%2Fdandi.000035%2F0.210729.1456  . Didn't try via API, and indeed we would not want to keep breeding DOIs (even in the test. instance)

PR #72, opened 2021-08-01
enh: add a mechanism to process metadata

satra:
This PR adds the ability to process versions of schema to automate migration when possible. We can use comments here to work through the design and use cases.

One use case is to allow dandi api server to perform migrations on save.


yarikoptic:
looking quickly back to it, not clear to me how it is different from `migrate` which can be used to "process to migrate" versions of the schema.  Since it is 1.5 years old, I even wonder if worth keeping it open at all, @satra ?

PR #73, opened 2021-08-03
add AssetSummary.dataStandard namespace

surchs:
the assetsummary `dataStandard` attribute didn't have the `nskey` set and was therefore not included in the generated JSONLD context. 
- add nskey
- give dandi namespace for JSONLD context generation

### DIFF
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -620,7 +620,7 @@ class AssetsSummary(DandiBaseModel):
     numberOfSamples: Optional[int] = Field(None, readOnly=True)  # more of NWB
     numberOfCells: Optional[int] = Field(None, readOnly=True)
 
-    dataStandard: Optional[List[StandardsType]] = Field(readOnly=True)
+    dataStandard: Optional[List[StandardsType]] = Field(readOnly=True, nskey="dandi")
     # Web UI: icons per each modality?
     approach: Optional[List[ApproachType]] = Field(readOnly=True)
     # Web UI: could be an icon with number, which if hovered on  show a list?


surchs:
I'm not quite clear why this change would lead to this reduction in coverage.

satra:
@surchs - i also don't know why the coverage is reducing with that one change, but i just realized there are a few fields in there which don't have an associated schema. perhaps we should use this PR as a way of checking if we are missing others and also to write a test to ensure that the ones that are missing are intentional and the others have it. 

`assetsSummary` was our approach of getting some descriptor of the assets into the dataset metadata. i will have to at least try and recollect if i simply missed a namespace or intentionally left them blank because we don't know what that property should be. 

currently my thought is that everything should have a namespace, and preferably no conflict when used in other parts of the model.

surchs:
@satra - I like that, makes sense to me to make the assumptions explicit. A couple of points:
1. How would you keep track of terms that are not supposed to have namespaces yet?
2. Happy to take a stab at this, however I'm leaving on a 2 week vacation next Monday and I probably won't get to it before then.
3. Any pointers for the tests you have in mind? I guess we could do something like [`test_duplicate_classes`](https://github.com/dandi/dandischema/blob/master/dandischema/tests/test_models.py#L331) and see if all except exceptions have `nskey` set. Do you have a cleaner idea?

satra:
@surchs - just getting back to this - perhaps we can meet next week to finalize this.

surchs:
Hi @satra, sorry for the delay here. Happy to have a chat about this. I'll send you an email!

satra:
closing in favor of #84 - let's continue discussion there

PR #77, opened 2021-08-14
handling schemaKey

satra:
@yarikoptic - this is an initial attempt to handle `schemaKey`, but this requires a weird thing from a python perspective.

in pydantic, a class is initialized with `kwargs`. this PR adds `schemaKey` to `kwargs` to get the validator to trigger a mismatch. this works for passing data from an existing document. however, this also means that any initialization of the class will also require the appropriate `schemaKey`. see the `test_missing_schemakey` in the PR. this is quite weird.

the alternative is a function that checks every object (i.e. dictionary) in a json record for the presence of `schemaKey`, but then this function would need to be called before initializing the class, which also seems weird.

things addressed
- [X] adding schemaKey as a requirement to JSON schema

### DIFF
--- a/dandischema/metadata.py
+++ b/dandischema/metadata.py
@@ -195,7 +195,7 @@ def migrate(
             f"master/releases/{schema_version}/dandiset.json"
         ).json()
         jsonschema.validate(obj, schema)
-    if version2tuple(schema_version) < version2tuple(DANDI_SCHEMA_VERSION):
+    if version2tuple(schema_version) < version2tuple("0.4.0"):
         if obj.get("schemaKey") is None:
             obj["schemaKey"] = "Dandiset"
         id = str(obj.get("id"))
@@ -219,6 +219,13 @@ def migrate(
             obj["assetsSummary"] = {"numberOfFiles": 0, "numberOfBytes": 0}
         if obj.get("manifestLocation") is None:
             obj["manifestLocation"] = []
+    if version2tuple(schema_version) < version2tuple(DANDI_SCHEMA_VERSION):
+        for val in obj.get("about", []):
+            if "schemaKey" not in val:
+                val["schemaKey"] = "GenericType"
+        for val in obj.get("access", []):
+            if "schemaKey" not in val:
+                val["schemaKey"] = "AccessRequirements"
         obj["schemaVersion"] = to_version
     return obj
 
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -135,6 +135,11 @@ class DandiBaseModelMetaclass(ModelMetaclass):
 class DandiBaseModel(BaseModel, metaclass=DandiBaseModelMetaclass):
     id: Optional[str] = Field(description="Uniform resource identifier", readOnly=True)
 
+    def __init__(self, **kwargs):
+        if "schemaKey" not in kwargs:
+            kwargs["schemaKey"] = "MissingSchemaKey"
+        super().__init__(**kwargs)
+
     def json_dict(self):
         """
         Recursively convert the instance to a `dict` of JSONable values,
@@ -155,6 +160,14 @@ class DandiBaseModel(BaseModel, metaclass=DandiBaseModelMetaclass):
                 f"schemaKey {tempval} does not match classname {cls.__name__}"
             )
         return val
+
+    @root_validator(pre=True)
+    def check_schemaKey_present(cls, values):
+        if "schemaKey" not in values:
+            raise ValueError(
+                f"schemaKey missing from object {values} in class {cls.__name__}"
+            )
+        return values
 
     @classmethod
     def unvalidated(__pydantic_cls__: Type[BaseModel], **data: Any) -> BaseModel:
@@ -190,6 +203,9 @@ class DandiBaseModel(BaseModel, metaclass=DandiBaseModelMetaclass):
         @staticmethod
         def schema_extra(schema: Dict[str, Any], model) -> None:
             schema["title"] = name2title(schema["title"])
+            if schema["type"] == "object":
+                schema["required"] = schema.get("required", [])
+                schema["required"].append("schemaKey")
             for prop, value in schema.get("properties", {}).items():
                 if schema["title"] == "Person":
                     if prop == "name":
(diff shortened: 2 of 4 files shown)


yarikoptic:
> @yarikoptic - this is an initial attempt to handle `schemaKey`, but this requires a weird thing from a python perspective.

Could you please elaborate on what you mean by "handle" in this context?

satra:
`schemaKey` was not being stored in object, resulting in no clear way of assigning some types. and apparently since it wasn't required validation wasn't catching it.

this PR (which has changed since the original attempt) now:
- adds `schemaKey` as a required property for every object
- detects and raises a migration error for some known places where conflict could arise
- adds schemaKey for places where there should not be a conflict
- adds jsonschema based validation to the `validate` function. 

so handle means to always track `schemaKey`

satra:
@yarikoptic - i think this is good for a review/merge. i would still like to get seb and dorota's PRs in before release.

dandischema/metadata.py
```diff
@@ -101,19 +101,34 @@ def publish_model_schemata(releasedir: str) -> Path:
...
+    validator = jsonschema.Draft7Validator(
+        schema, format_checker=jsonschema.draft7_format_checker
+    )
+    error_list = []
+    for error in sorted(validator.iter_errors(data), key=str):
+        if missing_ok and "is a required property" in error.message:
+            continue
+        error_list.append([error.message, tuple(error.absolute_path)])
+    if error_list:
+        raise jsonschema.ValidationError(str(error_list))
```
yarikoptic, line 114: this would collapse all errors into one string of a list, making it impossible to print it nicely or "analyze" above.
I guess should be ok for now, but I think we might need to come up with some consistent mechanism between dandischema and dandi-cli to  "encapsulate"  multiple errors as a singular exception which would 
- contain `.errors` list of all individual errors/exceptions (not sure if wrapped into some adapter here to provide consistent interface across errors)
- have nice `__str__` and/or `__repr__`

dandischema/metadata.py
```diff
@@ -168,7 +195,7 @@ def validate(obj, schema_version=None, schema_key=None, missing_ok=False):
                 reraise = True
                 messages.append(el["msg"])
         if reraise:
-            ValueError(messages)
+            raise ValueError(messages)
```
yarikoptic, line 198: and that would be another location to use the aforementioned encapsulation of errors.

dandischema/models.py
```diff
@@ -135,6 +135,13 @@ def __new__(cls, name, bases, dct):
 class DandiBaseModel(BaseModel, metaclass=DandiBaseModelMetaclass):
     id: Optional[str] = Field(description="Uniform resource identifier", readOnly=True)
 
+    """
+    def __init__(self, **kwargs):
+        if "schemaKey" not in kwargs:
+            kwargs["schemaKey"] = "MissingSchemaKey"
+        super().__init__(**kwargs)
+    """
+
```
yarikoptic, line 144: what are those docstrings with code left for? Delete:
```suggestion
```
or add some comment on their purpose?

dandischema/models.py
```diff
@@ -156,6 +163,16 @@ def ensure_schemakey(cls, val):
...
+    """
+    @root_validator(pre=True, allow_reuse=True)
+    def check_schemaKey_present(cls, values):
+        if "schemaKey" not in values:
+            raise ValueError(
+                f"schemaKey missing from object {values} in class {cls.__name__}"
+            )
+        return values
+    """
+
```
yarikoptic, line 175:
```suggestion
```

dandischema/models.py
```diff
@@ -190,6 +207,9 @@ class Config:
         @staticmethod
         def schema_extra(schema: Dict[str, Any], model) -> None:
             schema["title"] = name2title(schema["title"])
+            if schema["type"] == "object":
+                schema["required"] = schema.get("required", [])
+                schema["required"].append("schemaKey")
```
yarikoptic, line 212: a matter of taste but could be 
```suggestion
                schema["required"] = schema.get("required", []) + ["schemaKey"]
```
BTW -- are we certain it will not be in "required" already somehow and we would end up with two? may be should be conditioned?

dandischema/tests/test_metadata.py
```diff
@@ -44,13 +44,40 @@ def test_dandiset(schema_dir):
 def test_pydantic_validation(schema_dir):
     with (METADATA_DIR / "meta_000004.json").open() as fp:
         data_as_dict = json.load(fp)
+    data_as_dict["schemaVersion"] = "0.4.4"
```
yarikoptic, line 47: why just not to modify in the file since that schema version is not supported?
```
$> grep schemaVersion ./dandischema/tests/data/metadata/meta_000004.json
  "schemaVersion": "0.4.0",
```
and if there is a test to verify that unsupported causes error, only then do explicit resetting of the version with a comment on purpose .
Otherwise later requires "thinking" on why it is reset here

yarikoptic (commented):
Left some comments/notes, nothing critical really

dandischema/metadata.py
satra: it would be nice to have this. let me look into if python itself has a way of doing this.

dandischema/tests/test_metadata.py
yarikoptic: I think Satra has RFed for this, can be resolved

dandischema/metadata.py
```diff
@@ -101,19 +101,34 @@ def publish_model_schemata(releasedir: str) -> Path:
...
+    validator = jsonschema.Draft7Validator(
+        schema, format_checker=jsonschema.draft7_format_checker
+    )
+    error_list = []
+    for error in sorted(validator.iter_errors(data), key=str):
+        if missing_ok and "is a required property" in error.message:
+            continue
+        error_list.append([error])
+    if error_list:
+        raise ValueError(error_list)
```
yarikoptic, line 114: Echoing prior discussion, Let's at least define our own `class ValidationError(ValueError)` may be?

satra: do you want to do it now? or later?

yarikoptic: If no objections, will push that change in 5 minutes. Better not to delay ;-)

yarikoptic: Later we would need to reactor anyways though

satra: ok then go for it.

dandischema/metadata.py
```diff
@@ -13,6 +13,7 @@
     ALLOWED_VALIDATION_SCHEMAS,
     DANDI_SCHEMA_VERSION,
 )
+from .exceptions import ValidationError
```
satra, line 16: `exceptions.py` ?

dandischema/metadata.py
satra: any other changes you plan, or should i merge?

PR #78, opened 2021-08-17
[fix] creating identifier from url in relatedResource

djarecka:
I've modified `to_datacite`, so it should try to get DOI identifier from identifier that has form of url.

I'm using some assumption about doi from `odf.io` and `biorxiv.org`, not sure if they are always right...

@bendichter - could you point me to datasets/metadata that I could test this fix.

addresses issue from `dandiarchive`: https://github.com/dandi/dandiarchive/issues/843


djarecka:
@bendichter -pinging

bendichter:
@djarecka I don't know if there are any dandisets with improperly formatted DOIs. I think I fixed them manually so that publication could go through. You could try this yourself though. See if it works on e.g. http://purl.obolibrary.org/obo/UBERON_0001384

djarecka:
wasn't sure how to use the website here, but I added some basic unit tests.

dandischema/datacite.py
```diff
@@ -176,21 +176,39 @@ def to_datacite(
...
-                    continue
-                ident_tp, ident_id = rel_el.identifier.split(":", 1)
-                if ident_tp.lower() in DATACITE_MAP:
-                    ident_tp = DATACITE_MAP[ident_tp.lower()]
-                else:
-                    raise ValueError(
-                        f"identifier has to be from the list: {DATACITE_IDENTYPE}, "
-                        f"but {ident_tp} provided"
-                    )
+                if "doi:" in rel_el.identifier.lower():
```
yarikoptic, line 179: Why `in` and not `.startswith`? If some free text with embedded doi, url - might error

dandischema/datacite.py
```diff
@@ -176,21 +176,39 @@ def to_datacite(
...
+                    else:
+                        raise ValueError(
+                            f"identifier has to be from the list: {DATACITE_IDENTYPE}, "
+                            f"but {ident_tp} provided"
+                        )
+                else:  # trying to find identifier if url provided
+                    if "doi.org/" in rel_el.identifier:
+                        ident_tp = "DOI"
+                        ident_id = rel_el.identifier.split("doi.org/")[1]
+                    elif "osf.io/" in rel_el.identifier:
```
satra, line 192: we may want to keep osf.io urls as urls. not all osf.io urls are DOIs. only some are. so turning them into DOIs would require actually checking if the DOI is valid.

PR #81, opened 2021-09-08
fixing to_datacie, allowing for None in roleName

### LINKED ISSUE #80: to_datacite: TypeError: argument of type 'NoneType' is not iterable
received in sentry

```
Issue                                                                                                                                                                                                                
error   TypeError /api/dandisets/{dandiset__pk}/versions/{version}|draft)/publish/                                                                                                                                   
        argument of type 'NoneType' is not iterable                                                                                                                                                                  
                                                                                                                                                                                                                     
ID: bae6a1e3a1224df4a9b94c59fd137e2a                                                                                                                                                                                 
Sept. 8, 2021, 9:39:48 a.m. UTC                                                                                                                                                                                      
                                                                                                                                                                                                                     
  Exception                                                                                                                                                                                                          
                                                                                                                                                                                                                     
TypeError: argument of type 'NoneType' is not iterable                                                                                                                                                               
(9 additional frame(s) were not displayed)                                                                                                                                                                           
...                                                                                                                                                                                                                  
  File "guardian/decorators.py", line 124, in _wrapped_view                                                                                                                                                          
    return view_func(request, *args, **kwargs)                                                                                                                                                                       
  File "dandiapi/api/views/version.py", line 116, in publish                                                                                                                                                         
    new_version.doi = doi.create_doi(new_version)                                                                                                                                                                    
  File "dandiapi/api/doi.py", line 38, in create_doi                                                                                                                                                                 
    doi, request_body = _generate_doi_data(version)                                                                                                                                                                  
  File "dandiapi/api/doi.py", line 25, in _generate_doi_data                                                                                                                                                         
    return (doi, to_datacite(metadata, publish=publish))                                                                                                                                                             
  File "dandischema/datacite.py", line 113, in to_datacite                                                                                                                                                           
    if RoleType("dcite:Sponsor") in contr_el.roleName:                                                                                                                                                               
                                                                                                                                                                                                                     
  Request                                                                                                                                                                                                            
                                                                                                                                                                                                                     
URL    https://api.dandiarchive.org/api/dandisets/000034/versions/draft/publish/                                                                                                                                     
Method POST   
```

djarecka:
fixes #80

### DIFF
--- a/dandischema/datacite.py
+++ b/dandischema/datacite.py
@@ -110,7 +110,7 @@ def to_datacite(
     contributors = []
     creators = []
     for contr_el in meta.contributor:
-        if RoleType("dcite:Sponsor") in contr_el.roleName:
+        if contr_el.roleName and RoleType("dcite:Sponsor") in contr_el.roleName:
             # no info about "funderIdentifierType", "awardUri", "awardTitle"
             dict_fund = {"funderName": contr_el.name}
             if contr_el.identifier:
@@ -143,7 +143,7 @@ def to_datacite(
         elif isinstance(contr_el, Organization):
             contr_dict["nameType"] = "Organizational"
 
-        if RoleType("dcite:Author") in getattr(contr_el, "roleName", []):
+        if contr_el.roleName and RoleType("dcite:Author") in contr_el.roleName:
             create_dict = deepcopy(contr_dict)
             create_dict["creatorName"] = create_dict.pop("contributorName")
             creators.append(create_dict)
@@ -153,7 +153,7 @@ def to_datacite(
                 continue
 
         contr_all = []
-        if getattr(contr_el, "roleName"):
+        if contr_el.roleName:
             contr_all = [
                 el.name for el in contr_el.roleName if el.name in DATACITE_CONTRTYPE
             ]
--- a/dandischema/tests/test_datacite.py
+++ b/dandischema/tests/test_datacite.py
@@ -131,7 +131,7 @@ def test_datacite(dandi_id, schema):
                 ),
             },
         ),
-        # additional contributor with dandi:Author
+        # additional contributor with dandi:Author, and one without roleName
         (
             {
                 "contributor": [
@@ -143,12 +143,16 @@ def test_datacite(dandi_id, schema):
                         "name": "B_last, B_first",
                         "roleName": [RoleType("dcite:Author")],
                     },
+                    {
+                        "name": "C_last, C_first"
+                    },
+
                 ],
             },
             {
                 "creators": (1, {"name": "B_last, B_first"}),
                 "contributors": (
-                    1,
+                    2,
                     {"name": "A_last, A_first", "contributorType": "ContactPerson"},
                 ),
             },


dandischema/datacite.py
```diff
@@ -153,7 +153,7 @@ def to_datacite(
                 continue
 
         contr_all = []
-        if getattr(contr_el, "roleName"):
+        if contr_el.roleName:
```
yarikoptic, line 156: what was rationale for this particular line change? `getattr` is safer and I might have been used for a reason (some prior version having no `.roleName`?) so I would have just kept this as is.

Otherwise looks good to me

waxlamp: `getattr()` raises `AttributeError` if the attribute doesn't exist (according to the [`hasattr()` docs](https://docs.python.org/3/library/functions.html#hasattr)), so this change is actually better than calling `getattr()`, because it's a more standard expression of pulling an attribute from an object.

djarecka: I've changed it only because I realize that `roleName` has `None` as default, so the attribute should always exist, but I could revert

waxlamp: As I mentioned, I'd keep this the way it is now, since the old and new versions have exactly the same effect, but the new version is more standard.

yarikoptic: Thank you @waxlamp ! Indeed, this RF is ok here (although unrelated to the purpose of the PR).  I guess while analyzing for it, I have also misguided myself by the fact that `dict.get(key)` does default to not blowing up but just returning `None` if `key` is not in the `dict`.  May be original intention was similarly misguided in 4878d233b7f9b8fbbc81423e8eb8d6645d773f92 @djarecka but lacked that specification of the default?

waxlamp: A perhaps little known fact (one that I didn't know until recently myself): `dict.get()` takes a second argument for the default value to return for a missing key (which itself defaults to `None`).

djarecka: @yarikoptic - the only reason why I've changes it here was because I was exploring the default behavior if nothing is passed as `roleName` (that was the example that caused an issue for you).

djarecka: @waxlamp - yes, that's true that `getattr` returns `None`, so it's often a save check that avoids raising any exceptions. Probably that was the reason why I used it at the beginning.

yarikoptic: > @waxlamp - yes, that's true that `getattr` returns `None`, so it's often a save check that avoids raising any exceptions. Probably that was the reason why I used it at the beginning.

may be a typo (`dict.get`  instead of `getattr`?) since that is the point @waxlamp [mentioned above](https://github.com/dandi/dandischema/pull/81#discussion_r705548634) that `getattr` does raise exception if the default is not specified explicitly (which differs in behavior from `dict.get` where default does not need to be specified and defaults to `None`)

yarikoptic: overall -- there is no point to use `getattr(obj, "string with the fieldname")` over `obj.fieldname`, unless providing an explicit default, as e.g. `getattr(obj, "string with the fieldname", None)`

djarecka: yes, I'm sorry. I've mixed also in the discussion `getattr` and `dict.get`. But I understand that this could stay like this

yarikoptic:
anyways , for the purpose of a fix I think we have reviewed it and there is even a test -- nothing should hold the progress. to facilitate immediate availability of the fix for dandi-api I will add 'release' label and merge. Current git describe: 0.3.2-3-g1054377 -- with the release will test new versiongit ;-)

PR #83, opened 2021-09-15
remove NC license

satra:
this should remove NC license, which should not allow validation for any dandiset still remaining with the license. this updates the schema version as well.


satra:
@yarikoptic - i'm thinking of getting this out there before the bigger migration PR.

PR #84, opened 2021-09-21
adds all properties to context with implicit dandi schema

satra:
supercedes #73 

@surchs - this PR incorporates the changes in #77 and then extends it to address the namespace issue.  i did not include explicit namespace in this PR.

- adds all properties to context
- adds test to ensure there is no conflict

### DIFF
--- a/dandischema/metadata.py
+++ b/dandischema/metadata.py
@@ -62,15 +62,18 @@ def generate_context():
                 fields[name] = "@type"
             elif name == "digest":
                 fields[name] = "@nest"
-            elif "nskey" in field.field_info.extra:
-                if name not in fields:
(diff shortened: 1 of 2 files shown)


surchs:
Thanks @satra for the ping and really sorry I didn't manage to help with this in in time - I got swamped. I had a look and I think this will be very helpful to continue playing with the jsonld of the dandisets in the future. I'll give it a try.

satra:
@surchs - thanks. i'm going to merge this in and we can evolve the schema in the future.

@yarikoptic - i can label this for release, unless we want to get any other things in. and perhaps test out the schema with vjsf before release to make sure that bit works. that's something we don't have any automation on.

PR #86, opened 2021-09-25
Release 0.6.0

satra:
Creating this branch so that we can release a new version of the schema.

- [X] Checked briefly on codepen that schema works with 0.6.0
- [x] Recheck dandiset metadata migration
- [x] Check asset metadata migration

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,4 +4,20 @@ and assets.
 
 To use: `pip install dandischema`
 
-More documentation coming soon.
+Every Dandiset and associated asset has a metadata object that can be retrieved using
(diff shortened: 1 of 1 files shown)


satra:
things to consider - whether everything is migrated to 0.6.0

notes for API/CLI:
- asset metadata will need to migrated for existing dandiset (without asset id modification)
- dandiset metadata will need to migrated for existing dandiset drafts
- for any published version including some where we need to adjust licenses.
- CLI should move up to this version and declare prior versions to be upgraded and add more validation logic

PR #88, opened 2021-09-27
Mark tests requiring network access and add a test workflow that disables network access

### LINKED ISSUE #87: dandischema/tests/test_metadata.py::test_migrate_041 requires network access
The `dandischema/tests/test_metadata.py::test_migrate_041` test fails inside the root sandbox for system-wide install. It works if I test it manually based on the source directory as my user. Not really a big issue since it can be easily be blacklisted for the system-wide install.

Is it in any case possible to make it work without network access?
This is the full log showing it fail: https://ppb.chymera.eu/cbaa53

jwodder:
Closes #87.


yarikoptic:
are test fails against dandi-cli  the same as recently encountered in dandisets (https://github.com/dandi/dandisets/pull/87/commits/7ec49def8935aca47a4efc2b49bfe943325634c4) requiring to wait for it to become "valid"?  I even wonder now if such logic should be somehow incorporated into `.publish` with some time out (e.g. of 5 seconds, which should be sufficient for unittesting) since it seems to become a common pattern and ideally should not be duplicated across codebase?

jwodder:
@yarikoptic That failure reason only applies to one test.  Most of the failing tests are likely due to some mismatch in dandischema versions, and the failure is only exacerbated by the fact that trying to show the validation errors fails because dandi-cli's master branch expects version validation errors to be reported in a key that was removed from the API without telling me (addressed in https://github.com/dandi/dandi-cli/pull/775/commits/45a4215089699366c4b067d86417b86cd5005c50).

yarikoptic:
> is only exacerbated by the fact that trying to show the validation errors fails because dandi-cli's master branch expects version validation errors to be reported in a key that was removed from the API without telling me (addressed in [dandi/dandi-cli@45a4215](https://github.com/dandi/dandi-cli/commit/45a4215089699366c4b067d86417b86cd5005c50)).

shouldn't we have caught it somehow by integration testing in dandi-api against dandi-cli -- https://github.com/dandi/dandi-api/actions/workflows/cli-integration.yml seems to stay green

jwodder:
@yarikoptic We don't have any tests that expect Dandiset validation to fail.

yarikoptic:
Thank you @jwodder, let's proceed.

@Thechymera , set `DANDI_TESTS_NONETWORK=1` env var while running tests and those which require outbound network will be skipped

TheChymera:
thanks, it worked ^^

PR #89, opened 2021-09-27
Better exception classes

dchiquito:
I had some difficulty parsing the exceptions being thrown now that I need to differentiate between pydantic and jsonschema validation errors. This PR should mitigate that pain a bit.

Not a blocker, but nice to have in the future.

### DIFF
--- a/dandischema/exceptions.py
+++ b/dandischema/exceptions.py
@@ -1,9 +1,18 @@
-class ValidationError(ValueError):
-    """Validation errors were detected.
+class JsonschemaValidationError(ValueError):
+    """Validation errors were detected by jsonschema.
 
     All errors are contained in the .args[0] of the exception instance.
     """
+    @property
+    def errors(self):
+        return self.args[0]
 
+
+class PydanticValidationError(ValueError):
+    """Validation errors were detected by pydantic.
+
+    All errors are contained in the .args[0] of the exception instance.
+    """
     @property
     def errors(self):
         return self.args[0]
--- a/dandischema/exceptions.py
+++ b/dandischema/exceptions.py
@@ -2,7 +2,11 @@ from typing import Any, Dict, List
 import jsonschema.exceptions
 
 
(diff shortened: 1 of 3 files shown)


dandischema/exceptions.py
```diff
@@ -1,9 +1,22 @@
-class ValidationError(ValueError):
-    """Validation errors were detected.
+from typing import Any, Dict, List
+import jsonschema.exceptions
+
+
+class JsonschemaValidationError(ValueError):
```
yarikoptic, line 5: Let's then may be keep original `ValidationError` as the `super` class for both more specific types of errors?

dandischema/metadata.py
```diff
@@ -113,9 +113,9 @@ def _validate_obj_json(data, schema, missing_ok=False):
     for error in sorted(validator.iter_errors(data), key=str):
         if missing_ok and "is a required property" in error.message:
             continue
-        error_list.append([error])
+        error_list.append(error)
```
yarikoptic, line 116: yikes, thank you @dchiquito

dandischema/tests/test_metadata.py
```diff
@@ -53,7 +53,7 @@ def test_pydantic_validation(schema_dir):
 
 
 def test_json_schemakey_validation():
-    with pytest.raises(ValidationError) as exc:
+    with pytest.raises(JsonschemaValidationError) as exc:
```
yarikoptic, line 56: would you be so kind to also change then the line below (can't offer suggestion out of diff scope) to be
```
assert "'schemaKey' is a required property" in str(exc.value.errors)
```
instead of `.args[0]`? or if you like - I can push that change

yarikoptic (approved):
approve in general, thank you!  might be worth tuning up test just a bit

yarikoptic:
thank you @dchiquito , looks kosher to me, so let's proceed... and let's release right away

PR #91, opened 2021-10-01
fix: associate schema key with schemas for older versions

satra:
this fixes that only dandiset.json was being used for json validation when schema version of the object did not match current schema.

### DIFF
--- a/dandischema/metadata.py
+++ b/dandischema/metadata.py
@@ -177,7 +177,6 @@ def validate(
             f"Allowed are: {', '.join(ALLOWED_VALIDATION_SCHEMAS)}."
         )
     if json_validation:
-        print(schema_version, DANDI_SCHEMA_VERSION)
         if schema_version == DANDI_SCHEMA_VERSION:
             klass = getattr(models, schema_key)
             schema = klass.schema()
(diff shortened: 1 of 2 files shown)


dandischema/metadata.py
```diff
@@ -181,9 +181,21 @@ def validate(
             klass = getattr(models, schema_key)
             schema = klass.schema()
         else:
+            schema_map = {
+                "Dandiset": "dandiset",
+                "PublishedDandiset": "published-dandiset",
+                "Asset": "asset",
+                "PublishedAsset": "published-asset",
+            }
```
yarikoptic, line 189: Later (away from laptop ATM) we should centralize (the same dict should be used there) with with https://github.com/dandi/dandischema/blob/master/dandischema/metadata.py#L90

yarikoptic (approved):
Looks good to me. Left a comment on need to avoid duplication, could be done later

dandischema/metadata.py
yarikoptic: FWIW -- pushed quick RF to "centralize" in 44845973ed0e90a117c7ef427439655f8dd4632f

dchiquito (approved):

PR #92, opened 2021-10-05
enh: add subject and sample aggregating for bids dandisets

satra:
@djarecka and @thechymera - this is a quick hack to do bids counting at the dandiset level. ideally once you update the metadata extraction we can get rid of these changes.


dandischema/metadata.py
```diff
@@ -291,8 +292,10 @@ def _add_asset_to_stats(assetmeta: Dict[str, Any], stats: _stats_type) -> None:
             if "species" in value:
                 if value["species"] not in stats["species"]:
                     stats["species"].append(value["species"])
-            if value["identifier"] not in stats["subjects"]:
-                stats["subjects"].append(value["identifier"])
+            if value.get("identifier", None):
+                subject = value["identifier"].replace("_", "-")
```
yarikoptic, line 296: what is this for?

yarikoptic: so here it seems you are trying to get `subject` from two possible places -- wasAttributedTo (to be extracted by dandi-cli) and then the filename.  I don't think we have any consistency checks (e.g. wasAttributedTo says one thing and filename -- another), so may be at least here we could
- just assign `subject = None` first
- here assign to the value from identifier
- when discovering from filename below (I would also add check that part index is 0 -- shouldn't be in the middle of the name) -- if `subject is not None and subject != subject_from_filename:` -- issue a warning, and keep the one from identifier.  
This way later whenever dandi-cli gets proper bids extraction and thus attributedTo -- it would take precedence etc.

WDYT?

satra: the replacement is there because dandi cli replaces underscores with hyphens to make values of keys in a filename bids compliant.

this is the aggregate function. perhaps that kind of thing should go into a validation function. it would be a good sanity check. but i don't want to turn this into a bids validator (at least not yet). the main purpose is to provide some additional summary for these dandisets so we can improve the reporting. 

this also assumes there is a sample even if it's just a json file. technically it should really also have a proper microscopy extension (png/tiff/ngff/h5). very much a hack for now.

Issue #76, opened 2021-08-11
Need to provide support for different "instances" of DANDI archive


yarikoptic:
ATM
```
id: DANDI:000018/draft
identifier: DANDI:000018
```
on both staging instance AND on deployed for completely disconnected dandisets.   I have committed a "crime" and with my glorious admin powers uploaded some test file to the main deployment instance instead of the staging (not a bigger -- this dandiset is empty on deployment) since forgot to add instance I want to upload to.

I think:
- staging instance should acquire its own prefix, such as `DANDI-STAGING` so any new dandiset created on staging would get that new identifier
- schema should allow for having those few (+ for tests etc) to start with explicitly listed prefixes, but eventually we might make it more of a schema "parametrization" somehow e.g. via env var (but see #67 for DANDI_ALLOW_LOCALHOST_URLS - so may be some "better" way)
- `dandischema` model for the *Dandiset should gain some `sameAs` or alike to provide alternative `identifier`s (if not `id`s) to allow for mapping
- `dandi-cli` should default to the instance to interact with corresponding prefix, and allowed to be overridden -- but then consult the `sameAs` for possible mapping to another "identifier" on the other instance
- `dandi-api` instance should be provided prefix to be used (so `DANDI` for deployment, `DANDI-STAGING` for staging)

This might become also relevant for orchestrating data interactions for embargoed datasets, if we would allow for some kind of a hybrid, where some prior versions might be made public and subsequent embargoed/private and we have separate DANDI archives for public/private.

kabilar:
Thanks @yarikoptic.  This would also be helpful for our LINC staging and production instances.  cc @aaronkanzer

yarikoptic:
@satra @kabilar @aaronkanzer we better attack this one asap.  I think there should be allowance for an arbitrarily long list of `id` and `identifier`s, but then we would get affects on `manifestLocation` and `url`.
We should review that in the scope of "instances" as described in dandi-cli https://github.com/dandi/dandi-cli/blob/e8611498496f90f722370e7b852cfc2ef1fd08b5/dandi/consts.py#L120 and extend with missing metadata (like `identifiers_org_prefix` which is present for main instance as `DANDI`) etc.

kabilar:
Thank you, Yarik.  Agree that this is an important issue.

Is this a blocker for something you are working on?  Trying to decide if we should change our current priority list and bump down an active target or the targets in the queue (i.e. improving the user sign up flow, and DOI handling).

[...]

PR #98, opened 2021-11-23
Modify `PublishedDandiset` `url` regex to allow any URL

mvandenburgh:
Currently, the `url` permalink of a published dandiset must begin with `https://dandiarchive.org` in order to pass validation. [This dandi-api PR](https://github.com/dandi/dandi-api/pull/613) eliminates the use of that hardcoded URL, so dandisets published in staging or local dev environments will fail validation. 

This PR modifies the regex for the `url` permalink of a published dandiset so that any URL is valid.

### DIFF
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -1151,7 +1151,7 @@ class PublishedDandiset(Dandiset, Publishable):
     url: DANDIURL = Field(
         readOnly=True,
         description="Permalink to the Dandiset.",
-        regex=r"^https://dandiarchive.org/dandiset/\d{6}/\d+\.\d+\.\d+$",
+        regex=r"^https?:\/\/\S+\/dandiset\/\d{6}\/\d+\.\d+\.\d+$",
         nskey="schema",
     )
 
--- a/dandischema/tests/test_models.py
+++ b/dandischema/tests/test_models.py
@@ -285,8 +285,8 @@ def test_dantimeta_1():
     error_msgs = [
         "field required",
         "A Dandiset containing no files or zero bytes is not publishable",
-        'string does not match regex "^https://dandiarchive.org/dandiset/'
-        '\\d{6}/\\d+\\.\\d+\\.\\d+$"',
+        'string does not match regex "^https?:\\/\\/\\S+\\/dandiset\\/'
+        '\\d{6}\\/\\d+\\.\\d+\\.\\d+$"',
         'string does not match regex "^DANDI:\\d{6}/\\d+\\.\\d+\\.\\d+"',
     ]
     assert all([el["msg"] in error_msgs for el in exc.value.errors()])


satra:
@mvandenburgh - at the model level i don't think we can change this, since part of the goal here is that the published schema is not dependent on where it's deployed. we could consider overwriting the model code to take into account an environment variable, but it should default to dandiarchive.org. 

this also means that in the production site the WEB_APP_URL cannot be gui.dandiarchive.org it has to be dandiarchive.org

the alternative would be to also publish staging schemas.

yarikoptic:
> @mvandenburgh - at the model level i don't think we can change this, since part of the goal here is that the published schema is not dependent on where it's deployed

I think that in the long run (of "distributed" network of dandi archives) it would not necessarily be true -- people should be able to publish on some other instances.  Thus also relates to #76 .  

> we could consider overwriting the model code to take into account an environment variable, but it should default to dandiarchive.org.

I think that this is a good direction to take.

satra:
@yarikoptic - i simply mean that currently we have several services (identifiers.org, dandiarchive.org, datacite) and jsonld context files (github.com), in addition to our production and staging instance. and we cannot dissociate all mappings easily at this stage. we would essentially need parallel mappings to truly dissociate all of these components. doable but a heavier lift than just simply updating the schema.

dchiquito:
I would have thought https://github.com/dandi/dandischema/blob/master/dandischema/models.py#L41-L45 would suffice for any test deployments.

mvandenburgh:
> I would have thought [`master`/dandischema/models.py#L41-L45](https://github.com/dandi/dandischema/blob/master/dandischema/models.py?rgh-link-date=2021-11-29T22%3A20%3A51Z#L41-L45) would suffice for any test deployments.

Yes that would work for this. But, I think the more high level objective is to completely decouple the DANDI stack from a specific URL as @yarikoptic describes in https://github.com/dandi/dandischema/issues/76. In other words, a deployment at some other URL wouldn't be a "test deployment", but simply another deployment of DANDI, and the URL should pass validation without needing to set a test flag in dandischema.


> this also means that in the production site the WEB_APP_URL cannot be gui.dandiarchive.org it has to be dandiarchive.org

I don't think there would be any issues with setting `WEB_APP_URL` to `https://dandiarchive.org`.

yarikoptic:
So, how should we proceed forward on this PR - RF to use env vars?

satra:
> RF to use env vars?

i would say yes. the better change will require significant rework of many elements and how external services are used. we should do it, but just not immediately.

satra:
@mvandenburgh - i think this PR can be closed right? we have resolved this with environment vars?

mvandenburgh:
> @mvandenburgh - i think this PR can be closed right? we have resolved this with environment vars?

Sorry, missed this. Yes, the usage of `DANDI_ALLOW_LOCALHOST_URLS` along with the new environment vars for the client/server URLs has resolved this.

PR #99, opened 2021-11-29
Test against Python 3.10

jwodder:


yarikoptic:
It was so green, should be merged. Thank you @jwodder

PR #100, opened 2021-12-06
Fully type-annotate and type-check everything

jwodder:
@satra At the moment, the following nontrivial issues are causing problems with the type checking:

* The dynamic creation of enum types from dicts does not play well with mypy at all.
* mypy does not like these two lines, and I'm not even convinced they work as intended anyway:

https://github.com/dandi/dandischema/blob/62067fbb0184d51f2dd9ee415cf014f74249cd78/dandischema/models.py#L137-L138

* [The definition of `wasGeneratedBy` in `Dandiset`](https://github.com/dandi/dandischema/blob/62067fbb0184d51f2dd9ee415cf014f74249cd78/dandischema/models.py#L1031) — which declares the field as a `Optional[List[Project]]` — is incompatible with [the definition of the field in the supertype](https://github.com/dandi/dandischema/blob/62067fbb0184d51f2dd9ee415cf014f74249cd78/dandischema/models.py#L960) — which declares it as `Optional[List[Activity]]`.

* [The definition of `url` in `PublishedDandiset `](https://github.com/dandi/dandischema/blob/62067fbb0184d51f2dd9ee415cf014f74249cd78/dandischema/models.py#L1161) — which declares the field as a `str` — is incompatible with [the definition of the field in the supertype](https://github.com/dandi/dandischema/blob/62067fbb0184d51f2dd9ee415cf014f74249cd78/dandischema/models.py#L949) — which declares it as `Optional[HttpUrl]`.  If `PublishedDandiset.url` is changed to an `HttpUrl`, pydantic then complains about the field's `regex` parameter not applying.
* mypy doesn't like the overriding of `HttpUrl` based on an environment variable.


dandischema/digests/dandietag.py
```diff
@@ -122,7 +122,8 @@ def get_next_part(self) -> Optional[Part]:
             return None
 
     def get_part_etag(self, p: Part) -> Optional[str]:
-        return self._md5_digests[p.number - 1].hex()
+        d = self._md5_digests[p.number - 1]
+        return d.hex() if d is not None else None
```
yarikoptic, line 126: this seems to also provide a functionality fix, so we do not `.hex` a None (for a not yet processed part).  I am just surprised that this function is not tests covered at all. Probably it was covered  whenever this code lived in dandi-cli (IIRC) since we do upload.  I wonder, if while at it, a dedicated unittest should be added?

jwodder: @yarikoptic
* Note that the code still also lives in dandi-cli, as it hasn't been removed yet.
* Test added.

yarikoptic: > Note that the code still also lives in dandi-cli, as it hasn't been removed yet.

oh -- I haven't realized that.  Could you please remove it and RF to use dandischema's (we depend on dandischema anyways)

> Test added

Thank you!

dandischema/datacite.py
```diff
@@ -134,7 +134,7 @@ def to_datacite(
                 NAME_PATTERN, contr_el.name
             ).pop()
 
-            if getattr(contr_el, "affiliation"):
+            if hasattr(contr_el, "affiliation") and contr_el.affiliation is not None:
```
satra, line 137: can this not be replaced by the following pattern?

```suggestion
            if getattr(contr_el, "affiliation", None):
```

dandischema/exceptions.py
```diff
@@ -7,20 +8,14 @@ class ValidationError(ValueError):
...
-    """Validation errors were detected by jsonschema.
+    """Validation errors were detected by jsonschema"""
 
-    All errors are contained in the .args[0] of the exception instance.
-    """
-    @property
-    def errors(self) -> List[jsonschema.exceptions.ValidationError]:
-        return self.args[0]
+    def __init__(self, errors: List[jsonschema.exceptions.ValidationError]) -> None:
+        self.errors = errors
```
satra, line 14: doesn't this change the semantics from something that is readonly to something that is editable?

satra (commented):
some comments. i'll address your comments in a separate thread.

satra:
> The dynamic creation of enum types from dicts does not play well with mypy at all.

mypy doesn't tend to like anything dynamic. if you have a suggestion let me know. we need more info around those enums, so turning them into enums to start with will lose that capability.

> mypy does not like these two lines, and I'm not even convinced they work as intended anyway:
dandischema/dandischema/models.py

this was added by @yarikoptic instead of manually writing schemaKey in every class. in terms of functionality i believe it works, can you suggested why you think it doesn't work?

> The definition of wasGeneratedBy in Dandiset — which declares the field as a Optional[List[Project]] — is incompatible with the definition of the field in the supertype — which declares it as Optional[List[Activity]].

a `Project` is a subclass of `Activity`, which should be fine from a class perspective.

> The definition of url in PublishedDandiset — which declares the field as a str — is incompatible with the definition of the field in the supertype — which declares it as Optional[HttpUrl]. If PublishedDandiset.url is changed to an HttpUrl, pydantic then complains about the field's regex parameter not applying.

pydantic doesn't allow regex's for url patterns. and we need to be more specific as to what that published url looks like. this is why it was changed to a `str`.

> mypy doesn't like the overriding of HttpUrl based on an environment variable.

i don't have a good suggestion, but we will likely continue with dynamic configurations.

dandischema/datacite.py
jwodder: If using `getattr`, mypy can't tell whether `contr_el.affiliation` is `None` or not, so it'll warn when it's used in a non-`None` way inside the block.

dandischema/exceptions.py
jwodder: Yes.  Is that a problem?

jwodder:
> > The dynamic creation of enum types from dicts does not play well with mypy at all.
> 
> mypy doesn't tend to like anything dynamic. if you have a suggestion let me know. we need more info around those enums, so turning them into enums to start with will lose that capability.

Need more info for what?  In the current state, I don't see what is gained from the dynamic construction.

> > mypy does not like these two lines, and I'm not even convinced they work as intended anyway:
> > dandischema/dandischema/models.py
> 
> this was added by @yarikoptic instead of manually writing schemaKey in every class. in terms of functionality i believe it works, can you suggested why you think it doesn't work?

While the `Field` assignment works, the `Literal` annotation does not, as annotating a dict key doesn't do anything; [quoting the Python docs:](https://docs.python.org/3/reference/simple_stmts.html#annotated-assignment-statements) "For expressions as assignment targets, the annotations are evaluated if in class or module scope, but not stored."

> > The definition of wasGeneratedBy in Dandiset — which declares the field as a Optional[List[Project]] — is incompatible with the definition of the field in the supertype — which declares it as Optional[List[Activity]].
> 
> a `Project` is a subclass of `Activity`, which should be fine from a class perspective.

And yet mypy complains.  I think it has something to do with the covariance/contravariance of `List`.

> > The definition of url in PublishedDandiset — which declares the field as a str — is incompatible with the definition of the field in the supertype — which declares it as Optional[HttpUrl]. If PublishedDandiset.url is changed to an HttpUrl, pydantic then complains about the field's regex parameter not applying.
> 
> pydantic doesn't allow regex's for url patterns. and we need to be more specific as to what that published url looks like. this is why it was changed to a `str`.

You could use a validator instead of a regex.

> > mypy doesn't like the overriding of HttpUrl based on an environment variable.
> 
> i don't have a good suggestion, but we will likely continue with dynamic configurations.

What's wrong with just always using `AnyHttpUrl`?

satra:
> Need more info for what

for json-ld graph constructions. the dictionaries from which the enums are built have additional properties around those values. 

> You could use a validator instead of a regex.

we are validating through multiple routes using both python and json schema validator. regex's also apply to json schema. anyhttpurl is too general and also does not support regex's. the intent here is to be as specific as possible for both routes, and specialize in python only when necessary. regex's therefore support both routes to start with.

jwodder:
@satra

* Is anything currently consuming the dicts for JSON-LD purposes?  What other ways can the data be provided to such consumers?
* #76 and some other issues I've seen around about supporting alternative URLs imply that you may want to drop the regex restriction on `PublishedDandiset.url` soon.  If that happens, could we just change that field and all other URL fields to plain `AnyHttpUrl`?

yarikoptic:
should we resurrect this one? seems has aged enough, would need rebase etc...

satra:
for AnyHttpUrl see this open issue: https://github.com/samuelcolvin/pydantic/issues/3143 (so just shifting to anyhttpurl won't solve the mypy issue). also we do have something specific that makes it httpurl, it seems forcing something to be more general just for static validation seems to be counter to the idea of specificity of elements.

i welcome suggestions, but we should not move away from our overall goals of maintaining both a json schema that can do a fair bit of validating and pydantic, which can add additional validators. the dandi web app uses the json schema for validation, so we can't simply move away from those elements.

jwodder:
I'm not saying we should move away from the JSON Schema, but I would like to know how the current enum metadata is consumed & used so that a static alternative can be devised.  The dynamic enum generation is currently the biggest problem with getting mypy to be happy with this library, and I have yet to see a real reason why it's done the way it is.

yarikoptic:
I would like to revive this development so we could get dandischema annotated and then dandi-cli.
It seems that some (not sure if applicable) workaround for pydantic/AnyHttpUrl issue was suggested, see https://github.com/pydantic/pydantic/issues/3143#issuecomment-1139527466 - could it be applied here?
@jwodder, would you be kind to merge in master (not sure if comments would survive with rebase)?

jwodder:
@yarikoptic

> would you be kind to merge in master

You mean rebase this PR against master, not merge this PR into master, right?

yarikoptic:
> @yarikoptic
> 
> > would you be kind to merge in master
> 
> You mean rebase this PR against master, not merge this PR into master, right?

correct. I guess I should have said "merge master in [this branch]" ;-)

jwodder:
@yarikoptic Rebased.

yarikoptic:
@satra - could you please re-review this PR, and follow up on existing discussions. It would be nice to finish up this PR and get it merged so the work doesn't get wasted and we get better type annotations checking through out.

yarikoptic:
Meanwhile: I've rebased  and force-pushed to get a fresh state. Locally I get "Found 132 errors in 7 files (checked 21 source files)", so delaying finishing it up causes the growth of annotations to "catch up on".  I can promise keeping an eye on this PR to get it over the finish line if @jwodder makes another push to add missing annotations.

dandischema/metadata.py
```diff
@@ -104,7 +104,7 @@ def publish_model_schemata(releasedir: str) -> Path:
     return vdir
 
 
-def _validate_obj_json(data, schema, missing_ok=False):
+def _validate_obj_json(data: Any, schema: Any, missing_ok: bool = False) -> None:
```
satra, line 107:
```suggestion
def _validate_obj_json(data: dict, schema: Any, missing_ok: bool = False) -> None:
```

satra:
@jwodder - there are several intentional tests for checking validation, which the type checker seems to pick up on as invalid.

jwodder:
@satra Are you referring to the errors of the form `Missing named argument "id" for
"CommonModel"` in `test_models.py`?  The `id` field is annotated as `Optional[str]`, so pydantic lets it be omitted despite no explicit default being given, but the presence of `= Field(... args that are not "default" ...)` causes mypy to think that the field is required.  I'm not sure whether this is a bug in mypy or pydantic, but the easy solution would be to add `default=None` to `id`'s `Field()` call.

satra:
@jwodder - yup that's the one.

jwodder:
I'm currently trying to address the flawed attempt at defining `schemaKey` by instead hardcoding it separately for each & every class, but this brings to the forefront the fact that the current structure violates the Liskov Substitution Principle: if all instances of `Contributor` have to have `"schemaKey": "Contributor"`, then this must hold for subclasses as well, yet we're also requiring the `Contributor` subclass `Person` to have `"schemaKey": "Person"` for all instances (and likewise for all other parent-child pairs).

@satra Would it break anything if the type annotations for `schemaKey` in parent classes listed all child classes as well?  That's the only way I currently see to resolve this without abandoning inheritance entirely.

jwodder:
@satra Problem: In this code:

https://github.com/dandi/dandi-schema/blob/f72275161ac1102d94e26d56cef6e15b68c68d50/dandischema/models.py#L970-L971

you're trying to use `DANDI_INSTANCE_URL` as the default for an `HttpUrl` field, yet `DANDI_INSTANCE_URL` will be `None` if the `DJANGO_DANDI_WEB_APP_URL` envvar is not set.  Should `repository` become `Optional`, or should `DANDI_INSTANCE_URL` be set to some default if the envvar is not set?

jwodder:
@satra Now, the enumephant in the room: The enums need to be defined statically instead of via dynamic construction from dicts if we don't want mypy to vomit every time it sees this package.  Is there any desire to keep the various metadata in the dicts around somehow?  As far as I can tell, it's currently completely unused.

satra:
@jwodder - on the enum front, i think we will be ok to strip them, but at this stage i don't remember all the implications. the only place i can think of is here: https://github.com/dandi/dandi-schema/blob/master/dandischema/metadata.py#L57

technically as long as this PR doesn't change the schemata release we should be ok.

jwodder:
@satra Based on lines 59-60, I don't believe that code cares about enums.

yarikoptic:
Yarik-bot summarizes the progress/distance to the finish line: "Found 22 errors in 3 files (checked 20 source files)"

jwodder:
@satra Ping.  Please respond to my questions about [`schemaKey`](https://github.com/dandi/dandi-schema/pull/100#issuecomment-1434791488) and [`repository`](https://github.com/dandi/dandi-schema/pull/100#issuecomment-1434810951).

satra:
@jwodder 

> Would it break anything if the type annotations for `schemaKey` in parent classes listed all child classes as well?

we are using `schemaKey` as a constant associated with a class reflecting it's name. this allows an object to be appropriately initialized from a dictionary when the only attributes overlap with another class. Do you mean `Contributor` would have the `schemaKey` literal `Contributor.Person.Organization` or would you turn the `Literal` to a list?

Essentially isn't the same thing as a hidden variable name like `__class__`, which takes on different values independent of inheritance?

> Should `repository` become `Optional`, or should `DANDI_INSTANCE_URL` be set to some default if the envvar is not set?

it could be set to something like `localhost` if the envvar is not set.

jwodder:
@jwodder 

> Do you mean `Contributor` would have the `schemaKey` literal `Contributor.Person.Organization` or would you turn the `Literal` to a list?

I mean that the `Literal` annotation would allow multiple options, here `Literal["Contributor", "Person", "Organization"]`.

> Essentially isn't the same thing as a hidden variable name like `__class__`, which takes on different values independent of inheritance?

`__class__` is magic and is not a model to follow if you want proper inheritance.

> it could be set to something like `localhost` if the envvar is not set.

That would require `AnyHttpUrl` to be used instead of `HttpUrl`, which currently only happens if the envvar `DANDI_ALLOW_LOCALHOST_URLS` is set.

jwodder:
@satra How about we just always use `AnyHttpUrl` regardless of the environment?  That would make things simpler.

satra:
> How about we just always use AnyHttpUrl regardless of the environment? That would make things simpler.

let's do it.

jwodder:
@satra New problem: `url` is declared as `Optional[AnyHttpUrl]` in `CommonModel`, yet `PublishedDandiset.url` is declared as a `str` with a `regex` constraint, and changing it to `AnyHttpUrl` produces an error because the constraint no longer does anything.

satra:
it's because we wanted a simple regex-based validator. we can override with an attribute specific validator in that class.

jwodder:
This PR is now ready for review.

Regarding the failing "Test against dandi-cli" tests, I see two kinds of test failures:

- Some `test_prepare_metadata` tests are failing because the generated metadata is now expected to contain a `repository: "https://localhost/"` field
- The other failing tests are due to errors trying to connect to www.ontobee.org.

jwodder:
@satra Ping.

yarikoptic:
re - `repository` and original question/decision:

> > Should `repository` become `Optional`, or should `DANDI_INSTANCE_URL` be set to some default if the envvar is not set?
> 
> it could be set to something like `localhost` if the envvar is not set.

*interestingly*,

<details>
<summary>although many (96) dandisets have `repository` mentioned among assets metadata records</summary> 

```shell
(dandisets) dandi@drogon:/mnt/backup/dandi/dandisets$ grep -l repository */.dandi/assets.json   | xargs ls -l | nl
     1  -rw-r--r-- 1 dandi dandi   325021 Mar  2 01:50 000003/.dandi/assets.json
     2  -rw-r--r-- 1 dandi dandi   250075 Mar  2 01:50 000004/.dandi/assets.json
     3  -rw-r--r-- 1 dandi dandi   456958 Mar  2 01:50 000005/.dandi/assets.json
     4  -rw-r--r-- 1 dandi dandi   164554 Mar  2 01:50 000006/.dandi/assets.json
     5  -rw-r--r-- 1 dandi dandi   167003 Mar  2 01:50 000007/.dandi/assets.json
     6  -rw-r--r-- 1 dandi dandi   548399 Mar  2 01:50 000009/.dandi/assets.json
     7  -rw-r--r-- 1 dandi dandi   512965 Mar  2 01:50 000010/.dandi/assets.json
     8  -rw-r--r-- 1 dandi dandi   306587 Mar  2 01:50 000011/.dandi/assets.json
     9  -rw-r--r-- 1 dandi dandi   677951 Mar  2 01:50 000012/.dandi/assets.json
    10  -rw-r--r-- 1 dandi dandi   165751 Mar  2 01:50 000013/.dandi/assets.json
    11  -rw-r--r-- 1 dandi dandi   627372 Mar  2 01:50 000015/.dandi/assets.json
    12  -rw-r--r-- 1 dandi dandi   315254 Mar  2 01:50 000016/.dandi/assets.json
    13  -rw-r--r-- 1 dandi dandi   118042 Mar  2 01:50 000017/.dandi/assets.json
    14  -rw-r--r-- 1 dandi dandi    88286 Mar  2 01:50 000019/.dandi/assets.json
    15  -rw-r--r-- 1 dandi dandi 12914541 Mar  2 01:50 000020/.dandi/assets.json
    16  -rw-r--r-- 1 dandi dandi   610545 Mar  2 01:50 000021/.dandi/assets.json
    17  -rw-r--r-- 1 dandi dandi   481904 Mar  2 01:50 000022/.dandi/assets.json
    18  -rw-r--r-- 1 dandi dandi   909895 Mar  2 01:50 000023/.dandi/assets.json
    19  -rw-r--r-- 1 dandi dandi     2918 Mar  2 01:50 000025/.dandi/assets.json
    20  -rw-r--r-- 1 dandi dandi 95494014 Mar  2 11:04 000026/.dandi/assets.json
    21  -rw-r--r-- 1 dandi dandi     2553 Mar  2 01:50 000027/.dandi/assets.json
    22  -rw-r--r-- 1 dandi dandi     6781 Mar  2 01:50 000028/.dandi/assets.json
    23  -rw-r--r-- 1 dandi dandi    14481 Mar  2 01:50 000029/.dandi/assets.json
    24  -rw-r--r-- 1 dandi dandi    15811 Mar  2 01:51 000034/.dandi/assets.json
    25  -rw-r--r-- 1 dandi dandi   598609 Mar  2 01:51 000035/.dandi/assets.json
    26  -rw-r--r-- 1 dandi dandi   127632 Mar  2 01:51 000036/.dandi/assets.json
    27  -rw-r--r-- 1 dandi dandi   348131 Mar  2 01:52 000039/.dandi/assets.json
    28  -rw-r--r-- 1 dandi dandi    68648 Mar  2 01:52 000041/.dandi/assets.json
    29  -rw-r--r-- 1 dandi dandi   188529 Mar  2 01:52 000043/.dandi/assets.json
    30  -rw-r--r-- 1 dandi dandi    26140 Mar  2 01:52 000044/.dandi/assets.json
    31  -rw-r--r-- 1 dandi dandi 23144712 Mar  2 01:54 000045/.dandi/assets.json
    32  -rw-r--r-- 1 dandi dandi     2537 Mar  2 01:53 000048/.dandi/assets.json
    33  -rw-r--r-- 1 dandi dandi   260685 Mar  2 01:54 000049/.dandi/assets.json
    34  -rw-r--r-- 1 dandi dandi   160832 Mar  2 01:54 000050/.dandi/assets.json
    35  -rw-r--r-- 1 dandi dandi     1878 Mar  2 01:54 000051/.dandi/assets.json
    36  -rw-r--r-- 1 dandi dandi   911288 Mar  2 01:54 000052/.dandi/assets.json
    37  -rw-r--r-- 1 dandi dandi  1119664 Mar  2 01:54 000053/.dandi/assets.json
    38  -rw-r--r-- 1 dandi dandi   265174 Mar  2 01:54 000054/.dandi/assets.json
    39  -rw-r--r-- 1 dandi dandi   168730 Mar  2 01:54 000055/.dandi/assets.json
    40  -rw-r--r-- 1 dandi dandi   168872 Mar  2 01:55 000056/.dandi/assets.json
    41  -rw-r--r-- 1 dandi dandi    29168 Mar  2 01:55 000058/.dandi/assets.json
    42  -rw-r--r-- 1 dandi dandi   192063 Mar  2 01:55 000059/.dandi/assets.json
    43  -rw-r--r-- 1 dandi dandi   259609 Mar  2 01:55 000060/.dandi/assets.json
    44  -rw-r--r-- 1 dandi dandi   165866 Mar  2 01:55 000061/.dandi/assets.json
    45  -rw-r--r-- 1 dandi dandi     2198 Mar  2 01:55 000064/.dandi/assets.json
    46  -rw-r--r-- 1 dandi dandi     1657 Mar  2 01:55 000065/.dandi/assets.json
    47  -rw-r--r-- 1 dandi dandi     6656 Mar  2 01:55 000066/.dandi/assets.json
    48  -rw-r--r-- 1 dandi dandi   114436 Mar  2 01:55 000067/.dandi/assets.json
    49  -rw-r--r-- 1 dandi dandi     4762 Mar  2 01:55 000068/.dandi/assets.json
    50  -rw-r--r-- 1 dandi dandi     2640 Jan 27  2022 000069/.dandi/assets.json
    51  -rw-r--r-- 1 dandi dandi    23257 Mar  2 01:55 000070/.dandi/assets.json
    52  -rw-r--r-- 1 dandi dandi     3494 Mar  2 01:55 000105/.dandi/assets.json
    53  -rw-r--r-- 1 dandi dandi     1690 Mar  2 01:55 000107/.dandi/assets.json
    54  -rw-r--r-- 1 dandi dandi 11434044 Feb 23 09:34 000108/.dandi/assets.json
    55  -rw-r--r-- 1 dandi dandi  1001388 Mar  2 01:55 000109/.dandi/assets.json
    56  -rw-r--r-- 1 dandi dandi   172201 Mar  2 01:55 000115/.dandi/assets.json
    57  -rw-r--r-- 1 dandi dandi   411942 Mar  2 01:55 000117/.dandi/assets.json
    58  -rw-r--r-- 1 dandi dandi    11652 Mar  2 01:55 000122/.dandi/assets.json
    59  -rw-r--r-- 1 dandi dandi    12752 Mar  2 01:55 000126/.dandi/assets.json
    60  -rw-r--r-- 1 dandi dandi     6888 Mar  2 01:55 000127/.dandi/assets.json
    61  -rw-r--r-- 1 dandi dandi     6476 Mar  2 01:55 000128/.dandi/assets.json
    62  -rw-r--r-- 1 dandi dandi     6112 Mar  2 01:55 000129/.dandi/assets.json
    63  -rw-r--r-- 1 dandi dandi     6624 Mar  2 01:55 000130/.dandi/assets.json
    64  -rw-r--r-- 1 dandi dandi     6481 Mar  2 01:55 000138/.dandi/assets.json
    65  -rw-r--r-- 1 dandi dandi     6487 Mar  2 01:55 000139/.dandi/assets.json
    66  -rw-r--r-- 1 dandi dandi     6479 Mar  2 01:55 000140/.dandi/assets.json
    67  -rw-r--r-- 1 dandi dandi  1915543 Mar  2 01:55 000142/.dandi/assets.json
    68  -rw-r--r-- 1 dandi dandi    83240 Mar  2 01:55 000143/.dandi/assets.json
    69  -rw-r--r-- 1 dandi dandi     4481 Mar  2 01:55 000144/.dandi/assets.json
    70  -rw-r--r-- 1 dandi dandi    29971 Mar  2 01:55 000147/.dandi/assets.json
    71  -rw-r--r-- 1 dandi dandi    13357 Mar  2 01:55 000149/.dandi/assets.json
    72  -rw-r--r-- 1 dandi dandi  2141451 Mar  2 01:56 000165/.dandi/assets.json
    73  -rw-r--r-- 1 dandi dandi    81047 Mar  2 01:55 000166/.dandi/assets.json
    74  -rw-r--r-- 1 dandi dandi    15667 Mar  2 01:55 000167/.dandi/assets.json
    75  -rw-r--r-- 1 dandi dandi   454401 Mar  2 01:56 000168/.dandi/assets.json
    76  -rw-r--r-- 1 dandi dandi   357200 Mar  2 01:56 000173/.dandi/assets.json
    77  -rw-r--r-- 1 dandi dandi     3364 Mar  2 01:56 000206/.dandi/assets.json
    78  -rw-r--r-- 1 dandi dandi    56896 Mar  2 01:56 000207/.dandi/assets.json
    79  -rw-r--r-- 1 dandi dandi   764133 Mar  2 01:56 000209/.dandi/assets.json
    80  -rw-r--r-- 1 dandi dandi  2735742 Mar  2 01:56 000212/.dandi/assets.json
    81  -rw-r--r-- 1 dandi dandi   293264 Mar  2 01:56 000213/.dandi/assets.json
    82  -rw-r--r-- 1 dandi dandi  2924657 Mar  2 01:56 000217/.dandi/assets.json
    83  -rw-r--r-- 1 dandi dandi   429693 Mar  2 01:56 000218/.dandi/assets.json
    84  -rw-r--r-- 1 dandi dandi   178917 Mar  2 01:56 000219/.dandi/assets.json
    85  -rw-r--r-- 1 dandi dandi    95422 Mar  2 01:56 000220/.dandi/assets.json
    86  -rw-r--r-- 1 dandi dandi   888600 Mar  2 01:56 000221/.dandi/assets.json
    87  -rw-r--r-- 1 dandi dandi    71722 Mar  2 01:56 000223/.dandi/assets.json
    88  -rw-r--r-- 1 dandi dandi   263858 Mar  2 01:56 000228/.dandi/assets.json
    89  -rw-r--r-- 1 dandi dandi  9565293 Mar  2 01:56 000231/.dandi/assets.json
    90  -rw-r--r-- 1 dandi dandi   294890 Mar  2 01:56 000232/.dandi/assets.json
    91  -rw-r--r-- 1 dandi dandi  1184124 Mar  2 01:56 000233/.dandi/assets.json
    92  -rw-r--r-- 1 dandi dandi    37404 Mar  2 01:56 000235/.dandi/assets.json
    93  -rw-r--r-- 1 dandi dandi    31704 Mar  2 01:56 000236/.dandi/assets.json
    94  -rw-r--r-- 1 dandi dandi    34586 Mar  2 01:56 000237/.dandi/assets.json
    95  -rw-r--r-- 1 dandi dandi    19630 Mar  2 01:56 000238/.dandi/assets.json
    96  -rw-r--r-- 1 dandi dandi  1917176 Mar  2 01:56 000239/.dandi/assets.json

```
</details>

<details>
<summary>which have a spectrum of schemaVersions</summary> 

```shell
(dandisets) dandi@drogon:/mnt/backup/dandi/dandisets$ grep -l repository */.dandi/assets.json   | xargs ls -l | grep -v '2 ... .. ' | sed -e 's,.* ,,g' | xargs grep -h -A1 schemaVersion | grep '^"0' | sort | uniq -c
   8076 "0.4.4",
    477 "0.5.1",
     65 "0.5.2",
  18784 "0.6.0",
   2911 "0.6.2"
  20123 "0.6.2",
      3 "0.6.3"
  30802 "0.6.3",
```
</details>

<details>
<summary>there is 35 with assets listed and no repository mentioned</summary> 

```shell
(dandisets) dandi@drogon:/mnt/backup/dandi/dandisets$ grep -L repository */.dandi/assets.json   | xargs ls -l | grep -v '2 ... .. ' | nl
     1  -rw-r--r-- 1 dandi dandi 4134879 Mar  2 01:50 000008/.dandi/assets.json
     2  -rw-r--r-- 1 dandi dandi  318311 Mar  2 01:51 000037/.dandi/assets.json
     3  -rw-r--r-- 1 dandi dandi  141404 Mar  2 01:55 000148/.dandi/assets.json
     4  -rw-r--r-- 1 dandi dandi  172636 Mar  2 01:56 000226/.dandi/assets.json
     5  -rw-r--r-- 1 dandi dandi    7700 Mar  2 01:56 000243/.dandi/assets.json
     6  -rw-r--r-- 1 dandi dandi   83110 Mar  2 01:56 000244/.dandi/assets.json
     7  -rw-r--r-- 1 dandi dandi   75174 Mar  2 01:56 000245/.dandi/assets.json
     8  -rw-r--r-- 1 dandi dandi  213757 Mar  2 01:56 000246/.dandi/assets.json
     9  -rw-r--r-- 1 dandi dandi 1959099 Mar  2 01:56 000249/.dandi/assets.json
    10  -rw-r--r-- 1 dandi dandi 1248825 Mar  2 01:56 000251/.dandi/assets.json
    11  -rw-r--r-- 1 dandi dandi  114066 Mar  2 01:56 000288/.dandi/assets.json
    12  -rw-r--r-- 1 dandi dandi   33769 Mar  2 01:56 000292/.dandi/assets.json
    13  -rw-r--r-- 1 dandi dandi  381751 Mar  2 01:56 000293/.dandi/assets.json
    14  -rw-r--r-- 1 dandi dandi    5826 Mar  2 01:56 000294/.dandi/assets.json
    15  -rw-r--r-- 1 dandi dandi   82401 Mar  2 01:56 000295/.dandi/assets.json
    16  -rw-r--r-- 1 dandi dandi 3993803 Mar  2 01:56 000296/.dandi/assets.json
    17  -rw-r--r-- 1 dandi dandi  317763 Mar  2 01:56 000297/.dandi/assets.json
    18  -rw-r--r-- 1 dandi dandi    2889 Mar  2 01:56 000299/.dandi/assets.json
    19  -rw-r--r-- 1 dandi dandi   44695 Mar  2 01:56 000301/.dandi/assets.json
    20  -rw-r--r-- 1 dandi dandi   59523 Mar  2 01:56 000337/.dandi/assets.json
    21  -rw-r--r-- 1 dandi dandi   38987 Mar  2 01:56 000339/.dandi/assets.json
    22  -rw-r--r-- 1 dandi dandi 2199860 Mar  2 01:56 000341/.dandi/assets.json
    23  -rw-r--r-- 1 dandi dandi   26890 Mar  2 01:56 000347/.dandi/assets.json
    24  -rw-r--r-- 1 dandi dandi   37848 Mar  2 01:56 000350/.dandi/assets.json
    25  -rw-r--r-- 1 dandi dandi 1062745 Mar  2 01:56 000351/.dandi/assets.json
    26  -rw-r--r-- 1 dandi dandi  177409 Mar  2 01:56 000362/.dandi/assets.json
    27  -rw-r--r-- 1 dandi dandi    7939 Mar  2 01:56 000397/.dandi/assets.json
    28  -rw-r--r-- 1 dandi dandi   70056 Mar  2 01:57 000402/.dandi/assets.json
    29  -rw-r--r-- 1 dandi dandi   33985 Mar  2 01:57 000404/.dandi/assets.json
    30  -rw-r--r-- 1 dandi dandi  634985 Mar  2 01:57 000405/.dandi/assets.json
    31  -rw-r--r-- 1 dandi dandi 2865647 Mar  2 01:57 000409/.dandi/assets.json
    32  -rw-r--r-- 1 dandi dandi   68345 Mar  2 01:57 000410/.dandi/assets.json
    33  -rw-r--r-- 1 dandi dandi    2704 Mar  2 01:57 000411/.dandi/assets.json
    34  -rw-r--r-- 1 dandi dandi   17645 Mar  2 01:57 000447/.dandi/assets.json
    35  -rw-r--r-- 1 dandi dandi    2479 Mar  2 01:57 000448/.dandi/assets.json

```
</details>

<details>
<summary>
and they all have `schemaVersion` 0.6.3
</summary> 

```
(dandisets) dandi@drogon:/mnt/backup/dandi/dandisets$ grep -L repository */.dandi/assets.json   | xargs ls -l | grep -v '2 ... .. ' | sed -e 's,.* ,,g' | xargs grep -h -A1 schemaVersion | grep '^"0' | sort | uniq -c
   7453 "0.6.3",
```
</details>

so it feels that we just better for consistency make `repository` Optional and be done without affects on `dandi-cli` tests etc. WDYT @satra ?

satra:
> so it feels that we just better for consistency make repository Optional and be done without affects on dandi-cli tests etc. WDYT @satra ?

sounds good to me.

but `repository` should always be there in dandiset metadata, as it should be inserted by the server.

yarikoptic:
> but `repository` should always be there in dandiset metadata, as it should be inserted by the server.

then we might have a bug, filed https://github.com/dandi/dandi-archive/issues/1516 with some prior context I found.

dandischema/datacite.py
yarikoptic: that sounds like a deficiency in `mypy` since logically it cannot be `None` since then `if` would not evaluate to True.  Do you know of the issue may be already in mypy on this?

yarikoptic:
ok, the testing is green now. @satra - it might just need your final review/blessing.

satra (approved):

yarikoptic:
I would vote for  release so we can make us of type annotations in derived projects. Added a label.  Let's see how things break! ;)  Thank you @jwodder for bringing it to the finish line.  Here goes nothing!

yarikoptic:
oh, I just now mentioned that the check for no changes to exported schemata is failing, there is a number of changes so if we are to proceed - we would need to release schema as well. Some changes are

- throughout for many
```
-        "maxLength": 2083,
+        "maxLength": 65536,
```
- added enums
```
+          "enum": [
170
+            "Activity",
171
+            "Project",
172
+            "Session",
173
+            "PublishActivity"
174
+          ],
```
- change to use url format
```
     "url": {
211
       "title": "URL",
212
       "description": "Permalink to the Dandiset.",
213
-      "pattern": "^.*/dandiset/\\d{6}/\\d+\\.\\d+\\.\\d+$",
214
       "readOnly": true,
215
       "nskey": "schema",
216
+      "minLength": 1,
217
+      "maxLength": 1000,
218
+      "format": "uri",
219
       "type": "string"
220
     },
```

so -- what do we do? embrace tune ups and release new schema version, @satra?

satra:
> so -- what do we do? embrace tune ups and release new schema version, @satra?

yes.

yarikoptic:
I boosted only patch to 0.6.4 from 0.6.3.  I think it might be appropriate although there is no guarantee that changes in the schema would be compatible with some uses where 0.6.3 was compatible since I don't know if we used any value not in Enums or some odd urls .   Or do you think @satra we better go for 0.7.0 to be entirely "true to ourselves"? ;)

satra:
fine with me.

yarikoptic:
which one is fine with you - 0.6.4 or 0.7.0 @satra?

also any change in schema version results in dandi-cli testing going kaboom since dandi-archive is not accepting future/unknown version of the schema, so I guess we nohow tune up the docker compose setup while testing dandischema to use this particular dandischema "version" @jwodder ?

satra:
i should have been clearer. for this `0.6.4` is fine. there are probably bigger changes coming, for which we will want a bump to `0.7.0`. (btw, dandischema will probably go through a significant overhaul as we align neuro data models across projects. eta: 3 months or so)

PR #101, opened 2021-12-06
Remove duplicate "requests" dependency

jwodder:
"requests" was declared as both a normal installation requirement and as a member of the "test" extra.  This PR removes the superfluous latter declaration.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -49,7 +49,6 @@ style =
 test =
     pytest
     pytest-cov
-    requests
 all =
     #%(doc)s
     %(style)s


yarikoptic:
not sure why dandi-cli tests fail with e.g.
```
E                   dandi.exceptions.FailedToConnectError: Response for getting https://identifiers.org/dandi:000027/0.210831.2033 to redirect returned 400. Please verify that it is a URL related to dandiarchive and supported by dandi client
```
but unrelated. Let's proceed

PR #105, opened 2021-12-31
fix: allows aggregation after discarding path extensions

### LINKED ISSUE #104: counting subjects returns erroneous aggregation
@yarikoptic replicated based on staging metadata. the reason this is erroneous is because the code assumed a bids like filename which meant that there was always going to be a `_<suffix>.<ext>` in the file. the dandiset in question does not have that and is considered valid from a CLI perspective. so i suspect the CLI needs a change as well.

i'll release a basic fix in the dandischema code in a sec.

```
{'schemaKey': 'AssetsSummary',
 'numberOfBytes': 175662888,
 'numberOfFiles': 24,
 'numberOfSubjects': 48,
 'dataStandard': [{'schemaKey': 'StandardsType',
   'identifier': 'RRID:SCR_015242',
   'name': 'Neurodata Without Borders (NWB)'}],
 'approach': [],
 'measurementTechnique': [{'schemaKey': 'MeasurementTechniqueType',
   'name': 'analytical technique'}],
 'variableMeasured': ['ProcessingModule'],
 'species': [{'schemaKey': 'SpeciesType',
   'identifier': 'http://purl.obolibrary.org/obo/NCBITaxon_7227',
   'name': 'Common fruit fly'}]}
```

```
import requests
def _fetch(url):
    querystring = {"page_size": "100"}
    headers = {
    "Accept": "application/json",
    "Content-Type": "application/json"
    }
    resp = requests.request("GET", url, headers=headers,
                            params=querystring).json()
    results = resp["results"]
    while resp["next"]:
        resp = requests.request("GET", resp["next"], headers=headers).json()
        results.extend(resp["results"])
    assert len(results) == resp["count"]
    return results
assets = _fetch("https://api-staging.dandiarchive.org/api/dandisets/100978/versions/draft/assets/")
meta = [requests.get(f"https://api-staging.dandiarchive.org/api/dandisets/100978/versions/draft/assets/{asset['asset_id']}", headers={"Accept": "application/json",}).json() for asset in assets]
aggregate_assets_summary(meta)
```

satra:
closes #104

### DIFF
--- a/dandischema/metadata.py
+++ b/dandischema/metadata.py
@@ -305,7 +305,7 @@ def _add_asset_to_stats(assetmeta: Dict[str, Any], stats: _stats_type) -> None:
             stats = _get_samples(value, stats, hierarchy)
             break
 
-    for part in Path(assetmeta["path"]).name.split("_"):
+    for part in Path(assetmeta["path"]).name.split(".")[0].split("_"):
         if part.startswith("sub-"):
             subject = part.replace("sub-", "")
             if subject not in stats["subjects"]:


satra:
the change in schema is because pydantic fixed something in their latest release. `Any` allows None so the key `value` in `PropertyValue` is no longer required. thinking of a way of fixing this without creating a new schema release.

satra:
@yarikoptic - i can do a minor release of the schema - but this seems like this is not a schema change, since the order of keys in the `required` field does not matter in jsonschema.

djarecka (approved):

yarikoptic:
Thank you @satra for filing that issue and fixing it -- I forgot about it.
It would've been great if a test added which would also help to understand the purpose of the fix later on.
Looking at the diff I do not get what "order" effect you are talking about above.

satra:
> Looking at the diff I do not get what "order" effect you are talking about above.

sorry not diff, but the test that failed when schema is generated. originally i had put this as a release. so the required fields in the json-schema change their order. this leads to the schemata workflow determining that schema version needs to be updated. i couldn't find an easy schema diff library that would determine that a change in the order of elements in the required field is not a difference in schema. for now i will update the version in PR #97, which does change the schema, but something to perhaps consider in the future.

PR #109, opened 2022-01-12
Zarr checksum calculation

dchiquito:
Copy over the non-django-specific code from https://github.com/dandi/dandi-api/blob/master/dandiapi/api/zarr_checksums.py and associated tests.

There is more code in [zarr_checksums.py](https://github.com/dandi/dandi-api/blob/master/dandiapi/api/zarr_checksums.py) that deals with adding/updating/removing batches of checksums to/from the zarr archive, and also with writing the files to S3, but that didn't seem pertinent to the CLI, and it relied heavily on the Django model classes.

The CLI will need to do the legwork of walking the zarr file to feed the checksum algorithm.

### DIFF
--- a/dandischema/digests/zarr.py
+++ b/dandischema/digests/zarr.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import hashlib
-from typing import List, Optional
+from typing import Dict, List, Optional
 
 import pydantic
 
@@ -119,6 +119,7 @@ class ZarrJSONChecksumSerializer:
 
         This method wraps aggregate_checksum and should not be overridden.
         """
+        print(files)
         if checksums is None:
             checksums = ZarrChecksums(
                 files=files if files is not None else [],
@@ -134,3 +135,14 @@ class ZarrJSONChecksumSerializer:
 # S3. However, an empty zarr file still needs to have a checksum, even if it has no checksum file.
 # For convenience, we define this constant as the "null" checksum.
 EMPTY_CHECKSUM = ZarrJSONChecksumSerializer().generate_listing(ZarrChecksums()).md5
+
+
+def get_checksum(files: Dict[str, str], directories: Dict[str, str]) -> str:
+    """Calculate the checksum of a directory."""
+    checksum_listing = ZarrJSONChecksumSerializer().generate_listing(
+        files=sorted([ZarrChecksum(md5=md5, path=path) for path, md5 in files.items()]),
+        directories=sorted(
+            [ZarrChecksum(md5=md5, path=path) for path, md5 in directories.items()]
+        ),
+    )
+    return checksum_listing.md5
--- a/dandischema/digests/zarr.py
+++ b/dandischema/digests/zarr.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from functools import total_ordering
 import hashlib
 from typing import Dict, List, Optional
 
@@ -10,6 +11,7 @@ import pydantic
 ENCODING_KWARGS = {"separators": (",", ":")}
 
 
+@total_ordering
 class ZarrChecksum(pydantic.BaseModel):
     """
     A checksum for a single file/directory in a zarr file.
--- a/dandischema/digests/zarr.py
+++ b/dandischema/digests/zarr.py
@@ -42,7 +42,7 @@ class ZarrChecksums(pydantic.BaseModel):
         return self.files == [] and self.directories == []
 
     def _index(self, checksums: List[ZarrChecksum], checksum: ZarrChecksum):
-        # O(n) performance, consider an ordered dict for optimization
+        # O(n) performance, consider using the bisect module or an ordered dict for optimization
         for i in range(0, len(checksums)):
             if checksums[i].path == checksum.path:
                 return i
--- a/dandischema/digests/zarr.py
+++ b/dandischema/digests/zarr.py
@@ -121,11 +121,10 @@ class ZarrJSONChecksumSerializer:
 
         This method wraps aggregate_checksum and should not be overridden.
         """
-        print(files)
         if checksums is None:
             checksums = ZarrChecksums(
-                files=files if files is not None else [],
-                directories=directories if directories is not None else [],
+                files=sorted(files) if files is not None else [],
+                directories=sorted(directories) if directories is not None else [],
             )
         return ZarrChecksumListing(
             checksums=checksums,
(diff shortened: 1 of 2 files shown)


jwodder:
@dchiquito Could you show an example code that uses this API to calculate the checksum of an arbitrary directory tree?

dchiquito:
@jwodder 
```
import hashlib
import os
import sys

from dandischema.digests.zarr import ZarrChecksum, ZarrChecksumListing, ZarrJSONChecksumSerializer, ZarrChecksums, EMPTY_CHECKSUM

def get_file_checksum(path: str):
    """Generate a ZarrChecksum for the given file."""
    hasher = hashlib.md5()
    with open(path, 'rb') as f:
        hasher.update(f.read())
    return ZarrChecksum(path=path, md5=hasher.hexdigest())

def get_directory_checksum(path: str):
    files = []
    directories = []
    with os.scandir(path) as it:
        for entry in it:
            if entry.is_file():
                files.append(get_file_checksum(entry.path))
            else:
                directories.append(get_directory_checksum(entry.path))
    checksum_listing = ZarrJSONChecksumSerializer().generate_listing(files=files, directories=directories)
    return ZarrChecksum(path=path, md5=checksum_listing.md5)

if __name__ == '__main__':
    if len(sys.argv) < 2:
        print(EMPTY_CHECKSUM)
    else:
        path = sys.argv[1]
        if os.path.isfile(path):
            print(get_file_checksum(path).md5)
        else:
            print(get_directory_checksum(path).md5)
```

jwodder:
@dchiquito I'm not a fan of this API; it seems to shove off too much bookkeeping and class instantiation onto the API user.  (My dream would be a single function that takes a dict that maps each filepath in the tree to its MD5 hash and returns the final checksum, but that would ignore empty directories.)  How about only exposing a single `mkchecksum(directories: Dict[str, str], files: Dict[str, str]) -> str` function that takes dicts mapping the paths of the directories & files in a single directory to their checksums/MD5 hashes and returns the checksum?  This would turn your example into (with fixes for path handling):

```python
import hashlib
import os
import sys

from dandischema.digests.zarr import mkchecksum

def get_file_checksum(path: str):
    """Generate a md5 hash for the given file."""
    hasher = hashlib.md5()
    with open(path, 'rb') as f:
        hasher.update(f.read())
    return hasher.hexdigest()

def get_directory_checksum(path: str, basepath: str):
    files = {}
    directories = {}
    with os.scandir(path) as it:
        for entry in it:
            p = Path(entry.path).relative_to(basepath).as_posix()
            if entry.is_file():
                files[p] = get_file_checksum(entry.path)
            else:
                directories[p] = get_directory_checksum(entry.path, basepath)
    return mkchecksum(directories, files)

if __name__ == '__main__':
    if len(sys.argv) < 2:
        print(EMPTY_CHECKSUM)
    else:
        path = sys.argv[1]
        if os.path.isfile(path):
            print(get_file_checksum(path))
        else:
            print(get_directory_checksum(path, path))
```

dchiquito:
@jwodder The API looks the way it does because it was built to support [this](https://github.com/dandi/dandi-api/blob/master/dandiapi/api/zarr_checksums.py#L137-L342) body of code that updates existing checksum files. The bits I included here are the basic serialization classes and the serializer which calculates the checksums.

I initially wanted to do something with dicts as you suggested, but it made adding/removing entries more complicated, and dicts are much more difficult to sort/traverse in order. Using lists of `ZarrChecksums` makes the rest of my code much more legible.

I think your proposed `mkchecksum(directories: Dict[str, str], files: Dict[str, str]) -> str` could easily use the existing API under the hood. How about including it as an extra helper?

dchiquito:
@jwodder I added the helper in https://github.com/dandi/dandischema/pull/109/commits/552b41da569c8e522fb8a9f409e5a53724455dc4. Does that work for you?

dandischema/digests/zarr.py
```diff
@@ -0,0 +1,148 @@
...
+    A checksum for a single file/directory in a zarr file.
+
+    Every file and directory in a zarr archive has a path and a MD5 hash.
+    """
+
+    md5: str
+    path: str
+
+    # To make ZarrChecksums sortable
+    def __lt__(self, other: ZarrChecksum):
```
jwodder, line 24: You should decorate the class with [`@functools.total_ordering`](https://docs.python.org/3/library/functools.html#functools.total_ordering) for the remaining comparison operators.

dandischema/digests/zarr.py
```diff
@@ -0,0 +1,148 @@
...
+    """
+
+    directories: List[ZarrChecksum] = pydantic.Field(default_factory=list)
+    files: List[ZarrChecksum] = pydantic.Field(default_factory=list)
+
+    @property
+    def is_empty(self):
+        return self.files == [] and self.directories == []
+
+    def _index(self, checksums: List[ZarrChecksum], checksum: ZarrChecksum):
```
jwodder, line 42: I suggest you consider using the [`bisect`](https://docs.python.org/3/library/bisect.html) module.

dandischema/digests/zarr.py
```diff
@@ -0,0 +1,148 @@
...
+        """
+        Generate a new ZarrChecksumListing from the given checksums.
+
+        This method wraps aggregate_checksum and should not be overridden.
+        """
+        print(files)
+        if checksums is None:
+            checksums = ZarrChecksums(
+                files=files if files is not None else [],
+                directories=directories if directories is not None else [],
```
jwodder, line 126: I think it would be safer to sort `files` and `directories` here rather than doing it in `get_checksum()` or expecting the caller to do it.

jwodder:
@dchiquito Yes, that works.

jwodder:
@satra Could this PR be tagged with "release"?  Having this functionality available to dandi-cli would be very useful.

dandischema/digests/zarr.py
```diff
@@ -0,0 +1,148 @@
...
+            except ValueError:
+                self.directories.append(new_checksum)
+        self.directories = sorted(self.directories)
+
+    def remove_checksums(self, paths: List[str]):
+        """Remove a list of paths from the listing."""
+        self.files = sorted(
+            filter(lambda checksum: checksum.path not in paths, self.files)
+        )
+        self.directories = sorted(
```
yarikoptic, line 73: didn't check in details what tests do, but it seems that all above functionality in ZarrChecksums is not tested at all! Could some basic test(s) be added to help ensure no surprises? (ideally on some case where there is a known target checksum(s))

dandischema/digests/zarr.py
```diff
@@ -0,0 +1,148 @@
...
+        checksums: Optional[ZarrChecksums] = None,
+        files: Optional[List[ZarrChecksum]] = None,
+        directories: Optional[List[ZarrChecksum]] = None,
+    ) -> ZarrChecksumListing:
+        """
+        Generate a new ZarrChecksumListing from the given checksums.
+
+        This method wraps aggregate_checksum and should not be overridden.
+        """
+        print(files)
```
yarikoptic, line 122:
```suggestion
```
I guess it was some leftover

dandischema/digests/zarr.py
dchiquito: Added some tests, `zarr.py` now has 100% coverage

satra:
@jwodder - i'll couple this with a few other changes. there is one change to the schema generation that does not guarantee order of a particular field. i need to change that.

satra:
we will also need to update the asset models to allow for this checksum. let me add that in a separate pr.

PR #111, opened 2022-01-13
adding Orcid ID to the contributors

### LINKED ISSUE #102: add orcid to contributor/creator metadata when available
@djarecka - could you please update the doi metadata generator to add orcid ids for contributor and creator when available. some details here: https://support.datacite.org/docs/can-i-add-orcid-ids-to-the-datacite-metadata-1

djarecka:
- adding `orcidID` to the contributors fields. It's added to `nameIdentifiers` list in `to_datacite` function. 

fixes #102

### DIFF
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -423,6 +423,12 @@ class Contributor(DandiBaseModel):
         description="Identifier associated with a sponsored or gift award.",
         nskey="dandi",
     )
+    orcidID: Optional[Identifier] = Field(
+        None,
+        title="Orcid ID",
+        description="Orcid ID....",
+        nskey="dandi",
+    )
 
 
 class Organization(Contributor):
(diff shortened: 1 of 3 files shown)


djarecka:
I'm not sure why codecov does't like me...

dandischema/models.py
```diff
@@ -423,6 +423,12 @@ class Contributor(DandiBaseModel):
         description="Identifier associated with a sponsored or gift award.",
         nskey="dandi",
     )
+    orcidID: Optional[Identifier] = Field(
+        None,
+        title="Orcid ID",
+        description="Orcid ID....",
+        nskey="dandi",
+    )
```
satra, line 431: we already have orcids as `identifier` for `Person` class.

djarecka: forgot to ask, assumed that you want to have a new field only for ORCID. However `identifier` is also used for ORCID only right now... 
Will change

PR #112, opened 2022-01-14
Various edge case fixes

satra:
- [x] schema generation with sorted sets
- [x] added support for wider name formats
- [x] add validation check for zarr checksum
- [x] enable embargo support

### DIFF
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -1204,8 +1204,8 @@ class PublishedAsset(Asset, Publishable):
         if digest is not None:
             if not re.match(DandiETag.REGEX, digest):
                 raise ValueError("Digest is missing dandi-etag value")
-            digest = values[DigestType.sha2_256]
-            if len(digest) != 64:
+            digest = values.get(DigestType.sha2_256, None)
+            if digest is None or len(digest) != 64:
                 raise ValueError("Digest is missing sha2_256 value")
             if values.get(DigestType.dandi_zarr_checksum, None) is not None:
                 raise ValueError("Digest cannot have both etag and zarr checksums.")
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -190,13 +190,13 @@ class DandiBaseModel(BaseModel, metaclass=DandiBaseModelMetaclass):
         @staticmethod
         def schema_extra(schema: Dict[str, Any], model) -> None:
             if schema["title"] == "PropertyValue":
-                schema["required"] = list(
-                    set(["value"]).union(schema.get("required", []))
+                schema["required"] = sorted(
+                    list(set(["value"]).union(schema.get("required", [])))
                 )
             schema["title"] = name2title(schema["title"])
             if schema["type"] == "object":
-                schema["required"] = list(
-                    set(schema.get("required", [])).union(["schemaKey"])
+                schema["required"] = sorted(
+                    list(set(schema.get("required", [])).union(["schemaKey"]))
                 )
             for prop, value in schema.get("properties", {}).items():
                 if schema["title"] == "Person":
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -1206,7 +1206,7 @@ class PublishedAsset(Asset, Publishable):
             digest = v.get(DigestType.sha2_256)
             if not re.fullmatch(SHA256_PATTERN, digest):
                 raise ValueError(
-                    f"Digest must have an appropriate sha2_256 value. " f"Got {digest}"
+                    f"Digest must have an appropriate sha2_256 value. Got {digest}"
                 )
         return v
 
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -1118,7 +1118,7 @@ class BareAsset(CommonModel):
     def digest_check(cls, values):
         digest = values.get(DigestType.dandi_etag, None)
         if digest is not None:
-            if not re.match(DandiETag.REGEX, digest):
+            if not re.fullmatch(DandiETag.REGEX, digest):
                 raise ValueError(
                     f"Digest must have an appropriate dandi-etag value. "
                     f"Got {values[DigestType.dandi_etag]}"
@@ -1202,7 +1202,7 @@ class PublishedAsset(Asset, Publishable):
     def digest_allhashes(cls, values):
         digest = values.get(DigestType.dandi_etag, None)
         if digest is not None:
-            if not re.match(DandiETag.REGEX, digest):
+            if not re.fullmatch(DandiETag.REGEX, digest):
                 raise ValueError("Digest is missing dandi-etag value")
             digest = values.get(DigestType.sha2_256, None)
             if digest is None or len(digest) != 64:
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -91,7 +91,6 @@ def split_name(name):
 
 if len(AccessTypeDict["@graph"]) > 2:
     AccessTypeDict["@graph"].pop()
-    AccessTypeDict["@graph"].pop()
 AccessType = create_enum(AccessTypeDict)
 AgeReferenceType = create_enum(AgeReferenceTypeDict)
 RoleType = create_enum(RoleTypeDict)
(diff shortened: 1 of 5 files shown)


dandischema/models.py
```diff
@@ -190,13 +190,13 @@ class Config:
         @staticmethod
         def schema_extra(schema: Dict[str, Any], model) -> None:
             if schema["title"] == "PropertyValue":
-                schema["required"] = list(
-                    set(["value"]).union(schema.get("required", []))
+                schema["required"] = sorted(
+                    list(set(["value"]).union(schema.get("required", [])))
```
jwodder, line 194: There's no need to cast to a list before calling `sorted()`.  `sorted()` can take any iterable as input.

dandischema/models.py
```diff
@@ -1116,19 +1116,38 @@ class BareAsset(CommonModel):
...
+                )
+            if values.get(DigestType.dandi_zarr_checksum, None) is not None:
+                raise ValueError("Digest cannot have both etag and zarr checksums.")
+            return values
+        digest = values.get(DigestType.dandi_zarr_checksum, None)
+        if digest is not None:
+            try:
+                if len(digest) != 32:
+                    raise ValueError
+            except KeyError:
```
jwodder, line 1140: This `except KeyError:` block is also unreachable.

dandischema/models.py
```diff
@@ -1116,19 +1116,38 @@ class BareAsset(CommonModel):
...
-            raise ValueError(
-                f"Digest must have an appropriate dandi-etag value. "
-                f"Got {values[DigestType.dandi_etag]}"
-            )
-        return values
+    def digest_check(cls, values):
+        digest = values.get(DigestType.dandi_etag, None)
+        if digest is not None:
+            try:
+                if "-" not in digest or len(digest.split("-")[0]) != 32:
```
jwodder, line 1123: You know there's [a regex for Dandi etags](https://github.com/dandi/dandischema/blob/0099e5f0c3a695db2db8336a13553c9fd0be9ff2/dandischema/digests/dandietag.py#L95), right?  Why don't you test against that?

dandischema/models.py
```diff
@@ -1191,17 +1210,39 @@ class PublishedAsset(Asset, Publishable):
...
+                raise ValueError("Digest is missing dandi-etag or sha256 keys.")
+            if values.get(DigestType.dandi_zarr_checksum, None) is not None:
+                raise ValueError("Digest cannot have both etag and zarr checksums.")
+            return values
+        digest = values.get(DigestType.dandi_zarr_checksum, None)
+        if digest is not None:
+            try:
+                if len(digest) != 32:
+                    raise ValueError
+            except KeyError:
```
jwodder, line 1232: Unreachable.

satra:
@jwodder - thank you.

satra:
@dchiquito @yarikoptic - it may be worthwhile looking at this PR as this changes a few things that impact different parts of the infrastructure.

dandischema/models.py
```diff
@@ -1116,19 +1115,26 @@ class BareAsset(CommonModel):
...
-                raise ValueError
-        except KeyError:
-            raise ValueError("Digest is missing dandi-etag value.")
-        except ValueError:
-            raise ValueError(
-                f"Digest must have an appropriate dandi-etag value. "
-                f"Got {values[DigestType.dandi_etag]}"
-            )
-        return values
+    def digest_check(cls, values):
```
jwodder, line 1118: If you rename `values` to something else and add a separate `values` argument for receiving previously-validated fields, you'll be able to take `encodingFormat` into consideration, thereby letting the code determine precisely which of `dandi_etag` and `dandi_zarr_checksum` is required.

jwodder: No, it doesn't need to be a root validator.  See the `password2` validator [in this page of the pydantic docs](https://pydantic-docs.helpmanual.io/usage/validators/) for an example.

dandischema/models.py
```diff
@@ -606,17 +606,15 @@ class AccessRequirements(DandiBaseModel):
...
     @root_validator
     def open_or_embargoed(cls, values):
         status, embargoed = values.get("status"), values.get("embargoedUntil")
         if status == AccessType.EmbargoedAccess and embargoed is None:
-            raise ValueError("An embargo date. For NIH supported awards, this "
-            "must comply with NIH policy. For other datasets, this will be at "
-            "most a year from the dandiset creation.")
+            raise ValueError(
+                "An embargo end date is required for NIH awards that "
+                "complies with NIH resource sharing policy."
```
yarikoptic, line 615:
```suggestion
                "An embargo end date is required for NIH awards to be "
                "in compliance with NIH resource sharing policy."
```

dandischema/models.py
```diff
@@ -1191,17 +1197,25 @@ class PublishedAsset(Asset, Publishable):
...
     @validator("digest")
-    def digest_bothhashes(cls, values):
-        try:
-            digest = values[DigestType.dandi_etag]
-            if "-" not in digest or len(digest.split("-")[0]) != 32:
+    def digest_allhashes(cls, values):
+        digest = values.get(DigestType.dandi_etag, None)
+        if digest is not None:
+            if not re.fullmatch(DandiETag.REGEX, digest):
                 raise ValueError("Digest is missing dandi-etag value")
```
yarikoptic, line 1204: message seems to be incorrect, 

```suggestion
                raise ValueError(
                    f"Digest must have an appropriate dandi-etag value. "
                    f"Got {digest}")
```
possibly again asking for some `_check_dandi_etag(values)` to be reused?

dandischema/models.py
```diff
@@ -1191,17 +1197,25 @@ class PublishedAsset(Asset, Publishable):
...
-            digest = values[DigestType.sha2_256]
-            if len(digest) != 64:
+            digest = values.get(DigestType.sha2_256, None)
+            if digest is None or len(digest) != 64:
                 raise ValueError("Digest is missing sha2_256 value")
```
yarikoptic, line 1207: again possibly a confusing message since might have a digest but which is not legit.  Might need to be split into two conditions.

yarikoptic (changes requested):
Some code/messages might benefit from tune ups.

dandischema/models.py
```diff
@@ -91,7 +92,6 @@ def split_name(name):
 
 if len(AccessTypeDict["@graph"]) > 2:
     AccessTypeDict["@graph"].pop()
```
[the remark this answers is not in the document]
satra: i'm removing the hack altogether. it was a hack anyway.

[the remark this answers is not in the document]
yarikoptic: If it isn't ok to have more than 1, I would have added assert just in case we don't miss anything

dandischema/models.py
satra: thanks. there are a few changes that has been made based on this which simplifies the checking.

dandischema/models.py
```diff
@@ -1191,17 +1199,16 @@ class PublishedAsset(Asset, Publishable):
...
-            raise ValueError("Digest is missing dandi-etag or sha256 keys.")
-        return values
+    def digest_sha256check(cls, v, values, **kwargs):
+        if values.get("encodingFormat") != "application/x-zarr":
+            if DigestType.sha2_256 not in v:
+                raise ValueError("A non-zarr asset must have a sha2_256.")
+            digest = v.get(DigestType.sha2_256)
+            if not re.fullmatch(SHA256_PATTERN, digest):
+                raise ValueError(
+                    f"Digest must have an appropriate sha2_256 value. " f"Got {digest}"
```
yarikoptic, line 1209:
```suggestion
                    f"Digest must have an appropriate sha2_256 value. Got {digest}"
```

yarikoptic:
Yet to check whole on laptop - metadata record from 0.6.1 should migrate to 0.6.2 no problem right?

satra:
> metadata record from 0.6.1 should migrate to 0.6.2 no problem right?

yup no backward incompatible changes.

PR #116, opened 2022-01-28
Define `PartGenerator` default part size as class member

AlmightyYakob:
This change would allow `PartGenerator` to be more extensible, as if you wanted to use a different part size than 64MB, you could just do the following:

```python
class MyPartGenerator(PartGenerator):
    DEFAULT_PART_SIZE = mb(500)
```

### DIFF
--- a/dandischema/digests/dandietag.py
+++ b/dandischema/digests/dandietag.py
@@ -40,6 +40,8 @@ class PartGenerator:
     MIN_PART_SIZE = mb(5)
     # 5GB is the maximum part size allowed by S3
     MAX_PART_SIZE = gb(5)
+    # 64MB is the default part size
+    DEFAULT_PART_SIZE = mb(64)
 
     @classmethod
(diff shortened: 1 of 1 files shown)


jwodder (approved):

satra:
@AlmightyYakob - let me know if a release is needed. in which case we can set a patch tag and release. otherwise we can merge and couple with other updates for a future release.

AlmightyYakob:
> @AlmightyYakob - let me know if a release is needed. in which case we can set a patch tag and release. otherwise we can merge and couple with other updates for a future release.

If a release is possible that'd be ideal, since this would simplify downstream code in the API. However it's not critical, so if it's inconvenient to make a release now it's not a problem.

AlmightyYakob:
@satra Could a release be made?

yarikoptic:
For the release, ideally this PR should've had release label added before merging. Now it would need an extra PR to kick out a release

PR #117, opened 2022-01-31
Specify part generator to DandiETag class

AlmightyYakob:
This is another small change that accompanies #116.

If this is approved I'd like to add the `release` label to this so that changes in both this and #116 are available.

### DIFF
--- a/dandischema/digests/dandietag.py
+++ b/dandischema/digests/dandietag.py
@@ -97,8 +97,8 @@ class DandiETag:
     REGEX = r"[0-9a-f]{32}-\d{1,5}"
     MAX_STR_LENGTH = 38
 
-    def __init__(self, file_size: int) -> None:
-        self._part_gen: PartGenerator = PartGenerator.for_file_size(file_size)
+    def __init__(self, file_size: int, part_gen=PartGenerator) -> None:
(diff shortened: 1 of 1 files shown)


jwodder (approved):

jwodder:
@AlmightyYakob Ping.

AlmightyYakob:
I can't find anywhere in the current API codebase that this is used. I know it _was_ used at some point, but that code has now either been refactored or removed entirely.

jwodder:
@AlmightyYakob So, would it be OK if this PR was reverted?

AlmightyYakob:
> @AlmightyYakob So, would it be OK if this PR was reverted?

At least from the API side I think it's safe to revert, but I haven't checked any other repos. Although I'd be surprised if it was being made use of anywhere else.

jwodder:
@AlmightyYakob I think for now I'll keep this code, but I've annotated it such that `part_gen` needs to be either `PartGenerator` or a subclass thereof, not just a class that works the same way.

AlmightyYakob:
> @AlmightyYakob I think for now I'll keep this code, but I've annotated it such that `part_gen` needs to be either `PartGenerator` or a subclass thereof, not just a class that works the same way.

That seems perfectly reasonable to me.

PR #120, opened 2022-03-02
Update zarr checksums

dchiquito:
Fixes https://github.com/dandi/dandi-archive/issues/937
Fixes https://github.com/dandi/dandi-archive/issues/925
Fixes https://github.com/dandi/dandi-archive/issues/931

attn @jwodder note that the method signature for `get_checksum` has changed. Files/directores are now represented by a tuple `(checksum, size)` instead of just a string `checksum`.


jwodder:
@dchiquito How are you definining the size of a directory?

jwodder (approved):

dandischema/digests/zarr.py
```diff
@@ -15,15 +15,16 @@ class ZarrChecksum(pydantic.BaseModel):
...
     A checksum for a single file/directory in a zarr file.
 
-    Every file and directory in a zarr archive has a path and a MD5 hash.
+    Every file and directory in a zarr archive has a name and a MD5 hash.
     """
 
     md5: str
-    path: str
+    name: str
+    size: int
```
yarikoptic, line 23: I wonder if to simplify transition there could be a proper deprecation cycle which would support path, and decompose it into name and size as needed. Should be simple right? Wdyt?

dandischema/digests/zarr.py
```diff
@@ -125,9 +127,13 @@ def generate_listing(
                 files=sorted(files) if files is not None else [],
                 directories=sorted(directories) if directories is not None else [],
             )
+        size = sum([file.size for file in checksums.files]) + sum(
+            [directory.size for directory in checksums.directories]
+        )
```
yarikoptic, line 132: Minor

dandischema/digests/zarr.py
dchiquito: I don't think that's necessary. As soon as this is released and incorporated in dandi-archive and dandi-cli, we can just recompute checksums on all zarrs and `path` will be updated to `name`.

yarikoptic: well, since there is no way to announce that `dandischema` (next release) breaks `dandi-cli << (whatever would become compatible with new interface)` people would end up with a dandi-cli which might puke an exception if they manage to upgrade dandischema but not dandi-cli.  unlikely but possible.  deprecation cycle allows to amortize perturbation and avoid such problems...
oh well -- since python is dynamic and zarr isn't primary target yet for users, I am ok to proceed without deprecation cycle at hopefully low or none count of affected users.

dandischema/digests/zarr.py
```diff
@@ -125,9 +127,13 @@ def generate_listing(
                 files=sorted(files) if files is not None else [],
                 directories=sorted(directories) if directories is not None else [],
             )
+        size = sum(file.size for file in checksums.files) + sum(
+            directory.size for directory in checksums.directories
+        )
```
yarikoptic, line 132: yeap, itertools, smth like

```suggestion
        size = sum(
            _.size for _ in itertools.chain(checksums.files, checksums.directories)
        )
```

dandischema/digests/zarr.py
jwodder: If we make the next dandischema release a new minor version, it would be incompatible with dandi-cli's `dandischema ~= 0.5.1` requirement and thus help to prevent that from happening (but there would still be problems with checksum failures if people use the old client to upload Zarrs).

dandischema/digests/zarr.py
jwodder: (a) The attributes are lists, so it'd be more efficient to add them than use `chain`; (b) I think this might caught problems if we ever get full type-checking in this library (See #100).

yarikoptic: > (a) The attributes are lists, so it'd be more efficient to add them than use `chain`;

why?  how creating one more list would be more efficient?

<details>
<summary>some timing shows that chain is efficient</summary> 

```shell
$> ipython3
Python 3.9.10 (main, Jan 16 2022, 17:12:18) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.31.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: files = [f'file{i}' for i in range(1000)]

In [2]: dirs = [f'dir{i}' for i in range(1000)]

In [3]: %timeit sum(len(_) for _ in files + dirs)
147 µs ± 11.2 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [4]: from itertools import chain

In [5]: %timeit sum(len(_) for _ in chain(files, dirs))
138 µs ± 6.17 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [6]: %timeit sum(len(_) for _ in files + dirs)
155 µs ± 15.7 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [7]: %timeit sum(len(_) for _ in chain(files, dirs))
149 µs ± 6.33 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

```
</details> 

but overall -- either way is fine with me.  I just dealing with duplication allergy of having multiple avoidable comprehensions

re type-checking -- again why?  if files and directories entries types are the ones having .size (may be from the same base class even to not rely on duck typing), type checking should then just get smarter since it is IMHO all legit.

dandischema/digests/zarr.py
yarikoptic: ah, right, nice.
@jwodder Could you prep a PR for dandi-cli to accompany this one?

dandischema/digests/zarr.py
jwodder: Regarding the type-checking: I thought the code was using different type for files & directories, but they're actually both `ZarrChecksum`s, so there wouldn't be any problem.

dchiquito: I think this is not enough of a bottleneck to benefit from minor optimizations like this, and I think `sum(...) + sum(...)` is more legible an intuitive.

dchiquito:
@jwodder  In https://github.com/dandi/dandischema/pull/120/commits/38efd136ecac2b4359cdff2045f5bb2ee0fe0dc5 I changed the `md5` field to `digest`, since it no longer represents just an md5. You will need to calculate directory checksums with the new metadata appended. There are two new methods, `generate_directory_digest` and `parse_directory_digest` that might help.

jwodder:
@dchiquito Does `get_checksum()` return the new directory checksums?  If so, I don't think I need to change my code.

dchiquito:
@jwodder It does, yes. Just wanted to make you aware of the change.

satra (changes requested):
we need to update the schema checks for the digest as well. also any reason to use one `-` followed by two `--` in the digest?

dchiquito:
@satra one `-` then two `--` was Yarick's original suggestion in https://github.com/dandi/dandi-archive/issues/937#issuecomment-1057295702. I am ambivalent.

dchiquito:
@satra can you approve if you have no more qualms?

satra:
@dchiquito - sorry i wasn't clear in my previous comment.

when i meant schema checks, i meant changes in `models.py` and `test_models.py` to reflect the new format of the checksum, so that the metadata validates.

AlmightyYakob (commented):
This all looks good to me, approval pending Satra's comments being addressed.

dchiquito:
Ah I see, I'll fix that up

dchiquito:
@satra This PR was tagged `minor` but not `release` so the release was not cut. What's the procedure for releasing 0.6.0?

PR #123, opened 2022-03-09
fix: add type of id to hasMember

satra:
hasMember needs type augmentation to identify all the dandiassets as uris.


yarikoptic:
could some test be added (or just changed) to demonstrate/cover this case?

satra:
we haven't really added any of the jsonld stuff beyond just decorating things. a proper test for this would be that we are generating an appropriate quad. before i bring the whole pyld/shacl world into this, i think this is ok for now.

but i pushed a silly test.

satra:
@dchiquito - i'll wait for these tests and then this should trigger a release.

satra:
@jwodder i'll merge this despite the zarr test failures, since https://github.com/dandi/dandi-cli/pull/933 needs this new version of dandischema library to fix the test.

PR #124, opened 2022-03-24
Make models stricter

satra:
This makes a few properties required in line with the intent of this archive.

- [X] require ethics protocol for all dandisets
- [X] require biosample sex, species, age, type
~- [ ] turn missing funding into a warning~


satra:
@bendichter and @yarikoptic - we need to decide if this strictness proposed here should be enforced at the model level. my vote is +1, but we should discuss if there are any drawbacks to making these fields mandatory. i think it will improve the quality of datasets, but some of this will have to be included in messages from dandi cli as to what the user is expected to do.

also currently there is no way of separating exceptions and warning using pydantic. i will open a separate issue for enhancing the validation workflow. this should also help the api-server when providing into to users on the gui. i canceled the third todo.

satra:
@yarikoptic - also in this case the cli tests are failing because currently included schema in those tests are missing `ethicsApproval` and possibly other metadata that we have just made mandatory. perhaps we need to think about a better set of tests in the dandischema itself, and reuse those metadata jsons in the cli testing, rather then creating json files in the cli itself.

dandischema/models.py
```diff
@@ -796,17 +796,15 @@ class Participant(DandiBaseModel):
         nskey="dandi",
     )
     vendor: Optional[Organization] = Field(None, nskey="dandi")
-    age: Optional[PropertyValue] = Field(
-        None,
+    age: PropertyValue = Field(
```
bendichter, line 799: Age can be tricky.
* For some species, such as C. Elegens, the maturity of the organism is poorly indicated by the time since birth. They mature at different rates depending on their environment.
* Sharing are for humans can be against HIPAA.
    * Birthdays are considered HIPAA protected. 
    * If you give the session start time and the age in day resolution, the birthday would be recoverable, so this would also be against HIPAA.
    * It's also against HIPAA to provide the age of the participant if it is over 89 years.
* What would be the age of a fetus that was surgically removed from a pregnant mother?

yarikoptic:
@satra - I will convert this PR to draft since has not seen much activity and has unresolved discussions, but also with that pinging your attention to get back to it.

satra:
@yarikoptic - based on our last conversation, i think we would put this in some extra validation layer not enforce in the schema.

PR #126, opened 2022-03-24
Directly base Docker image off of dandi-archive image

jwodder:
As discussed in Slack.


yarikoptic:
@satra gave thumbs ups, I didn't analyze the situation, but will just trust @jwodder on this ;) Let's proceed

PR #128, opened 2022-04-01
ENH/RF: do not enforce any DANDI instance by default, support DJANGO_DANDI_WEB_APP_URL env var to specify one

yarikoptic:
Ref: https://github.com/dandi/dandi-archive/issues/961

I have decided to not hardcode gui-staging. one because the fix is targetting
specifically dandisets testing, and otherwise AFAIK is not needed elsewhere.
If something wants to upload to staging, they can also adjust their CI to point
to whatever instances they upload to

@jwodder -- do you think this would work if we set env var to point to staging in testing of dandisets?

### DIFF
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -40,14 +40,16 @@ if sys.version_info < (3, 8):
 else:
     from typing import Literal
 
+# Use DJANGO_DANDI_WEB_APP_URL to point to a specific deployment.
+DANDI_WEB_APP_URL = os.environ.get("DJANGO_DANDI_WEB_APP_URL", None)
+# Ensure no trailing / for consistency
+DANDI_ARCHIVE_PATTERN = (
+    re.escape(DANDI_WEB_APP_URL.rstrip("/")) if DANDI_WEB_APP_URL else ".*"
+)
+
 # Local or test deployments of dandi-api will insert URLs into the schema that refer to the domain
 # localhost, which is not a valid TLD. To make the metadata valid in those contexts, setting this
 # environment variable will use a less restrictive pydantic field that allows localhost.
-# It is possible also to use DANDI_ALLOWED_LOCALHOST_URLS to point to some other deployment
-# other than dandiarchive.org
-DANDI_ARCHIVE_PATTERN = os.environ.get(
-    "DANDI_ALLOWED_LOCALHOST_URLS", r"https://dandiarchive.org"
-)
 if "DANDI_ALLOW_LOCALHOST_URLS" in os.environ:
     HttpUrl = AnyHttpUrl  # noqa: F811
     DANDI_ARCHIVE_PATTERN = rf"({DANDI_ARCHIVE_PATTERN}|http://localhost(:\d+)?)"
@@ -962,7 +964,7 @@ class CommonModel(DandiBaseModel):
         None, readOnly=True, description="permalink to the item", nskey="schema"
     )
     repository: HttpUrl = Field(
-        "https://dandiarchive.org/",
+        DANDI_WEB_APP_URL,
         readOnly=True,
         description="location of the item",
         nskey="dandi",
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -43,10 +43,14 @@ else:
 # Local or test deployments of dandi-api will insert URLs into the schema that refer to the domain
 # localhost, which is not a valid TLD. To make the metadata valid in those contexts, setting this
 # environment variable will use a less restrictive pydantic field that allows localhost.
-DANDI_ARCHIVE_PATTERN = r"https://dandiarchive.org"
+# It is possible also to use DANDI_ALLOWED_LOCALHOST_URLS to point to some other deployment
+# other than dandiarchive.org
+DANDI_ARCHIVE_PATTERN = os.environ.get(
+    "DANDI_ALLOWED_LOCALHOST_URLS", r"https://dandiarchive.org"
+)
 if "DANDI_ALLOW_LOCALHOST_URLS" in os.environ:
     HttpUrl = AnyHttpUrl  # noqa: F811
-    DANDI_ARCHIVE_PATTERN = r"(https://dandiarchive.org|http://localhost(:\d+)?)"
+    DANDI_ARCHIVE_PATTERN = rf"({DANDI_ARCHIVE_PATTERN}|http://localhost(:\d+)?)"
 
 
 NAME_PATTERN = r"^([\w\s\-\.']+),\s+([\w\s\-\.']+)$"
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -41,10 +41,10 @@ else:
     from typing import Literal
 
 # Use DJANGO_DANDI_WEB_APP_URL to point to a specific deployment.
-DANDI_WEB_APP_URL = os.environ.get("DJANGO_DANDI_WEB_APP_URL", None)
+DANDI_INSTANCE_URL = os.environ.get("DJANGO_DANDI_WEB_APP_URL", None)
 # Ensure no trailing / for consistency
-DANDI_ARCHIVE_PATTERN = (
-    re.escape(DANDI_WEB_APP_URL.rstrip("/")) if DANDI_WEB_APP_URL else ".*"
+DANDI_INSTANCE_URL_PATTERN = (
+    re.escape(DANDI_INSTANCE_URL.rstrip("/")) if DANDI_INSTANCE_URL else ".*"
 )
 
 # Local or test deployments of dandi-api will insert URLs into the schema that refer to the domain
@@ -52,7 +52,9 @@ DANDI_ARCHIVE_PATTERN = (
 # environment variable will use a less restrictive pydantic field that allows localhost.
 if "DANDI_ALLOW_LOCALHOST_URLS" in os.environ:
     HttpUrl = AnyHttpUrl  # noqa: F811
-    DANDI_ARCHIVE_PATTERN = rf"({DANDI_ARCHIVE_PATTERN}|http://localhost(:\d+)?)"
+    DANDI_INSTANCE_URL_PATTERN = (
+        rf"({DANDI_INSTANCE_URL_PATTERN}|http://localhost(:\d+)?)"
+    )
 
 
 NAME_PATTERN = r"^([\w\s\-\.']+),\s+([\w\s\-\.']+)$"
@@ -63,7 +65,9 @@ ASSET_UUID_PATTERN = r"^dandiasset:" + UUID_PATTERN
 VERSION_PATTERN = r"\d{6}/\d+\.\d+\.\d+"
 DANDI_DOI_PATTERN = rf"^10.(48324|80507)/dandi\.{VERSION_PATTERN}"
 DANDI_PUBID_PATTERN = rf"^DANDI:{VERSION_PATTERN}"
-PUBLISHED_VERSION_URL_PATTERN = rf"^{DANDI_ARCHIVE_PATTERN}/dandiset/{VERSION_PATTERN}$"
+PUBLISHED_VERSION_URL_PATTERN = (
+    rf"^{DANDI_INSTANCE_URL_PATTERN}/dandiset/{VERSION_PATTERN}$"
+)
 MD5_PATTERN = r"[0-9a-f]{32}"
 SHA256_PATTERN = r"[0-9a-f]{64}"
 
@@ -964,7 +968,7 @@ class CommonModel(DandiBaseModel):
         None, readOnly=True, description="permalink to the item", nskey="schema"
     )
     repository: HttpUrl = Field(
-        DANDI_WEB_APP_URL,
+        DANDI_INSTANCE_URL,
         readOnly=True,
         description="location of the item",
         nskey="dandi",
--- a/dandischema/consts.py
+++ b/dandischema/consts.py
@@ -1,5 +1,5 @@
-DANDI_SCHEMA_VERSION = "0.6.2"
-ALLOWED_INPUT_SCHEMAS = ["0.4.4", "0.5.1", "0.5.2", "0.6.0", "0.6.1"]
+DANDI_SCHEMA_VERSION = "0.6.3"
+ALLOWED_INPUT_SCHEMAS = ["0.4.4", "0.5.1", "0.5.2", "0.6.0", "0.6.1", "0.6.2"]
 
 # ATM we allow only for a single target version which is current
 # migrate has a guard now for this since it cannot migrate to anything but current
--- a/dandischema/tests/test_models.py
+++ b/dandischema/tests/test_models.py
@@ -8,6 +8,7 @@ import pytest
 from .test_datacite import _basic_publishmeta
 from .. import models
 from ..models import (
+    DANDI_ARCHIVE_PATTERN,
     AccessRequirements,
     AccessType,
     Affiliation,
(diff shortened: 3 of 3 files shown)


jwodder:
Why are we even restricting the published URLs in the first place?  Don't we want to eventually support publishing on any arbitrary instance?

yarikoptic:
> Why are we even restricting the published URLs in the first place? Don't we want to eventually support publishing on any arbitrary instance?

well -- my thinking is that even in the bright future of distributed DANDIs, we might want to ensure that uploaded metadata does target that instance instead of some other or central dandiarchive.org. That is why probably it does make sense to restrict

jwodder:
@yarikoptic But it seems likely that people would want to validate metadata of assets from other instances, without knowing the instance URL, and/or against assets from multiple instances in a single Python process, all of which would be harmed by restricting the URL in the schema.

satra:
@yarikoptic - this looks good to me (in theory). btw, the cli tests take a long time to run. would be worthwhile taking a closer look.

@jwodder - i think we need to have a clear idea within our team of what we mean by urls, especially what it means for dandi as a project. despite decentralized stores, we may still have a central entity minting DANDI id's (just like dois are minted by particular agencies). we may consider being a registrar of decentralized instances, but we haven't really discussed the details.

on a more practical front, the only thing that can publish with an actual retrievable and followable doi is on the main instance. while the staging instance mints dois that are in the test doi server. 

i believe there is a separate issue where we have talked about closing loops completely. 
- dandiarchive redirector, gui, doi  
- staging redirector, staging gui, test doi. 

the piece that makes this a bit complicated is that there is only a single identifiers.org that resolves DANDI:<id>, and therefore we would also need a different prefix DANDIstaging:<id> (but identifiers.org may not allow us to register this - we will have to check).

perhaps we can attempt to set up these independent loops of dandi with an alteration of dandischema to allow different prefixes.

yarikoptic:
@satra , hm, not yet sure I would buy into the "DANDI registrars" idea.  I would have preferred more of a "freedom of git" approach toward decentralization. So overall, in the long run I think I would slide with @jwodder in making URLs unrestricted at the level of schema, and rather rely on tests for a specific instance (like e.g. dandi-cli knows with which instance it corresponds with) to ensure consistent URLs being used.

@satra If you feel strong about keeping schema "restrictive" in validation, then let's I guess proceed with this PR if @jwodder could confirm that it would be sufficient for us to resolve the failing tests situation.  But if you start to slide with us in making them less restrictive, then may be let's at once offload URL testing into unittests of dandischema (where/if it knows which instance it talks to) and dandi-cli?

dandischema/models.py
```diff
@@ -958,7 +964,7 @@ class CommonModel(DandiBaseModel):
         None, readOnly=True, description="permalink to the item", nskey="schema"
     )
     repository: HttpUrl = Field(
-        "https://dandiarchive.org/",
+        DANDI_WEB_APP_URL,
```
yarikoptic, line 967: @satra -- I wonder if we should completely remove a default value here in the "generic" `Resource` data type -- why should it be dandi specific by default?

yarikoptic:
ok, in the last commit I implemented refactoring to "opt-in" to have archive URL "hardcoded". 
It will be taken from the environment variable DJANGO_DANDI_WEB_APP_URL
if specified, as it is now set in the teraform configuration:

    $> git grep DJANGO_DANDI_WEB_APP_URL
    terraform/api.tf:    DJANGO_DANDI_WEB_APP_URL                     = "https://dandiarchive.org"
    terraform/staging_pipeline.tf:    DJANGO_DANDI_WEB_APP_URL                     = "https://gui-staging.dandiarchive.org"

so with this change we would get deployed setups ensuring that metadata
does have corresponding to that deployment metadata.

Note that  Resource.repository  is also set based on the
DJANGO_DANDI_WEB_APP_URL environment variable.
It is a datatype for CommonModel.relatedResource . To be generic, I do not
think we even should impose any specific default value there.

@jwodder and @dchiquito -- do you foresee or could test for any problems which could be potentially be brought up with this change?  Would it resolve the https://github.com/dandi/dandi-archive/issues/961 ?

dandischema/models.py
satra: the only reason is that the schema is quite dandi specific at the moment. btw this section says `DANDI_WEB_APP_URL`? can we call it `DANDI_INSTANCE_URL`.  i don't want this url to be `gui.dandiarchive.org`. this has to be the `dandiarchive.org` for the production instance.

satra:
as you can see in the staging terraform, we already have a mismatch in instance type. we should really create at least a redirect from `staging.dandiarchive.org` --> `gui-staging.dandiarchive.org`, if not a full redirector service just like production

yarikoptic:
I don't mind a redirect/tuning up terraform for staging. All those activities wouldn't need changes in this PR AFAIK

dandischema/models.py
yarikoptic: I don't care much about variable name, just was matching the one from terraform configuration, but can change tomorrow to the one you like

satra:
@yarikoptic - this PR is fine. it's just that this again seems like a short term patch, not fixing the separation problem. this can be merged so cli can work, but then it would be useful if we actually fixed the set of problems related to this issue, including introducing the idea of an instance `prefix` in the schema.

yarikoptic:
Indeed this is not the "optimal", but I think it is more than just a "fix" but rather an enhancement to support multiple instances (#76) and needs more thought!

<details>
<summary>dandi-cli is failing with obnoxious diff</summary> 

```shell
>       assert data == BareAssetMeta(**data_as_dict)
E       AssertionError: assert BareAsset(id=...sameAs=None)]) == BareAsset(id=...sameAs=None)])
E         Full diff:
E         - BareAsset(id='dandiasset:0b0a1a0b-e3ea-4cf6-be94-e02c830d54be', schemaKey='Asset', schemaVersion='0.6.2', name=None, description=None, contributor=None, about=None, studyTarget=None, license=None, protocol=None, ethicsApproval=None, keywords=['test', 'sample', 'example', 'test-case'], acknowledgement=None, access=[AccessRequirements(id=None, schemaKey='AccessRequirements', status=<AccessType.OpenAccess: 'dandi:OpenAccess'>, contactPoint=None, description=None, embargoedUntil=None)], url=None, repository=AnyHttpUrl('https://dandiarchive.org/', scheme='https', host='dandiarchive.org', tld='org', host_type='domain', path='/'), relatedResource=None, wasGeneratedBy=[Session(id=None, schemaKey='Session', identifier='XYZ789', name='XYZ789', description='Some test data', startDate=datetime.datetime(2020, 8, 31, 15, 58, 28, tzinfo=datetime.timezone(datetime.timedelta(days=-1, seconds=72000))), endDate=None, wasAssociatedWith=None, used=[Equipment(id=None, schemaKey='Equipment', identifier='probe:probe04', name='Ecephys Probe', description=None)])], contentSize=69105, encodingFormat='application/x-nwb', digest={<DigestType.dandi_etag: 'dandi:dandi-etag'>: 'e455839e5ab2fa659861f58a423fd17f-1'}, path='/test/path', dateModified=None, blobDateModified=None, dataType=None, sameAs=None, approach=None, measurementTechnique=None, variableMeasured=None, wasDerivedFrom=[BioSample(id=None, schemaKey='BioSample', identifier='cell01', sampleType=SampleType(id=None, schemaKey='SampleType', identifier=None, name='cell'), assayType=None, anatomy=None, wasDerivedFrom=[BioSample(id=None, schemaKey='BioSample', identifier='slice02', sampleType=SampleType(id=None, schemaKey='SampleType', identifier=None, name='slice'), assayType=None, anatomy=None, wasDerivedFrom=[BioSample(id=None, schemaKey='BioSample', identifier='tissue03', sampleType=SampleType(id=None, schemaKey='SampleType', identifier=None, name='tissuesample'), assayType=None, anatomy=None, wasDerivedFrom=None, wasAttributedTo=None, sameAs=None, hasMember=None)], wasAttributedTo=None, sameAs=None, hasMember=None)], wasAttributedTo=None, sameAs=None, hasMember=None)], wasAttributedTo=[Participant(id=None, schemaKey='Participant', identifier='a1b2c3', altName=None, strain=None, cellLine=None, vendor=None, age=PropertyValue(id=None, schemaKey='PropertyValue', maxValue=None, minValue=None, unitText='ISO-8601 duration', value='P170DT12212S', valueReference=PropertyValue(id=None, schemaKey='PropertyValue', maxValue=None, minValue=None, unitText=None, value='dandi:BirthReference', valueReference=None, propertyID=None), propertyID=None), sex=SexType(id=None, schemaKey='SexType', identifier=AnyHttpUrl('http://purl.obolibrary.org/obo/PATO_0000384', scheme='http', host='purl.obolibrary.org', tld='org', host_type='domain', path='/obo/PATO_0000384'), name='Male'), genotype='Typical', species=SpeciesType(id=None, schemaKey='SpeciesType', identifier=AnyHttpUrl('http://purl.obolibrary.org/obo/NCBITaxon_1234175', scheme='http', host='purl.obolibrary.org', tld='org', host_type='domain', path='/obo/NCBITaxon_1234175'), name='Cyperus bulbosus'), disorder=None, relatedParticipant=None, sameAs=None)])

E         + BareAsset(id='dandiasset:0b0a1a0b-e3ea-4cf6-be94-e02c830d54be', schemaKey='Asset', schemaVersion='0.6.2', name=None, description=None, contributor=None, about=None, studyTarget=None, license=None, protocol=None, ethicsApproval=None, keywords=['test', 'sample', 'example', 'test-case'], acknowledgement=None, access=[AccessRequirements(id=None, schemaKey='AccessRequirements', status=<AccessType.OpenAccess: 'dandi:OpenAccess'>, contactPoint=None, description=None, embargoedUntil=None)], url=None, repository=None, relatedResource=None, wasGeneratedBy=[Session(id=None, schemaKey='Session', identifier='XYZ789', name='XYZ789', description='Some test data', startDate=datetime.datetime(2020, 8, 31, 15, 58, 28, tzinfo=datetime.timezone(datetime.timedelta(days=-1, seconds=72000))), endDate=None, wasAssociatedWith=None, used=[Equipment(id=None, schemaKey='Equipment', identifier='probe:probe04', name='Ecephys Probe', description=None)])], contentSize=69105, encodingFormat='application/x-nwb', digest={<DigestType.dandi_etag: 'dandi:dandi-etag'>: 'e455839e5ab2fa659861f58a423fd17f-1'}, path='/test/path', dateModified=None, blobDateModified=None, dataType=None, sameAs=None, approach=None, measurementTechnique=None, variableMeasured=None, wasDerivedFrom=[BioSample(id=None, schemaKey='BioSample', identifier='cell01', sampleType=SampleType(id=None, schemaKey='SampleType', identifier=None, name='cell'), assayType=None, anatomy=None, wasDerivedFrom=[BioSample(id=None, schemaKey='BioSample', identifier='slice02', sampleType=SampleType(id=None, schemaKey='SampleType', identifier=None, name='slice'), assayType=None, anatomy=None, wasDerivedFrom=[BioSample(id=None, schemaKey='BioSample', identifier='tissue03', sampleType=SampleType(id=None, schemaKey='SampleType', identifier=None, name='tissuesample'), assayType=None, anatomy=None, wasDerivedFrom=None, wasAttributedTo=None, sameAs=None, hasMember=None)], wasAttributedTo=None, sameAs=None, hasMember=None)], wasAttributedTo=None, sameAs=None, hasMember=None)], wasAttributedTo=[Participant(id=None, schemaKey='Participant', identifier='a1b2c3', altName=None, strain=None, cellLine=None, vendor=None, age=PropertyValue(id=None, schemaKey='PropertyValue', maxValue=None, minValue=None, unitText='ISO-8601 duration', value='P170DT12212S', valueReference=PropertyValue(id=None, schemaKey='PropertyValue', maxValue=None, minValue=None, unitText=None, value='dandi:BirthReference', valueReference=None, propertyID=None), propertyID=None), sex=SexType(id=None, schemaKey='SexType', identifier=AnyHttpUrl('http://purl.obolibrary.org/obo/PATO_0000384', scheme='http', host='purl.obolibrary.org', tld='org', host_type='domain', path='/obo/PATO_0000384'), name='Male'), genotype='Typical', species=SpeciesType(id=None, schemaKey='SpeciesType', identifier=AnyHttpUrl('http://purl.obolibrary.org/obo/NCBITaxon_1234175', scheme='http', host='purl.obolibrary.org', tld='org', host_type='domain', path='/obo/NCBITaxon_1234175'), name='Cyperus bulbosus'), disorder=None, relatedParticipant=None, sameAs=None)])

```
</details>

in test_metadata2asset due to expected now `repository=AnyHttpUrl('https://dandiarchive.org/'` -> `repository=None` which is expected given the change -- by default model doesn't know which repository an asset belongs. 
We could in principle "fill it in" by dandi-cli (although we would first need to get staging redirector and provide it in https://github.com/dandi/dandi-cli/blob/HEAD/dandi/consts.py#L116 and then channel all the way into upload) upon upload but most logical IMHO would be to mutate metadata by API server upon receiving it since it is the ultimate location aware of itself. WDYT @satra and @dandi/dandiarchive folks?

I have pushed increment to DANDI_SCHEMA_VERSION (only a patch) and added `minor` version boost to library (so dandi-cli / server would consciously upgrade to support it) and we could proceed with a merge or should we decide first on what should fill in `repository`? 

note: `repository` is read-only, but I assume that the readOnly is only for web UI and dandi-cli/api would have no difficulty in changing it, right? ;)

satra:
> `repository` is read-only

we strip out many of the readonly pieces on the server side and inject them there. `repository` is one of those things i believe. so the user cannot override them through a `POST/PUT`. if `repository` is not being populated, it should.

dandischema/models.py
yarikoptic: ok, since we are still here -- I have renamed it into `DANDI_INSTANCE_URL` and `DANDI_ARCHIVE_PATTERN` (after searching for it in the organization -- not used) into `DANDI_INSTANCE_URL_PATTERN` for consistency.

yarikoptic:
> > `repository` is read-only
> 
> we strip out many of the readonly pieces on the server side and inject them there. `repository` is one of those things i believe. so the user cannot override them through a `POST/PUT`. if `repository` is not being populated, it should.

I see
```
dandiapi/api/models/version.py:            'repository': settings.DANDI_WEB_APP_URL,
```
so may be it does, dunno. @dchiquito @AlmightyYakob - could you please check on dandi-archive end if `repository` is populated? or should we just proceed with the merge and try it out in staging with upgrade of `dandischema`? ;)

AlmightyYakob:
> so may be it does, dunno. @dchiquito @AlmightyYakob - could you please check on dandi-archive end if `repository` is populated? or should we just proceed with the merge and try it out in staging with upgrade of `dandischema`? ;)

I don't know if I fully understand what you're asking, but both the production and staging instance have the `DJANGO_DANDI_WEB_APP_URL` environment variable (and thus the `DANDI_WEB_APP_URL` django setting) set. For the record, those values are set [here](https://github.com/dandi/dandi-infrastructure/blob/cf4938a77bdd6707d822fb082f6d2b13305ab5b1/terraform/api.tf#L40) and [here](https://github.com/dandi/dandi-infrastructure/blob/cf4938a77bdd6707d822fb082f6d2b13305ab5b1/terraform/staging_pipeline.tf#L39), respectively.

satra:
@yarikoptic - yes that [line](https://github.com/dandi/dandi-archive/blob/471f34cb850aec53c827a65d06836f460c7b21fa/dandiapi/api/models/version.py#L242) adds the repository field to the metadata.

PR #132, opened 2022-04-19
Set DJANGO_DANDI_WEB_APP_URL for dandi-cli integration tests

### LINKED ISSUE #131: Tests against dandi-cli fail
https://github.com/dandi/dandi-schema/actions/runs/2187811011

jwodder:
Closes #131.

### DIFF
--- a/.github/workflows/test-dandi-cli.yml
+++ b/.github/workflows/test-dandi-cli.yml
@@ -14,6 +14,7 @@ jobs:
     env:
       NO_ET: 1
       DANDI_ALLOW_LOCALHOST_URLS: 1
+      DJANGO_DANDI_WEB_APP_URL: https://dandiarchive.org/
     strategy:
       fail-fast: false
       matrix:


.github/workflows/test-dandi-cli.yml
```diff
@@ -14,6 +14,7 @@ jobs:
     env:
       NO_ET: 1
       DANDI_ALLOW_LOCALHOST_URLS: 1
+      DJANGO_DANDI_WEB_APP_URL: https://dandiarchive.org/
```
yarikoptic, line 17: hm, should it go into tests fixture or wouldn't otherwise we need to set this env var in every environment where we would run tests?

jwodder: dandischema reads the envvar when the dandischema package is loaded, which is before pytest fixtures take effect, so we need to set the envvar outside the Python process.

yarikoptic: ah, right. thanks! Let's proceed then, and see where else we would need to do that for now.

PR #133, opened 2022-04-26
add 0.6.2 to validation schemas

satra:
adds missing schema to validation logic


yarikoptic:
Just to make sure -- it is not to mitigate https://github.com/dandi/dandi-cli/pull/987#issuecomment-1111011012 right?

satra:
nope, but this is a bug.

yarikoptic:
@satra are you sure  you want `minor` release (thus making it 0.8.0) instead of just a `patch` (making it 0.7.1).  IMHO should be a patch

yarikoptic:
changing to patch, merging

PR #143, opened 2022-09-19
DOC: embargoedUntil -- extend semantic to be able to mark with the date when it was unembargoed

yarikoptic:
The motivation is to (ab)use this metadata field to enable annotating dandisets which were unembargoed in the archive. ATM it would be impossible (?) to tell from dandiset metadata if it was unemabrgoed since we do not carry much (if any) of provenance.  Whenever we unemabrgo a dandiset we could use this field then  to fill-in datetime for when it was unembargoed, which would actually be "semantically" correct since that would be the date time until when it was embargoed.  Of cause if the intention of this field is more of "unembargoUntil" (so no "done" notion), then it must remain as is.

If merged -- TODO: file an issue with dandi-archive to fill-in this field (possibly overwriting an existing value) upon "unembargo" action for the dandiset.

### DIFF
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -611,7 +611,7 @@ class AccessRequirements(DandiBaseModel):
     embargoedUntil: Optional[date] = Field(
         None,
         title="Embargo end date",
-        description="Date on which embargo ends.",
+        description="Date on which embargo ends or ended.",
         readOnly=True,
         nskey="dandi",
         rangeIncludes="schema:Date",


satra:
this will cause a change to the schema. but i think the field can be used when unembargo happens. so i would suggest doing the archive change first and use the field.

in terms of provenance, i've talked about a full audit log that's different from the schema. i think it should be prioritized once current issue are resolved. cc/ (@waxlamp)

yarikoptic:
> this will cause a change to the schema

rright... since such a minor change as a doc, I think we could make it just a patch and delay actual release until some other change "pushes it out"

>but i think the field can be used when unembargo happens. so i would suggest doing the archive change first and use the field.

coolio, so we are *inline* in our thinking then. Filed https://github.com/dandi/dandi-archive/issues/1286

yarikoptic:
ok, I will move it to draft, and say that it is blocked by https://github.com/dandi/dandi-archive/issues/1286

yarikoptic:
> > .... so i would suggest doing the archive change first and use the field.
> 
> coolio, so we are _inline_ in our thinking then. Filed [dandi/dandi-archive#1286](https://github.com/dandi/dandi-archive/issues/1286)

well -- as it appeared to be a chicken/egg issue as someone should do it first, and as schema is the "driver" here -- I will likely to proceed with this PR first so archive would 'match' the doc spec eventually ;)

PR #145, opened 2022-10-06
Update pre-commit hooks and rerun

jwodder:
This addresses a fatal problem with an older version of black using a now-removed private part of click.


yarikoptic:
I have merged #146 which is what I guess made it all red here, @jwodder please update/rerun .

jwodder:
@yarikoptic Rebased.

yarikoptic:
Looks good to me and non-functional changes, so to avoid conflicts via delaying I am merging now.

PR #148, opened 2022-10-12
Update GitHub Actions action versions

jwodder:
We are currently using older versions of the updated GitHub Actions actions, and these older versions still use Node 12, [which is deprecated](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/).

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -10,11 +10,11 @@ jobs:
 
     steps:
     - name: Set up environment
-      uses: actions/checkout@v1
+      uses: actions/checkout@v3
       with:  # no need for the history
         fetch-depth: 1
     - name: Set up Python
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v4
       with:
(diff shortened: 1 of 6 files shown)


yarikoptic:
hm, in another PR it errored on `Test without network / test (ubuntu-18.04, 3.8)` here on `windows-2019` but differently but for the same reason
```
ERROR: Could not install packages due to an OSError: HTTPSConnectionPool(host='files.pythonhosted.org', port=443): Max retries exceeded with url: /packages/dc/bf/5965230bf0547c5fa0005984564146dcc414e6e8b6349177eca413761013/pydantic-1.10.2-cp37-cp37m-win_amd64.whl (Caused by ConnectTimeoutError(<pip._vendor.urllib3.connection.HTTPSConnection object at 0x0000021F9B1F5E48>, 'Connection to files.pythonhosted.org timed out. (connect timeout=15)'))
```

which I don't think relate so merging. Thanks @jwodder

PR #149, opened 2022-10-12
Configure Dependabot to update GitHub Actions action versions

jwodder:


yarikoptic:
yeah, why not? Thank you @jwodder

PR #151, opened 2022-10-26
Test against Python 3.11

jwodder:


yarikoptic:
not sure why we have not merged it -- it was green, no conflicts (some concerns about out of date but should be ok). So let's proceed. Thank you @jwodder

Issue #153, opened 2023-01-09
Validating assets calls github repeatedly


danlamanna:
When calling `dandischema.metadata.validate` on n assets, n requests are made to github to fetch the schema. This makes validating assets take significantly longer than it should. The request also has no default timeout, meaning a call to `validate` can hang indefinitely. 

https://github.com/dandi/dandi-schema/blob/d34658cc24c0e3c0c3a88e92bcd29b158241448e/dandischema/metadata.py#L184-L187

Can `dandi-schema` be modified to avoid relying on the network for validation? Either by bundling the schemas from `dandi/schema` into package data, allowing the caller of `validate` to pass a schema directly, or some other means?

FWIW this problem appears to exist with `migrate` as well.

satra:
@danlamanna - we could probably package the schemas with dandi-schema. an alternative would be to cache the request on the server side. are the requests all in isolated processes or would a cache to keep a schema once downloaded work?

it's also the case that this download happens when an asset is using a different schema than the current one. this is true for many assets currently that were submitted a while back, but should not in theory be true for new assets being uploaded. i.e. the schema version should be the latest.

we have been planning to run a metadata update by processing the files with the latest extractor, but this hasn't been rolled into action.

danlamanna:
> we could probably package the schemas with dandi-schema. an alternative would be to cache the request on the server side. are the requests all in isolated processes or would a cache to keep a schema once downloaded work?

Avoiding network requests altogether would be best for maximizing reliability. A cache combined with giving the caller control over how network requests are performed (timeouts, retries, etc) would be the next best option.

satra:
dandischema in general requires access to online resources to carry out it's general work, so it will never be a network free library. but we can optimize it in some ways. we didn't want to make assumptions about availability of storage, persistence etc when we wrote that component, but i can try a few changes. @djarecka and @sooyounga - is this something you folks could take a stab at? happy to discuss details.

waxlamp:
@satra, Dan's idea has a lot of merit: even if the goal is to always be validating against the newest schema version, we are not there yet, and keeping the allowed schema versions as static package data would gain us an immediate and obvious win (while we are still litigating, so to speak, schema autoupgrades etc.).

Dan can create a quick proof of concept so we can observe the benefits/drawbacks of the approach. He can coordinate this idea with whatever Dorota and Sooyoung are looking into as well.

satra:
@waxlamp - i have no issues with a proof of concept.

PR #155, opened 2023-01-18
Statically bundle dandi/schema json

danlamanna:
This is a minimal PR to show an example of what was discussed in #153.

The git submodule approach seemed to be the simplest way to ensure the schema files were present in every environment: development, testing, and packaging. It should be easy to programatically "bump" this when new versions of `dandi/schema` are released as well.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -47,6 +47,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
+          submodules: true
           path: dandischema
           token: ${{ secrets.DANDI_GITHUB_TOKEN }}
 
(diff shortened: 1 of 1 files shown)


satra:
@danlamanna - the release workflow for this project generates the schema and pushes to the repo during an automated github action (https://github.com/dandi/dandi-schema/blob/master/.github/workflows/release.yml#L76) and then builds the package for distribution via pypi and uploads there.  

will this change then pick the latest release from the schema repo?

danlamanna:
> the release workflow for this project generates the schema and pushes to the repo during an automated github action ([master/.github/workflows/release.yml#L76](https://github.com/dandi/dandi-schema/blob/master/.github/workflows/release.yml?rgh-link-date=2023-01-31T14%3A26%3A54Z#L76)) and then builds the package for distribution via pypi and uploads there.

> will this change then pick the latest release from the schema repo?

It doesn't, I just made this to get feedback on the approach. If it looks good then I can adjust the `release.yml` to:
1. Generate the new schema files and commit/push them to `dandi/schema` 
2. Update the submodule revision and commit it in `dandi/dandi-schema`

setup.cfg
```diff
@@ -28,6 +28,9 @@ project_urls =
 [options]
 python_requires = >=3.7
 install_requires =
+    # importlib-resources can be removed once 3.10 becomes the minimum required version.
+    # https://setuptools.pypa.io/en/latest/userguide/datafiles.html#accessing-data-files-at-runtime
+    importlib-resources
```
satra, line 33:
```suggestion
    importlib-resources; python_version < "3.10"
```

satra:
the approach looks reasonable. i would add a github action test that checks that a package built from this repo with the submodule enabled can be installed and run, and can be used to access all acceptable versions of the schema.

danlamanna:
I attempted to update the release action in a64e806, but I don't have a way of testing it. I'm also unfamiliar with `auto` so I'm not sure how to grab the version that *will be* released (see the TODO). @jwodder It looks like you wrote the release action originally, would you mind taking a look? The new workflow should be 1) generate new schema versions and commit/push them to `dandi/schema` 2) update the submodule for `dandi-schema` to include the new schema versions and 3) commit/tag/push the new version of `dandi-schema` with the updated submodule.

> i would add a github action test that checks that a package built from this repo with the submodule enabled can be installed and run, and can be used to access all acceptable versions of the schema.

I'm not sure I follow. The nice thing about this approach is that the package is always used with the submodule enabled, so the existing tests using tox already build such a package.

jwodder:
@danlamanna You can get the to-be-released version via `${{ needs.release-check.outputs.auto-version }}`.

yarikoptic:
FWIW, I agree that it is a good approach. The only usecase I see this would disallow for is validation against a "future" (to the used older dandi-schema package version) version of schema. So it would mean to need to update `dandi-schema` package every time we release a new version.  If nobody sees a problem with that, as "we need to do that now anyways already in `dandi-archive`", then all good.

jwodder:
@danlamanna Unfortunately, auto does not expose the new version that it's releasing, so you'll have to take the version bump level from `${{ needs.release-check.outputs.auto-version }}` and apply it to the most recent prior version.  [Here's some code to help you out.](https://github.com/datalad/release-action/blob/2a1e20e6f9069149277e4a4d9f7f30bc38fb830c/datalad_release_action/versions.py)

danlamanna:
I think this latest revision makes sense, but I'm skeptical it will work whenever the next release is cut since I don't have a proper sandbox for testing it.

PR #161, opened 2023-02-17
Autoupdate .pre-commit-config.yaml

jwodder:
The version of isort currently specified in `pre-commit-config.yaml` fails to install, and we might as well update the other hooks while we're at it.


yarikoptic:
CI is green, so I don't see why not. Thanks!

PR #173, opened 2023-04-06
Count only for the first hit for subject or tissue within filename

### LINKED ISSUE #172: Incorrect summaries (at least number of subjects) in some dandisets
It was mentioned by the dandiset author for https://dandiarchive.org/dandiset/000212 that there is only 1013 files but summary has 1097 subjects!

I did check across all dandisets for which we have datalad representations:

```shell
(dandisets) dandi@drogon:/mnt/backup/dandi/dandisets$ for d in 0*; do ( cd $d; n1=$(awk '/numberOfSubjects:/{print $2}' dandiset.yaml || echo 0 ); n2=$(/bin/ls -1d sub-* 2>/dev/null | wc -l || echo 0); [ -n "$n1" ] && { [ "$n1" = "$n2" ] || echo "`pwd`: Mismatch $n1 != $n2"; } ); done
/mnt/backup/dandi/dandisets/000025: Mismatch 1 != 0
/mnt/backup/dandi/dandisets/000043: Mismatch 22 != 13
/mnt/backup/dandi/dandisets/000051: Mismatch 1 != 0
/mnt/backup/dandi/dandisets/000052: Mismatch 1 != 0
/mnt/backup/dandi/dandisets/000058: Mismatch 1 != 0
/mnt/backup/dandi/dandisets/000168: Mismatch 30 != 0
/mnt/backup/dandi/dandisets/000212: Mismatch 1097 != 1013
/mnt/backup/dandi/dandisets/000228: Mismatch 64 != 32
/mnt/backup/dandi/dandisets/000246: Mismatch 12 != 11
/mnt/backup/dandi/dandisets/000288: Mismatch 12 != 9
/mnt/backup/dandi/dandisets/000293: Mismatch 201 != 121
/mnt/backup/dandi/dandisets/000297: Mismatch 197 != 118
/mnt/backup/dandi/dandisets/000341: Mismatch 310 != 0
/mnt/backup/dandi/dandisets/000362: Mismatch 11 != 0
/mnt/backup/dandi/dandisets/000454: Mismatch 5 != 4
```

so it seems that the issue manifests itself quite wildly.

yarikoptic:
Also added assertion so we do not count incorrectly. But may be should be just a warning?

Closes #172


satra:
i think this also requires line 302 replaced with the sanitize function from dandi-cli.

yarikoptic:
> i think this also requires line 302 replaced with the sanitize function from dandi-cli.

did you try that and it resolved #172?  would better be done in a separate PR then since this one does resolve issue too

yarikoptic:
4

> > i think this also requires line 302 replaced with the sanitize function from dandi-cli.
> 
> did you try that and it resolved #172? would better be done in a separate PR then since this one does resolve issue too

I tested, it does. Will submit a complimentary PR.

yarikoptic:
@satra so what do you think about this one.  It is complimentary to #175

dandischema/metadata.py
```diff
@@ -311,13 +311,16 @@ def _add_asset_to_stats(assetmeta: Dict[str, Any], stats: _stats_type) -> None:
             stats = _get_samples(value, stats, hierarchy)
             break
 
+    # which components already found, so we do not count more than
+    # once in some incorrectly named datasets
+    found: Dict[str, str] = {}
     for part in Path(assetmeta["path"]).name.split(".")[0].split("_"):
-        if part.startswith("sub-"):
-            subject = part.replace("sub-", "")
+        if found.get("subject") and part.startswith("sub-"):
```
satra, line 318: shouldn't this also be `not found...` like `sample` below.

yarikoptic: rright! so
```suggestion
        if not found.get("subject") and part.startswith("sub-"):
```

Issue #176, opened 2023-04-17
FYI: Pydantic v2 is breaking things


effigies:
Just a heads up that we're getting failures in PyBIDS on our pre-release tests due to a dependency using pydantic. A quick glance didn't indicate that you're testing pre-releases, so I wanted to make sure you weren't caught off-guard when the full release comes out.

* https://docs.pydantic.dev/blog/pydantic-v2
* https://github.com/pydantic/pydantic/releases

yarikoptic:
Thank you @effigies for the heads up!  I guess, while addressing incompatibility, we better add a `dev` run on CI here to also be able to be proactive.  @jwodder -- please look into adding such a CI run and possibly fixing up any issue which might be arising?  If no issues, I guess we would better still keep the upper bound before 2.0 you added in #177 until 2.0 is actually released, whenever we could boost it to `< 3.0`.

yarikoptic:
ok, @candleindark let's look into just migrating over to use pydantic 2.0 -- first approach here, and then see what it would take in the projects which use dandischema (see https://github.com/dandi/dandi-schema/pull/178#issuecomment-1513374751)

satra:
> Do you mean migrating to Pydantic 2.0 for real now, without the support for Pydantic 1.0?

yup! no backwards compatibility wanted at this time.

candleindark:
> > Do you mean migrating to Pydantic 2.0 for real now, without the support for Pydantic 1.0?
> 
> yup! no backwards compatibility wanted at this time.

Great. That would be cleaner.

yarikoptic:
reconfirming -- full migration to v2 without using a shim

PR #177, opened 2023-04-17
Restrict pydantic requirement to < 2.0

jwodder:
We can update the code to use pydantic 2.0 after it comes out.

See #176.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,7 +30,7 @@ project_urls =
 python_requires = >=3.7
 install_requires =
     jsonschema[format]
-    pydantic[email] >= 1.8.1
+    pydantic[email] >= 1.8.1, < 2.0
     typing_extensions;  python_version < "3.8"
     requests
 zip_safe = False


yarikoptic:
Sad part of the story is that then `pip` would likely just to install some prior release of `dandi-schema` if some newer "bad" pydantic would already be installed, so overall this specification is good to have but is not really even a workaround.  Looking forward, I guess we better indeed strive to keep such specification in the future to prevent "known to break something next major release" from being installed.  Meanwhile will also add release label and merge.

jwodder:
@yarikoptic

> Sad part of the story is that then pip would likely just to install some prior release of dandi-schema if some newer "bad" pydantic would already be installed

A (somewhat drastic) way to prevent this from happening would be to yank (not delete!) all older dandischema releases from PyPI, which would prevent pip 22.0+ from installing those versions unless the user specifically requested the older version with `==`.  If someone then tried installing this latest dandi-schema in an environment in which pydantic 2.0 had already been installed, I believe pip would fail without installing anything.

PR #178, opened 2023-04-17
Test against pydantic 2.0 prereleases

jwodder:
Part of #176.

Note that we cannot make the code compatible with both pydantic 1.x and 2.x, as [all pydantic instance methods have been renamed](https://docs.pydantic.dev/blog/pydantic-v2/#model-namespace-cleanup).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,6 +25,11 @@ jobs:
           - 3.9
           - '3.10'
           - '3.11'
+        toxenv: [py]
+        include:
+          - os: ubuntu-latest
+            python: 3.7
+            toxenv: dev
     steps:
     - name: Set up environment
       uses: actions/checkout@v3
@@ -44,7 +49,7 @@ jobs:
         python -m pip install --upgrade tox
 
     - name: Run all tests
-      run: tox -e py -- -s --cov-report=xml
+      run: tox -e ${{ matrix.toxenv }} -- -s --cov-report=xml
 
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v3
--- a/.github/workflows/typing.yml
+++ b/.github/workflows/typing.yml
@@ -7,6 +7,12 @@ on:
 jobs:
   typing:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        testenv:
+          - typing
+          - devtyping
     steps:
       - name: Check out repository
         uses: actions/checkout@v3
@@ -24,4 +30,4 @@ jobs:
           python -m pip install --upgrade tox
 
       - name: Run type checker
-        run: tox -e typing
+        run: tox -e ${{ matrix.testenv }}
--- a/dandischema/digests/zarr.py
+++ b/dandischema/digests/zarr.py
@@ -5,7 +5,10 @@ import hashlib
 import re
 from typing import Dict, List, Optional, Tuple
 
-import pydantic
+try:
+    import pydantic.v1 as pydantic
+except ImportError:
+    import pydantic
 
 """Passed to the json() method of pydantic models for serialization."""
 ENCODING_KWARGS = {"separators": (",", ":")}
--- a/dandischema/metadata.py
+++ b/dandischema/metadata.py
@@ -4,7 +4,6 @@ from pathlib import Path
 from typing import Any, Dict, Iterable, Optional, TypeVar, Union, cast
 
 import jsonschema
-import pydantic
 import requests
 
 from .consts import (
@@ -17,6 +16,11 @@ from .exceptions import JsonschemaValidationError, PydanticValidationError
 from . import models
 from .utils import _ensure_newline, version2tuple
 
+try:
+    import pydantic.v1 as pydantic
+except ImportError:
+    import pydantic
+
 schema_map = {
     "Dandiset": "dandiset.json",
     "PublishedDandiset": "published-dandiset.json",
@@ -26,8 +30,6 @@ schema_map = {
 
 
 def generate_context() -> dict:
-    import pydantic
-
     field_preamble = {
         "@version": 1.1,
         "dandi": "http://schema.dandiarchive.org/",
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -7,17 +7,30 @@ import re
 import sys
 from typing import Any, Dict, List, Optional, Sequence, Type, TypeVar, Union, cast
 
-from pydantic import (
-    UUID4,
-    AnyHttpUrl,
-    BaseModel,
-    ByteSize,
-    EmailStr,
-    Field,
-    parse_obj_as,
-    root_validator,
-    validator,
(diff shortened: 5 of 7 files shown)


yarikoptic:
which methods do we actually use and can't we have some "shim" compatibility layer until we simply declare pydantic 2.0 to be the minimal supported version?

<details>
<summary>I started with this patch but immediately got into incompatible API and stopped -- but could you please check it deeper a bit?</summary> 

```patch
diff --git a/dandischema/compat.py b/dandischema/compat.py
new file mode 100644
index 0000000..6ae0b06
--- /dev/null
+++ b/dandischema/compat.py
@@ -0,0 +1,6 @@
+import pydantic
+
+class BaseModel(pydantic.BaseModel):
+    def json(self, *args, **kwargs):
+        # TODO: TypeError: BaseModel.model_dump_json() got an unexpected keyword argument 'separators'
+        return super().model_dump_json(*args, **kwargs)
\ No newline at end of file
diff --git a/dandischema/digests/zarr.py b/dandischema/digests/zarr.py
index b7f8eab..7f31a84 100644
--- a/dandischema/digests/zarr.py
+++ b/dandischema/digests/zarr.py
@@ -7,6 +7,8 @@ from typing import Dict, List, Optional, Tuple
 
 import pydantic
 
+from ..compat import BaseModel
+
 """Passed to the json() method of pydantic models for serialization."""
 ENCODING_KWARGS = {"separators": (",", ":")}
 ZARR_CHECKSUM_PATTERN = "([0-9a-f]{32})-([0-9]+)--([0-9]+)"
@@ -26,7 +28,7 @@ def parse_directory_digest(digest: str) -> tuple[str, int, int]:
 
 
 @total_ordering
-class ZarrChecksum(pydantic.BaseModel):
+class ZarrChecksum(BaseModel):
     """
     A checksum for a single file/directory in a zarr file.
 
@@ -42,7 +44,7 @@ class ZarrChecksum(pydantic.BaseModel):
         return self.name < other.name
 
 
-class ZarrChecksums(pydantic.BaseModel):
+class ZarrChecksums(BaseModel):
     """
     A set of file and directory checksums.
 
diff --git a/dandischema/models.py b/dandischema/models.py
index 0d73fca..292d100 100644
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -10,7 +10,6 @@ from typing import Any, Dict, List, Optional, Sequence, Type, TypeVar, Union, ca
 from pydantic import (
     UUID4,
     AnyHttpUrl,
-    BaseModel,
     ByteSize,
     EmailStr,
     Field,
@@ -20,6 +19,7 @@ from pydantic import (
 )
 
 from .consts import DANDI_SCHEMA_VERSION
+from .compat import BaseModel
 from .digests.dandietag import DandiETag
 from .digests.zarr import ZARR_CHECKSUM_PATTERN, parse_directory_digest
 from .utils import name2title
diff --git a/setup.cfg b/setup.cfg
index b71143b..cba00a1 100644
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,7 +30,7 @@ project_urls =
 python_requires = >=3.7
 install_requires =
     jsonschema[format]
-    pydantic[email] >= 1.8.1, < 2.0
+    pydantic[email] >= 1.8.1
     typing_extensions;  python_version < "3.8"
     requests
 zip_safe = False

```
</details>

jwodder:
@yarikoptic I'm not certain that trying to support pydantic 1.x and 2.x at the same time is worthwhile.

For the record, the repositories under the dandi org that use pydantic are:

- dandi/dandi-api-datasets: `tools/populate_dandiset_asset_yamls.py`: Does not specify a version
- dandi/dandi-archive: Does not specify a version.  However, the only recursive dependencies that require pydantic are dandischema and zarr-checksum, and so deployment of dandi-archive (assuming the only Python packages involved are those listed in its `setup.py` and their recursive dependencies) should continue to use pydantic 1.x even after 2.x comes out, up until dandischema and zarr-checksum both permit use with pydantic 2.x.
- dandi/dandi-cli: Requires `pydantic >= 1.9.0`.  However, it depends on `dandischema` and `zarr_checksum`, which both specify an upper bound on pydantic of `< 2.0` (and these are the only recursive dependencies that require pydantic), and so pip will either complain or (worse) backtrack if a user tries to install both dandi-cli and pydantic 2.0 in the same environment.
    - Honestly, if a user tries to install dandi in an environment that already has pydantic 2.x installed, I'd say they're doing something wrong.  We should be strongly encouraging users to only install dandi in a fresh virtualenv or Conda environment; installing a Python package at the system or user level is just playing with fire.
- dandi/dandi-schema: Now requires `pydantic >= 1.8.1, < 2.0`
- dandi/dandisets:
    - `tools/backups2datalad/`: Requires `pydantic >= 1.8, < 2.0`
    - `tools/get-upload-status.py`: Requires `pydantic >= 1.7, < 2.0`
- dandi/dandisets-healthstatus: Requires `pydantic >= 1.8, < 2.0`
- dandi/nwb-healthstatus: Requires `pydantic >= 1.7, < 2.0`
- dandi/zarr_checksum: Requires `pydantic >= 1.10.2, < 2.0`

yarikoptic:
> * We should be strongly encouraging users to only install dandi in a fresh virtualenv or Conda environment

overall: we can't really encourage such things, especially because dandi is also a library. E.g. if some tool (or even downstream internal script) would requires already pydantic 2.0 or would not be ready to upgrade to it yet -- we can't just demand them to either use outdated dandi (which might not be "good enough") or downgrade back to 1.x pydantic since dandi is not ready yet.
IMHO it was quite a bad decision from pydantic team to just break API without some compatibility shim with a DeprecationWarning and thus a more "gentle" deprecation cycle.

yarikoptic:
FWIW: I filed https://github.com/pydantic/pydantic/issues/5792

yarikoptic:
Ignoring by test suites is IMHO an acceptable cost during transition

jwodder:
@yarikoptic Remind me why we want to support both v1 and v2 at the same time in the first place.

yarikoptic:
> so I cannot test support for both versions.

can always be installed directly from git for testing, e.g. even from some specific committish to avoid "fluctuations" and later switch to pre-release whenever `a2` is out.

yarikoptic:
FWIW, I thought to ask for `a2` but it seems that overall pydantic release seems to be close, at least according to https://github.com/pydantic/pydantic/issues/5919 , so we better hurry with adding all the version restrictions and v1/v2 compatibility shimming.

jwodder:
@yarikoptic I've gotten the tests to pass with both v1 and v2 of pydantic, but the type-checking is failing with both.  I've posted https://github.com/pydantic/pydantic/discussions/5971 to ask for a solution.

yarikoptic:
@jwodder , I don't want the typing check oddity to stop us prepare in time.  Could we just skip that check inline in that location for now and get back to it whenever mypy folks reply?

satra:
closing this in relation to #203

PR #180, opened 2023-04-24
Make EthicsApproval non-optional

yarikoptic:
Per stand-up discussion.

Might need fixing etc

### DIFF
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -1215,8 +1215,8 @@ class CommonModel(DandiBaseModel):
         "protocols.io, or other DOIs).",
         nskey="dandi",
     )
-    ethicsApproval: Optional[List[EthicsApproval]] = Field(
-        None, title="Ethics approvals", nskey="dandi"
+    ethicsApproval: List[EthicsApproval] = Field(
+        title="Ethics approvals", nskey="dandi"
     )
     keywords: Optional[List[str]] = Field(
         None,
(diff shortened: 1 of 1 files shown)


djarecka:
just an idea to @bendichter comments. In datacite that would be under `Rights` I believe, [see p.27](https://schema.datacite.org/meta/kernel-4.4/doc/DataCite-MetadataKernel_v4.4.pdf)
The main field is "Free text, Provide a rights management statement for the resource or reference a service providing such information" and in addition you can have `rightsURI`, ` rightsIdentifier`, `rightsIdentifierScheme`, `schemeURI`

djarecka:
yes, this is optional now, but probably we can make this required and the identifier optional (to align also with @TheChymera point of view)?

The description says `Information about the ethics approval committee.`, but we could perhaps be more specific

satra:
> My suggestion would be either to require approving body (e.g. "NYU School of Medicine IACUC") or to require both approving body and protocol number. The protocol number on its own is in my opinion insufficient.

how about this? we change it to require approving body and an optional protocol number.

TheChymera:
@satra that's pretty good I'd say. Approving body would be a string, and nothing we need to validate or index, right?

satra:
@yarikoptic - do you want to pull this out of draft?

satra:
i'm going to close this as i want to separate the dandi requirements from type validation after pydantic 2.0 is merged.

PR #167, opened 2023-03-17
added funderidtype and support for funder and sponsor roles to datacite metadata

satra:
closes #168 

- added funderidtype
- added support for funders and sponsors

according to @djarecka's test the current release mints a doi for the failing metadata schema on dandiarchive, so unsure what the exact issue was. nonetheless this update should be used moving forward since people use funder/sponsor synonymously in dandiset metadata.

PR #181, opened 2023-05-19
CI: adding datacite validation 

djarecka:
added `publish=True` to one test to enable datacite validation. The doi will be published on https://doi.test.datacite.org/ as Findable, so will not be able to remove it!

I'm creating `doi` with `id = 000002` and random `version = 0.0.{random.randrange(0, 9999)}`

### DIFF
--- a/dandischema/tests/test_datacite.py
+++ b/dandischema/tests/test_datacite.py
@@ -14,7 +14,7 @@ from ..datacite import _get_datacite_schema, to_datacite
 from ..models import LicenseType, PublishedDandiset, RelationType, RoleType
 
 
-def datacite_post(datacite: dict, doi: str) -> None:
+def _datacite_post(datacite: dict, doi: str) -> None:
     """Post the datacite object and check the status of the request"""
 
     # removing doi in case it exists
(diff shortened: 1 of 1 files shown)


dandischema/tests/test_datacite.py
```diff
@@ -53,9 +53,9 @@ def schema() -> Any:
 
 @pytest.fixture(scope="function")
 def metadata_basic() -> Dict[str, Any]:
-    dandi_id_noprefix = f"000{random.randrange(100, 999)}"
+    dandi_id_noprefix = "000002"
     dandi_id = f"DANDI:{dandi_id_noprefix}"
-    version = "0.0.0"
+    version = f"0.0.{random.randrange(0, 9999)}"
```
satra, line 58: instead of random, perhaps use the current date+time in some form.

satra:
i think it would be good to limit the publishing test to only the platform that dandiarchive uses, so we don't have to test it for all options and limit the number of published dois on the test site.

djarecka:
@satra - you mean just to use for one of the linux containers we have, not adding a new one that is the same as dandiarchive uses, right?

satra:
closing this since #167 was merged.

PR #184, opened 2023-06-15
BF: make code not puke if there is no encodingFormat in asset metadata

### LINKED ISSUE #183: DANDI-API-C7: aggregate_assets_summary_task: KeyError encodingFormat since Feb
There was https://github.com/dandi/dandi-archive/pull/1490 which made such issues to bubble up, this is a list of sentry reports which seems keep mentioning this issue 

image

which IMHO should be given priority to be addressed even if just not to pollute our mailboxes with alerts, thus possibly hiding other, may be more important issues.

yarikoptic:
Although it is not clear how that can happen since it is not optional, worth 

- [ ] checking which assets lack it and why

Closes #183


satra:
closing in favor of https://github.com/dandi/dandi-archive/pull/1629

PR #185, opened 2023-06-19
add releaseNotes property to published dandiset

satra:
this will allow change of the UI to add a field alongside publishing for releaseNotes.

we should also consider adding optional `CHANGES` and `README` files to the Dandiset model.


dandischema/models.py
```diff
@@ -1484,6 +1484,11 @@ class PublishedDandiset(Dandiset, Publishable):
         description="Permalink to the Dandiset.",
         nskey="schema",
     )
+    releaseNotes: Optional[str] = Field(
+        None,
+        description="User notes for describing the release",
```
yarikoptic, line 1489: minor suggestion -- to shorten to

```suggestion
        description="Description of the release",
```
since could be a proud "Data curator" who submits the notes etc ;)

yarikoptic:
better add some record to some unittest to exercise this optional field, so there would be a "detector" for removing/breaking it.

dandischema/models.py
yarikoptic: FWIW, it is minor but I think we better use "my" version since would be consistent with other similar ones
```
dandischema/models.py:        None, description="The description of the equipment.", nskey="schema"
dandischema/models.py:        None, description="The description of the activity.", nskey="schema"
dandischema/models.py:        None, description="A brief description of the project.", nskey="schema"
dandischema/models.py:        None, description="A brief description of the session.", nskey="schema"
dandischema/models.py:        description="Genotype descriptor of participant or subject if available",
dandischema/models.py:        None, description="A description of the item.", nskey="schema"

```

since indeed could be not a "user" but someone else etc.

yarikoptic:
hm, I am confused why schemata version is not changed... we are adding a field so should get at least a patch version boost I guess in comparison to the latest release -- may be back then we had already some changes staged. Should rebase/push again and see what happens.  will do now

dandischema/models.py
```diff
@@ -1657,6 +1657,11 @@ class PublishedDandiset(Dandiset, Publishable):
         description="Permalink to the Dandiset.",
         json_schema_extra={"readOnly": True, "nskey": "schema"},
     )
+    releaseNotes: Optional[str] = Field(
+        None,
+        description="The description of the release",
+        nskey="schema",
```
yarikoptic, line 1663:
```suggestion
        json_schema_extra={"readOnly": True, "nskey": "schema"},
```

satra:
this will indeed have a schema change, but we only added the update to schema version in a relevant corresponding PR.

PR #193, opened 2023-10-23
Remove 3.7

yarikoptic:


yarikoptic:
type checking is clearly unrelated but we do have more of 3.7 to upgrade...

yarikoptic:
linters -- codespell
```
lint: commands[0]> codespell dandischema setup.py
dandischema/tests/data/metadata/meta_000004.json:6: re-use ==> reuse
dandischema/tests/data/metadata/meta_000004old.json:413: re-use ==> reuse
```

I will push to this PR just because so small

PR #196, opened 2023-11-09
Unpin versioningit version

jwodder:
`versioningit` is not meant to have an upper version bound (unless you're using custom methods, which you're not), as restricting it cuts you out of bugfixes & new features.  Also, the `{version}` field in the "format" table is now deprecated in favor of `{base_version}` (though `{version}` will continue to work).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
 # Setuptools version should match setup.py; wheel because pip will insert it noisily
-requires = ["setuptools >= 42.0.0", "versioningit ~= 0.3.0", "wheel"]
+requires = ["setuptools >= 42.0.0", "versioningit", "wheel"]
 build-backend = 'setuptools.build_meta'
 
 [tool.versioningit]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [build-system]
 # Setuptools version should match setup.py; wheel because pip will insert it noisily
-requires = ["setuptools >= 42.0.0", "versioningit ~= 0.3.0", "wheel"]
+requires = ["setuptools >= 42.0.0", "versioningit", "wheel"]
```
yarikoptic, line 3: shouldn't it be 
```suggestion
requires = ["setuptools >= 42.0.0", "versioningit >= 0.3.0", "wheel"]
```
then so some pip env with outdated version would not just silently proceed?

jwodder: Projects with `[build-system]` tables should always be built in isolated environments with freshly-installed dependencies (e.g., via [build](https://github.com/pypa/build)).  A "pip env with an outdated version" should have no effect.

yarikoptic: then we should not have versioning limit on `setuptools` either (consistency is good) and why did we get it there in the first place?

jwodder: `>= 42.0.0` just sets a lower bound (as opposed to `~= 0.3.0`, which also sets an upper bound), so it's harmless.

> why did we get it there in the first place?

Copy & paste from an earlier era.  When setuptools v42.0.0 was new, I believe it was possible for isolated build environments to be created with an older setuptools version automatically installed (because `venv` and `virtualenv` both bundle setuptools instead of downloading the latest version every time an environment is created), so specifying a minimum setuptools version forced an update if the environment's version was older.

PR #199, opened 2023-11-09
Rerun failing `test_dandimeta_datacite` tests

jwodder:
It frequently happens that a CI run fails because one instance of `test_dandimeta_datacite` on one platform got a 422 error.  The evidence suggests that this a problem on datacite's end, so this PR causes the test to simply be retried a few times until it succeeds.


yarikoptic (approved):
LGTM

PR #218, opened 2024-01-26
Allow dictionary representation of `Dandiset` to have extra attributes

candleindark:
This PR allows the dictionary representation of `Dandiset` to have extra attributes.

ATM, packages such as [dandi-archive](https://github.com/dandi/dandi-archive) construct `Dandiset` objects using `Dandiset.model_construct()` to allow extra attributes to be passed to the resulting `Dandiset` objects. While extra attributes can be passed to the resulting `Dandiset` objects this way, the dictionary representations of these `Dandiset` objects will not contain the extra attributes unless `Dandiset` is explicitly configured to allow extra attributes. 

[dandi-archive](https://github.com/dandi/dandi-archive) requires extra attributes passed in constructing a `Dandiset` object to be present in the dictionary representation of the object. This change in this PR explicitly configures `Dandiset` to allow extra attributes to meet this requirement, and meeting this requirement will eliminate some of the failures encountered in https://github.com/dandi/dandi-archive/pull/1823.

Please note, the change in this PR will modify the JSON schemata for `Dandiset` and `PublishedDandiset` slightly. Both schemata will contain the additional key of `"additionalProperties"` with the value of `true`. Please consider increase the version number of the JSON schemas to 0.6.6.

Note: This PR is a follow up to #203 to make a new version of dandi-schema with Pydantic 2.0 working for dandi-archive.

### DIFF
--- a/dandischema/consts.py
+++ b/dandischema/consts.py
@@ -1,4 +1,4 @@
-DANDI_SCHEMA_VERSION = "0.6.5"
+DANDI_SCHEMA_VERSION = "0.6.6"
 ALLOWED_INPUT_SCHEMAS = [
     "0.4.4",
     "0.5.1",
@@ -8,6 +8,7 @@ ALLOWED_INPUT_SCHEMAS = [
     "0.6.2",
     "0.6.3",
     "0.6.4",
+    "0.6.5",
 ]
 
 # ATM we allow only for a single target version which is current
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -9,6 +9,7 @@ from pydantic import (
     UUID4,
     AnyHttpUrl,
     BaseModel,
+    ConfigDict,
     EmailStr,
     Field,
     GetJsonSchemaHandler,
@@ -1401,6 +1402,8 @@ class CommonModel(DandiBaseModel):
 
 class Dandiset(CommonModel):
     """A body of structured information describing a DANDI dataset."""
+
+    model_config = ConfigDict(extra="allow")
 
     @field_validator("contributor")
     @classmethod


yarikoptic:
> ATM, packages such as [dandi-archive](https://github.com/dandi/dandi-archive) construct `Dandiset` objects using `Dandiset.model_construct()` to allow extra attributes to be passed to the resulting `Dandiset` objects.

could you please elaborate/point to what you mean since [I find no model_construct in dandi-archive codebase](https://github.com/search?q=repo%3Adandi%2Fdandi-archive+model_construct&type=code) and if there is some custom use like that I wonder if that is something we should aim to avoid.

candleindark:
Here is the reason why this PR is needed.

Before the deprecation of the [`json_dict` method](https://github.com/dandi/dandi-schema/blob/dbd704ace5b2abf5fb2718cca427c8fff76dcb62/dandischema/models.py#L358-L370), the method was a workaround to generate a dictionary representation of `DandiBaseModel` objects that's JSON serializable. I found that such workaround was not necessary and deprecated it in https://github.com/dandi/dandi-schema/pull/203. However, in resolving the failures in https://github.com/dandi/dandi-archive/pull/1823, I learned that dandi-archive attaches extra attributes to `Dandiset` objects (extra in the sense that these attributes are not defined as fields in the `Dandiset` model). To have these extra attributes included in the dictionary representation of the `Dandiset` objects generated by the facilities that provided by Pydantic, the model has to be marked explicitly to allow extra attributes, and the change in this PR is the explicit marking.

candleindark:
> > ATM, packages such as [dandi-archive](https://github.com/dandi/dandi-archive) construct `Dandiset` objects using `Dandiset.model_construct()` to allow extra attributes to be passed to the resulting `Dandiset` objects.
> 
> could you please elaborate/point to what you mean since [I find no model_construct in dandi-archive codebase](https://github.com/search?q=repo%3Adandi%2Fdandi-archive+model_construct&type=code) and if there is some custom use like that I wonder if that is something we should aim to avoid.

The two deprecated methods in this [line](https://github.com/dandi/dandi-archive/blob/880a1022c590e4ef6606e0f6a00e16a98e90da9b/dandiapi/api/services/version/metadata.py#L46) is the cause of the problem.

`unvalidated` calls `model_construct()` and `json_dict` calls `model_dump()`.

yarikoptic:
@jwodder and @satra please review on either you bless such changes.

jwodder:
If the JSON Schema has changed, we need to bump the schema version.

yarikoptic:
we would need to release again , so labeling accordingly

jwodder (approved):

candleindark:
Once https://github.com/dandi/dandi-schema/pull/221 is merged to master. This PR should be ready for merge.

candleindark:
@yarikoptic This one is ready to merge.

yarikoptic:
ok, let's see where it would get us

PR #220, opened 2024-01-30
Introduce ContentType as Blob vs Zarr and rename blobDateModified to contentDateModified

yarikoptic:
This is just an initial attempt open for discussion. @satra please chime in

I ran into "blobDateModified" in a zarr metadata and it raised my eyebrow since that is not really appropriate and confusing. Hence I decided to look into generalization.  I also thought that it would be valuable to make "type" of the content Asset points to explicit, although that could lead to inconsistencies since information is somewhat redundant with encodingFormat and potentially could also be deduced from contenUrl since we have different end points on S3, etc.

Nevertheless I think it might be better to make it explicit. Or at least we have to rename blobDateModified.

- ContenType name is quite suboptimal since there is a standard HTTP header https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type and thus we could potential confusion.

  But we should keep it a "Type" (not e.g. a Class) to be consistent with other type definitions among models.

  So the other part we could try to vary is "Content". Possible alternatives are "Object", "Data", "Resource"

- ATM we call all Zarrs just Zarr but it is a "ZarrFolder" really.  I wonder if it would be time to start to introduce differentiation here by making it "ZarrFolder", as later we might get "ZarrHDF5" or alike

### DIFF
--- a/dandischema/metadata.py
+++ b/dandischema/metadata.py
@@ -300,6 +300,18 @@ def migrate(
         if "schemaKey" not in obj:
             obj["schemaKey"] = "Dandiset"
         obj["schemaVersion"] = to_version
+    if version2tuple(schema_version) < version2tuple("0.7.0"):
+        if "blobDateModified" in obj:
+            obj["contentDateModified"] = obj.pop("blobDateModified")
+        # we need to deduce what type of content we have
+        obj["contentType"] = (
+            models.ContentType.Zarr
+            if (
+                obj.get("encodingFormat") == "application/x-zarr"
+                or obj.get("path", "").endswith(".zarr")
+            )
+            else models.ContentType.Blob
+        )
     return obj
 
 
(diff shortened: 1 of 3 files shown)


satra:
i'm fine with something like this being in the database, but not sure about changing the metadata model.  perhaps we can brainstorm when we meet.

satra (changes requested):
We have both `encodingFormat` and `dataType` - do we need yet another field or a clearer `encodingFormat`.

dandischema/metadata.py
```diff
@@ -300,6 +300,18 @@ def migrate(
         if "schemaKey" not in obj:
             obj["schemaKey"] = "Dandiset"
         obj["schemaVersion"] = to_version
+    if version2tuple(schema_version) < version2tuple("0.7.0"):
```
yarikoptic: here will need to become some future version now that 0.7.0 is out... we might want to introduce `next` branch alike to `next` in linux development, or more specifically `next-0.8.0` and position this PR against it so we could absorb meanwhile non-breaking changes

```suggestion
    if version2tuple(schema_version) < version2tuple("0.8.0"):
```

PR #225, opened 2024-02-07
Test against PR in dandi-cli to be compatible with pydantic v2

yarikoptic:
Now that dandi-schema is only pydantic 2.0 compatible, and while we are awaiting on dandi-archive to get ready -- there is probably no point to test against released/master dandi-cli -- just makes CI red hiding possible other fails.

We need to test against the dandi-cli in that PR for now


.github/workflows/test-dandi-cli.yml
```diff
@@ -75,6 +77,10 @@ jobs:
         if: ${{ matrix.version == 'release' }}
         run: pip install "dandi[test,extras]"
 
+      - name: Install dev dandi https://github.com/dandi/dandi-cli/tree/gh-1352 from https://github.com/dandi/dandi-cli/pull/1381
+        if: ${{ matrix.dandi-version == 'pydantic2' }}
```
jwodder, line 81:
```suggestion
        if: matrix.version == 'pydantic2'
```

`version`, not `dandi-version`, and you don't need the braces in an `if:`.

jwodder (changes requested):

.github/workflows/test-dandi-cli.yml
yarikoptic: doh - right. As `{{}}` -- why do we have them above?

yarikoptic: for now would do that one part of the fix:

```suggestion
        if: ${{ matrix.version == 'pydantic2' }}
```

yarikoptic:
no longer needed AFAIK

PR #232, opened 2024-03-16
Add ResourceType enum and associate with Resource model

### LINKED ISSUE #231: Resource type
https://github.com/dandi/dandi-schema/blob/e6cc5a06dd7f7e03e86316071ac9e541ff27d223/dandischema/models.py#L869-L900

Our current resource schema has no way of indicating what type of resource it is. This is currently a problem for me as I am trying to determine how many unique publications are associated with the data on DANDI. I need to infer the type based on the designated relation and the URL. It would be much easier for me if this infrormation were explicitly encoded. The Data Cite schema provides a [resourceType model](https://datacite-metadata-schema.readthedocs.io/en/4.5/appendices/appendix-1/resourceTypeGeneral/) that covers a lot of major resource categories. I propose we add this as an optional field to the Resource model.

bendichter:
fix #231

### DIFF
--- a/dandischema/consts.py
+++ b/dandischema/consts.py
@@ -1,4 +1,4 @@
-DANDI_SCHEMA_VERSION = "0.6.6"
+DANDI_SCHEMA_VERSION = "0.6.7"
 ALLOWED_INPUT_SCHEMAS = [
     "0.4.4",
     "0.5.1",
@@ -9,6 +9,7 @@ ALLOWED_INPUT_SCHEMAS = [
     "0.6.3",
     "0.6.4",
     "0.6.5",
+    "0.6.6",
 ]
 
 # ATM we allow only for a single target version which is current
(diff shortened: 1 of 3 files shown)


satra:
@bendichter - since this will create a new schema you should increment that to and make the necessary changes to consts.py

satra (changes requested):
overall looks great. left a comment about the consts.py change to increase schema version. can be a semver patch increase.

bendichter:
I'm not familiar with the merge protocol here. Is this all that's needed? Should I also add an entry to the `CHANGELOG.md` file?

satra (approved):
i'll approve this and wait for any comment from @jwodder before merging.

satra:
> I'm not familiar with the merge protocol here. Is this all that's needed? Should I also add an entry to the CHANGELOG.md file?

yes that's all that was needed. CHANGELOG is autogenerated+updated on release.

jwodder (approved):

bendichter:
OK great, now what's the process for propagating this change to the medator?

satra:
@bendichter - dandi-archive may need a schema bump.

satra:
however the auto release failed in this repo, so the version of auto may need to be updated to fix that.

jwodder:
@satra I suspect the `auto` failure has the same solution as https://github.com/nipy/heudiconv/pull/746.

Issue #189, opened 2023-07-27
is it possible to make ContactPerson be invalid without email?


yarikoptic:
We have https://github.com/dandisets/000108/blob/draft/dandiset.yaml where contact person is Lee , but he is listed without email.

```yaml
- affiliation: []
  includeInCitation: true
  name: Kamentsky, Lee
  roleName:
  - dcite:ContactPerson
  schemaKey: Person
- affiliation: []
  email: smarx@mit.edu
  identifier: 0000-0002-8209-0059
  includeInCitation: true
  name: Marx, Slayton
  roleName:
  - dcite:DataManager
  - dcite:DataCurator
  - dcite:Investigation
  - dcite:ProjectMember
  - dcite:Researcher
  - dcite:Software
  - dcite:Supervision
  - dcite:Visualization
  schemaKey: Person
```

PS I now made @slaytonmarx to be the contact person since AFAIK Lee is no longer with the project. Please adjust if that is not true

yarikoptic:
@satra @candleindark - how to express such a requirement in pydantic ?

candleindark:
> @satra @candleindark - how to express such a requirement in pydantic ?

Change https://github.com/dandi/dandi-schema/blob/57b503ade10140cec0a2c0cfab18e0dcba4977ee/dandischema/models.py#L823

to 

`email: Optional[EmailStr] = Field(..., json_schema_extra={"nskey": "schema"})` will do.

General reference for this can be found at https://docs.pydantic.dev/latest/concepts/models/#required-fields.

yarikoptic:
no -- it should be only if 
```
roleName:
  - dcite:ContactPerson
```

candleindark:
@yarikoptic I have submitted a draft PR to provide a model level validator to ensure that a contributor acting as a contact person has an email provided. Let me know if the PR is heading to the direction of solving the problem. If it is, I will finalize the PR by adjusting the tests, etc.

yarikoptic:
closed by 
- https://github.com/dandi/dandi-schema/pull/235

next time @candleindark please use `Closes #NUMBER` in PR description.

Issue #205, opened 2023-11-22
Overhaul models so that "unfinished" metadata can be represented without cheating Pydantic


jwodder:
(This is an accumulation of various things discussed elsewhere which need to be written down.)

Currently, it is possible for users of the Dandi Archive API to submit asset & Dandiset metadata that does not fully conform to the relevant dandischema model, and the Archive will accept, store, and return such flawed metadata, largely via use of `DandiBaseModel.unvalidated` (being replaced by Pydantic's `construct` in #203).  I believe part of the motivation for this is so that web users can fill in metadata over multiple sessions without having to fill in every field in a single sitting.

This results in API requests for asset & Dandiset metadata sometimes returning values that do not validate under dandischema's models; in particular, if a user of dandi-cli's Python API calls `get_metadata()` instead of `get_raw_metadata()`, the call may fail because our API returned metadata that doesn't conform to our own models (See dandi/dandi-cli#1205 and
dandi/dandi-cli#1363).

The dandischema models should therefore be overhauled as follows:

* There should exist models for representing Dandiset & asset metadata in a draft/unfinished state.  These models should accept all inputs that we want to accept from users (both via the API and the web UI), store in the database, and return in API responses.  (It is likely that such models will have all of their fields marked optional aside from the absolute bare minimum required.)
    * The `get_metadata()` methods of dandi-cli's Python API should return instances of these models.

* There should exist functionality for determining whether an instance of a draft/unfinished model meets all of the requirements for Dandiset publication.
    * One possible way to implement this would be to have separate models for published metadata that inherit from the draft models and make various fields non-optional.

CC @satra @dandi/dandiarchive @dandi/dandi-cli

yarikoptic:
I thought to argue that it would add another combinatorial dimension, but well, we kinda already have it half way that in terms of separation of "Common" vs "Publishable"

```
dandischema/models.py:class CommonModel(DandiBaseModel):
dandischema/models.py:class Dandiset(CommonModel):
dandischema/models.py:class BareAsset(CommonModel):
dandischema/models.py:class Asset(BareAsset):
dandischema/models.py:class Publishable(DandiBaseModel):
dandischema/models.py:class PublishedDandiset(Dandiset, Publishable):
dandischema/models.py:class PublishedAsset(Asset, Publishable):
```

the question then is really to "shift" validation into something like `.is_publishable()` and may be indeed mirroring as PublishableAsset/PublishableDandiset counterparts with more restrictive model?

[...]

PR #235, opened 2024-04-10
Add validator to ensure a contact person has email provided

candleindark:
This PR provides a model level validator to ensures that a contributor acting as a contact person has an email provided per request of https://github.com/dandi/dandi-schema/issues/189.

### DIFF
--- a/dandischema/tests/test_models.py
+++ b/dandischema/tests/test_models.py
@@ -694,32 +694,26 @@ def test_embargoedaccess() -> None:
     )
 
 
-def _non_contact_person_roles_args() -> List[List[RoleType]]:
-    """
-    Return a list of role lists that each do not contain the role of a contact person
-    """
-    return [[], [RoleType.Author, RoleType.DataCurator], [RoleType.Funder]]
+_NON_CONTACT_PERSON_ROLES_ARGS = [
+    [],
+    [RoleType.Author, RoleType.DataCurator],
+    [RoleType.Funder],
+]
 
-
-def _contact_person_roles_args() -> List[List[RoleType]]:
-    """
-    Return a list of role lists that each contains the role of a contact person
-    """
-    return [
-        role_lst + [RoleType.ContactPerson]
-        for role_lst in _non_contact_person_roles_args()
-    ]
+_CONTACT_PERSON_ROLES_ARGS = [
+    role_lst + [RoleType.ContactPerson] for role_lst in _NON_CONTACT_PERSON_ROLES_ARGS
+]
 
 
 class TestContributor:
-    @pytest.mark.parametrize("roles", _contact_person_roles_args())
+    @pytest.mark.parametrize("roles", _CONTACT_PERSON_ROLES_ARGS)
     def test_contact_person_with_email(self, roles: List[RoleType]) -> None:
         """
         Test creating a `Contributor` instance as a contact person with an email
         """
         Contributor(email="nemo@dandiarchive.org", roleName=roles)
 
-    @pytest.mark.parametrize("roles", _contact_person_roles_args())
+    @pytest.mark.parametrize("roles", _CONTACT_PERSON_ROLES_ARGS)
     def test_contact_person_without_email(self, roles: List[RoleType]) -> None:
         """
         Test creating a `Contributor` instance as a contact person without an email
@@ -729,14 +723,14 @@ class TestContributor:
         ):
             Contributor(roleName=roles)
 
-    @pytest.mark.parametrize("roles", _non_contact_person_roles_args())
+    @pytest.mark.parametrize("roles", _NON_CONTACT_PERSON_ROLES_ARGS)
     def test_non_contact_person_with_email(self, roles: List[RoleType]) -> None:
         """
         Test creating a `Contributor` instance as a non-contact person with an email
         """
         Contributor(email="nemo@dandiarchive.org", roleName=roles)
 
-    @pytest.mark.parametrize("roles", _non_contact_person_roles_args())
+    @pytest.mark.parametrize("roles", _NON_CONTACT_PERSON_ROLES_ARGS)
     def test_non_contact_person_without_email(self, roles: List[RoleType]) -> None:
         """
         Test creating a `Contributor` instance as a non-contact person without an email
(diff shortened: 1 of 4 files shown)


candleindark:
@yarikoptic @satra I think this is good to go except that you may want to adjust the dandi-cli due to the change brought by this PR.

dandischema/tests/test_models.py
```diff
@@ -689,3 +692,53 @@ def test_embargoedaccess() -> None:
...
+    """
+    Return a list of role lists that each contains the role of a contact person
+    """
+    return [
+        role_lst + [RoleType.ContactPerson]
+        for role_lst in _non_contact_person_roles_args()
+    ]
+
+
+class TestContributor:
```
satra, line 714: out of curiosity, any specific reason to have a class here?

satra:
@yarikoptic - should the CLI fail if this validation doesn't pass. i.e. do we force the user to adjust the metadata through the UI to fix validation issues and redownload the dandiset.yaml, before data can be uploaded?

dandischema/tests/test_models.py
candleindark: I use that to group all the tests related to `Contributor`. Since there are other tests for testing different aspects/entities in `test_models.py`, I think it is beneficial to put all the given tests into an easily identifiable group represented by the class.

yarikoptic:
> @yarikoptic - should the CLI fail if this validation doesn't pass. i.e. do we force the user to adjust the metadata through the UI to fix validation issues and redownload the dandiset.yaml, before data can be uploaded?

no, I do not think we anyhow concern ourselves with validity of the `dandiset.yaml` upon upload. Filed

- https://github.com/dandi/dandi-cli/issues/1427

dedicated to that. This PR is pretty much orthogonal IMHO.

dandischema/tests/test_models.py
```diff
@@ -689,3 +692,53 @@ def test_embargoedaccess() -> None:
...
+            pydantic.ValidationError, match="Contact person must have an email address"
+        ):
+            Contributor(roleName=roles)
+
+    @pytest.mark.parametrize("roles", _non_contact_person_roles_args())
+    def test_non_contact_person_with_email(self, roles: List[RoleType]) -> None:
+        """
+        Test creating a `Contributor` instance as a non-contact person with an email
+        """
+        Contributor(email="nemo@dandiarchive.org", roleName=roles)
```
yarikoptic, line 737: IMHO there is little point to separate out this one from `test_contact_person_with_email` -- we do not need to do full combinatoric combination 2x2:  let's merge into the 

```python

    @pytest.mark.parametrize("roles", _contact_person_roles_args() + _non_contact_person_roles_args())
    def test_person_with_email(self, roles: List[RoleType]) -> None:
        """
        Test creating a `Contributor` instance with an email
        """
        Contributor(email="nemo@dandiarchive.org", roleName=roles)
```

NB subject to also getting rid unnecessary function invocation - could be just const lists of roles.

yarikoptic (changes requested):
Let's simplify / shorten testing a bit. See individual comments.

yarikoptic:
Only loosely related to this PR (I guess) -- whenever we approach `linkml.io` expression, we would need to define `ContactPerson` as subclass of a `Person` and make `email` mandatory... Or is there some similarly flexible way to define conditional validations in linkml @satra @candleindark ?

dandischema/tests/test_models.py
```diff
@@ -689,3 +692,53 @@ def test_embargoedaccess() -> None:
...
+
+
+def _contact_person_roles_args() -> List[List[RoleType]]:
+    """
+    Return a list of role lists that each contains the role of a contact person
+    """
+    return [
+        role_lst + [RoleType.ContactPerson]
+        for role_lst in _non_contact_person_roles_args()
+    ]
```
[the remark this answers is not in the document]
candleindark: You are right. The tests were modifying the lists, but that's no longer the case. Thanks.

dandischema/tests/test_models.py
candleindark: I separated the tests cases to make it easier to locate bugs, but this way is indeed more compact.

candleindark:
> Only loosely related to this PR (I guess) -- whenever we approach `linkml.io` expression, we would need to define `ContactPerson` as subclass of a `Person` and make `email` mandatory... Or is there some similarly flexible way to define conditional validations in linkml @satra @candleindark ?

I don't know how at this point. I was actually thinking about this. The flexibility of the current arrangement comes from executing logic in Python defined as a validator for the `Contributor` Pydantic model. In fact, the logic currently is not even exported to the JSON schema. We may indeed need to use some other mechanism to express the given relationship in linkml.

sneakers-the-rat:
in linkml you'd use [rules](https://linkml.io/linkml/schemas/advanced.html#rules)

so that would be something like this

```yaml
classes:
  Person:
    slots:
      - role
      - email
    rules:
      - preconditions:
          slot_conditions:
            role:
              equals_string: ContactPerson
        postconditions:
          slot_conditions:
            email:
              required: true
```

which we would need to modify the pydantic generator to support rules with validators to make

```python
@model_validator(mode='after')
def rule_0(self):
    # some type inference would have to happen at generation time, but something like...
    if self.role is not None and 'ContactPerson' in self.role:
        assert self.email is not None
    
```

and then modify the pydantic model schema with `json_schema_extra`

```json
{
"if": {
  "properties": {"role": {"contains": { "const": "ContactPerson" }}},
},
"then": {
  "properties": {"email": {"required": true }}
}
```

but i think you're probably right that it would be cleaner to use a subclass with a [type designator](https://linkml.io/linkml/schemas/type-designators.html)

```yaml
enums:
  RoleName:
    permissible_values:
      ContactPerson:

classes:
  Person:
    attributes:
      role:
        range: RoleName
        designates_type: true
      email:
        range: string
        required: false
        
  ContactPerson:
    slot_usage:
      email:
        required: true
```

and reconfigure pydanticgen to make use of [discriminator](https://docs.pydantic.dev/latest/concepts/fields/#field-representation) in fields.

satra:
when we get to linkml, let's separate out validation logic (which changes over time), from schema and schema-types which are likely more stable. this is also going to be relevant to partial models (before/after upload, before/after publish, etc.,.).

yarikoptic:
good -- I think we are arriving at cleaner separation of "validation" against desired requirements vs "validation" of the model... as somewhat being also discussed in

- https://github.com/dandi/dandi-schema/issues/205

But as such I would take this PR as ready and merge it.

PR #236, opened 2024-04-13
Set the `$schema` key with the schema dialect

candleindark:
This PR closes #216.

PR #238, opened 2024-04-24
GitHub tests.yml CI: explicitly specify macos-12 and use python 3.11 to test against dandi-cli

yarikoptic:
The arm64  one seems to lack pythons 3.8 and 3.9. Ref: https://github.com/dandi/dandi-schema/pull/236#issuecomment-2073369726

### DIFF
--- a/.github/workflows/test-dandi-cli.yml
+++ b/.github/workflows/test-dandi-cli.yml
@@ -25,10 +25,10 @@ jobs:
         python:
           # Use the only Python which is ATM also used by dandi-api
           # - 3.7
-          - 3.8
+          # - 3.8
           # - 3.9
           # - '3.10'
-          # - '3.11'
+          - '3.11'
         version:
           - master
           - release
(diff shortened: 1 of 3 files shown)


yarikoptic:
so it didn't help -- still goes for arm64:
image

for now let's just keep good old macos-12 -- since not yet announced "deprecated" , I think we would have some time.

yarikoptic:
ok -- for macos-12 ones we now got tests actually ran without bailing, so provides part of the solution. But I do not get why all the rest fail whenever they do not fail e.g. in https://github.com/dandi/dandi-schema/pull/236 . Need to have a fresh look tomorrow.

candleindark:
> But I do not get why all the rest fail whenever they do not fail e.g. in #236 . Need to have a fresh look tomorrow.

All the failed tests seem to be failing the email requirement put in #235 which is very strange. I don't know why changing Python version has anything to do with these errors at this point.

yarikoptic:
ok, we need to undig CI from redness. I think this PR should be a step forward so I will merge

PR #242, opened 2024-05-23
Draft: Relax timestamp timezone validation for startDate and endDate fields on NWB file formats

aaronkanzer:
Relates to https://github.com/dandi/dandi-archive/issues/1944

and changes made for NWB in https://github.com/NeurodataWithoutBorders/pynwb/pull/1886

Observed in failed dandiset validation in staging environment here: https://gui-staging.dandiarchive.org/dandiset/213840

Cc @candleindark @yarikoptic @satra @bendichter

### DIFF
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -1203,14 +1203,16 @@ class Activity(DandiBaseModel):
     _ldmeta = {"rdfs:subClassOf": ["prov:Activity", "schema:Thing"], "nskey": "dandi"}
 
     @field_validator('startDate', 'endDate', mode='before')
-    def parse_datetime(cls, value):
+    def parse_datetime(cls, value: Any) -> Optional[datetime]:
         if isinstance(value, str):
             try:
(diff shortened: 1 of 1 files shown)


satra:
@aaronkanzer - i'm not sure this is necessary. i believe pydantic default validator for datetime works fine. at least on the example ben sent.

aaronkanzer:
> @aaronkanzer - i'm not sure this is necessary. i believe pydantic default validator for datetime works fine. at least on the example ben sent.

Hmmm, yeah accordingly to the docs _it should_ work fine (e.g. pydantic handling different formats of `datetime`), but it is still failing 🤔 

Perhaps a different root cause....looking further...

bendichter:
Yes, this is what has been confusing us. We don't see any logic in the dandi schema that would seem to require timezones for this field

aaronkanzer:
Closed for now -- as it doesn't seem that this is the root cause error of validation for certain `datetime` timestamps.

PR #245, opened 2024-06-21
Use discriminated unions to provide more helpful error messages

### LINKED ISSUE #244: Use discriminated unions to improve validation errors
We recently saw these metadata validation errors on a dandiset in production https://github.com/dandi/dandi-archive/issues/1958. These are the errors that were reported:

```
contributor: String should match pattern '^([\w\s\-\.']+),\s+([\w\s\-\.']+)$'
contributor: Input should be 'Organization'
contributor: String should match pattern 'https://ror.org/[a-z0-9]+$'
```

The invalid contributor in question turned out to be a `Person` with an invalid `name` field; in other words, the first validation error was the actual issue, while the other two were not relevant and somewhat misleading. What's happening here is pydantic has no idea from a validation perspective whether the object is intended to be a `Person` or an `Organization` , as `contributor` is of type `List[Union[Person, Organization]]`, so it's checking both cases (i.e., first it validates the object as if it were a `Person` and gets the first error, then it validates it as a `Organization` and gets the other two errors).

I propose that we use [discriminated unions](https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions) on the `schemaKey` field of each pydantic model so we can avoid this in the future. This would allow pydantic to scope down the validation to the specific type of the object based on its `schemaKey`. If we had this in the above mentioned scenario, pydantic would have recognized that the invalid contributor is supposed to be a `Person` and would not have reported the additional misleading validation errors that assume it's an `Organization`.

mvandenburgh:
Fixes #244 

This change allows Pydantic to generate more useful validation errors by making use of [discriminated unions](https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions) where possible.

### DIFF
--- a/dandischema/consts.py
+++ b/dandischema/consts.py
@@ -1,4 +1,4 @@
-DANDI_SCHEMA_VERSION = "0.6.7"
+DANDI_SCHEMA_VERSION = "0.6.8"
 ALLOWED_INPUT_SCHEMAS = [
     "0.4.4",
     "0.5.1",
@@ -10,6 +10,7 @@ ALLOWED_INPUT_SCHEMAS = [
     "0.6.4",
     "0.6.5",
     "0.6.6",
+    "0.6.7",
 ]
 
 # ATM we allow only for a single target version which is current
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -22,6 +22,9 @@ from pydantic import (
 )
 from pydantic.json_schema import JsonSchemaValue
 from pydantic_core import CoreSchema
+from typing_extensions import (
+    Annotated,  # TODO: import from `typing` when Python 3.8 support is dropped
+)
 from zarr_checksum.checksum import InvalidZarrChecksum, ZarrDirectoryDigest
 
 from .consts import DANDI_SCHEMA_VERSION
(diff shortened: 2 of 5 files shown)


mvandenburgh:
There's some test failures happening here that I have to look into, marking this as draft for now.

candleindark:
The changes look good to me.

The JSON schemas have changed though mostly because of the inclusion of the [discriminator object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#discriminator-object) as mentioned in [the discriminated union docs](https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions). We may need to increase the JSON schema version, but I don't expect much perceivable difference for the end users.

@yarikoptic @satra Should me up the JSON schema version for this one? We didn't up the JSON schema version for https://github.com/dandi/dandi-schema/pull/236, so the changes in JSON schemas from that PR are yet to be included in the latest published schemas.

yarikoptic:
my personal take is that for any change upon release we should up the json version, otherwise it all becomes ambiguous and possibly hard to troubleshoot later on.  "explicit better than implicit"!

mvandenburgh:
Is there something I need to add to the PR to bump the schema version?

satra:
just added labels so we can release with this. see the consts.py file. that's where the schema version will need to be incremented for the test to pass.

mvandenburgh:
> just added labels so we can release with this. see the consts.py file. that's where the schema version will need to be incremented for the test to pass.

Thanks, I bumped the version in `consts.py`.

satra:
@candleindark - could you please provide your feedback if any or approve this PR?

candleindark (approved):
Everything looks fine to me.

candleindark:
Screenshot 2024-07-10 at 1 38 57 PM

Any idea about these tests?  I can't locate them in the github workflow files. The only place we use `macos-latest` is in `test-dandi-cli.yml`, these tests don't seem to be defined by that file.

satra:
@candleindark - fixed the branch merge requirements. merging.

satra:
thanks @mvandenburgh for this PR. will you update the server to this release schema and python library?

PR #248, opened 2024-08-13
Declare `typing_extensions` dependency

jwodder:
Use of `typing_extensions` was reintroduced to the code in commit 0d381179 (#203), but `typing_extensions` was not added to the runtime requirements.  This PR fixes that.


yarikoptic:
Thank you!

PR #251, opened 2024-08-21
Properly handle failed requests to retrieve the schema

danlamanna:
We ran into this a week or so ago when github had their outage. This should provide a clearer exception than the `JSONDecodeError` we saw.


candleindark:
LGTM. Thank @danlamanna. @yarikoptic, merge? The [test-dandi-cli (ubuntu-latest, 3.8, dandi-devel, master)](https://github.com/dandi/dandi-schema/actions/runs/10489304852/job/29169578332#logs) is hung. I couldn't restart it.

yarikoptic:
FWIW, that test stall is not new:

- https://github.com/dandi/dandi-cli/issues/1450

I restarted that run just in case.

PR #252, opened 2024-08-30
Add support for detection and addition of ome/ngff "standard" into assets summary

yarikoptic:
TODOs

- [x] test

we will need to release so that dandi-archive could take advantage of it

### DIFF
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -817,8 +817,7 @@ bids_standard = StandardsType(
 
 ome_ngff_standard = StandardsType(
     name="OME/NGFF Standard",
-    # Seems no dedicated RRID, only for OME itself?
-    # identifier="RRID:SCR_???",
+    identifier="DOI:10.25504/FAIRsharing.9af712",
 ).model_dump(mode="json", exclude_none=True)
 
 
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -815,6 +815,12 @@ bids_standard = StandardsType(
     identifier="RRID:SCR_016124",
 ).model_dump(mode="json", exclude_none=True)
 
+ome_ngff_standard = StandardsType(
+    name="OME/NGFF Standard",
+    # Seems no dedicated RRID, only for OME itself?
+    # identifier="RRID:SCR_???",
+).model_dump(mode="json", exclude_none=True)
+
 
 class ContactPoint(DandiBaseModel):
     email: Optional[EmailStr] = Field(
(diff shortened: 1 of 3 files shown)


dandischema/models.py
```diff
@@ -807,6 +807,12 @@ class StandardsType(BaseType):
     identifier="RRID:SCR_016124",
 ).model_dump(mode="json", exclude_none=True)
 
+ome_ngff_standard = StandardsType(
+    name="OME/NGFF Standard",
+    # Seems no dedicated RRID, only for OME itself?
+    # identifier="RRID:SCR_???",
```
satra, line 813: @joshmoore - any thoughts here on identifier for the ome/ngff standard? or could we mint an RRID for it?

joshmoore: :wave: @satra:

 * No identifier (I know of) in use at the moment.
 * Does it need to be an RRID?
 * Closest I know of is https://fairsharing.org/FAIRsharing.9af712 (https://doi.org/10.25504/FAIRsharing.9af712)

satra: thanks @joshmoore - doesn't have to be RRID, we can use any persistent identifier. we can use the doi.

joshmoore: Let me run it by a few folks Monday AM, and I'll get back to you.

joshmoore: Sorry for the wait, @satra. No other identifiers from our side. If the DOI works for the moment, then :+1: (How hard would this be to change in the future?)

satra: hyphen is fine. we can change it in the future. will just need to encode the change and the mapping.

joshmoore: Sounds good.

yarikoptic: Thank you gentleman, I added DOI identifier.

dandischema/models.py
```diff
@@ -807,6 +807,11 @@ class StandardsType(BaseType):
     identifier="RRID:SCR_016124",
 ).model_dump(mode="json", exclude_none=True)
 
+ome_ngff_standard = StandardsType(
+    name="OME/NGFF Standard",
```
satra, line 811:
```suggestion
    name="OME-NGFF Standard",
```

satra: switched to @joshmoore request for hyphen

yarikoptic:
cli fail unrelated
```
=========================== short test summary info ============================
FAILED tests/test_dandiarchive.py::test_parse_api_url[https://gui.dandiarchive.org/#/dandiset/000001-parsed_url0] - requests.exceptions.ConnectionError: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))
= 1 failed, 706 passed, 1 skipped, 7 xfailed, 9 xpassed, 26 warnings in 720.33s (0:12:00) =
Error: Process completed with exit code 1.
```

Let's proceed.

PR #257, opened 2024-10-31
Refactoring model so we could avoid shortcoming of linkml model 

yarikoptic:
Shortcoming in question:
- https://github.com/orgs/linkml/discussions/2154

AFAIK this should result in no changes necessary to metadata records since none of those "features" was used. So even though we break the model, I think we could just go with "patch" level progress forward.

Individual commits have more information.

TODOs also
- [ ] do something about genotype "Identifier" being abused to encapsulate records describing the genotype, hence better be parsed etc ??? file a separate issue.

### DIFF
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -1262,7 +1262,7 @@ class PublishActivity(Activity):
 
 
 class Locus(DandiBaseModel):
-    identifier: Union[Identifier, List[Identifier]] = Field(
+    identifier: Identifier = Field(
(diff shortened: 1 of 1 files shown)


satra:
we should check the biological details of this change across organisms (nematode to humans).

yarikoptic:
@candleindark please review test fails and what needs to be done to make it green

satra:
did anyone check if an updated metamodel was released in linkml as chris mentioned in that original post on linkml.

Issue #259, opened 2024-11-04
Include explicitly dandiset ID and version within citation/datacite record


yarikoptic:
ATM our records look like

```shell
❯ (set -x; doiref 10.48324/dandi.000485/0.241014.2127 )
+zsh:47> doiref 10.48324/dandi.000485/0.241014.2127
+doiref:2> doi=+doiref:2> echo 10.48324/dandi.000485/0.241014.2127
+doiref:2> doi=+doiref:2> sed -e 's,https*://.*doi\.org/,,g'
+doiref:2> doi=10.48324/dandi.000485/0.241014.2127 
+doiref:3> curl --silent -L -d '' --header 'Accept: application/x-bibtex; charset=utf-8' https://doi.org/10.48324/dandi.000485/0.241014.2127
+doiref:4> sed -e s,%2F,/,g
+doiref:4> xclip -i
+doiref:5> xclip -o
@misc{https://doi.org/10.48324/dandi.000485/0.241014.2127,
  doi = {10.48324/DANDI.000485/0.241014.2127},
  url = {https://dandiarchive.org/dandiset/000485/0.241014.2127},
  author = {Haesemeyer, Martin},
  title = {Thermal Plaid Experiments},
  publisher = {DANDI Archive},
  year = {2024}
}
```

so neither title nor explicit field include identifier (000485, or `DANDI:000485`) or version (`0.241014.2127`).

@djarecka found already that there is `Version` attribute we should add to datacite records, but it does not then appear in bibtex rendering.  She also discovered that "identifiers" was removed in 4.5 from new JSON schema version since it is not really in original schema.  Apparently the entire datacite jsonschema serialization is a fruit of "manual labor":

- https://github.com/inveniosoftware/datacite/issues/101#issuecomment-2452613295

and is not "official":

- https://github.com/datacite/schema/issues/149#issuecomment-2452505803

I will file a separate issue/PR on `identifiers` and overall there is more to figure out how to improve our citation record.

PR #260, opened 2024-11-04
Add "version" to datacite record and bundle datacite json serializations

yarikoptic:
folded in a typo fix

partially addresses

- #259 

This PR also bundles (**all** known versions) datacite json schemas.


yarikoptic:
all tests are failing with similar

```
E   requests.exceptions.ProxyError: HTTPSConnectionPool(host='raw.githubusercontent.com', port=443): Max retries exceeded with url: /datacite/schema/732cc7ef29f4cad4d6adfac83544133cd57a2e5e/source/json/kernel-4.3/datacite_4.3_schema.json (Caused by ProxyError('Unable to connect to proxy', NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f87bab74170>: Failed to establish a new connection: [Errno 111] Connection refused')))
```

I think it might be worth just shipping copies of desired verion(s) of datacite serializations right in here, in particular since we are referencing a specific to commit version.

yarikoptic:
dandi-cli versions failing to install with

> ERROR: Package 'dandi' requires a different Python: 3.8.18 not in '>=3.9'

since here we have not yet dropped 3.8 and in dandi-cli already did...

djarecka:
i'm wondering if we really want to keep all the version from these repositories, or decide on one that we want to match. Or perhaps we want to create our own?

yarikoptic:
we had to do `diff` archaeological expedition so I thought to accumulate them all in one common space.  They are relatively lightweight so I saw no reservation from keeping them all. 
```
❯ du -sck dandischema/datacite/schema/*
16	dandischema/datacite/schema/datacite-4.3-17-gaa5db56.json
20	dandischema/datacite/schema/datacite-4.3-72-g732cc7e.json
20	dandischema/datacite/schema/inveniosoftware-4.3-66-g24fc2ba.json
24	dandischema/datacite/schema/inveniosoftware-4.5-81-g160250d.json
4	dandischema/datacite/schema/README.md
84	total
```
Moreover, with exposing version in the code we could try one or another easily and consistently across versions.

PR #261, opened 2024-11-04
Upgrade to datacite v4.5 serialization from inveniosoftware

yarikoptic:
Some extra info on "dichotomy" of datacite json serializations could be found in
- https://github.com/dandi/dandi-schema/issues/259

and there in.  Part of this effort is RFing use of "identifiers" (removed in 4.5 serialization)

Might want to wait for addressing need to include schemas as stated in https://github.com/dandi/dandi-schema/pull/260#issuecomment-2454982938 

@djarecka  you were looking into upgrades also on Friday -- what other items were due?

### DIFF
--- a/dandischema/datacite/__init__.py
+++ b/dandischema/datacite/__init__.py
@@ -79,11 +79,6 @@ def to_datacite(
         attributes["event"] = "publish"
 
     attributes["alternateIdentifiers"] = [
-        # TODO: the first element is ignored, not sure how to fix it...
-        {
-            "alternateIdentifier": f"https://doi.org/{meta.doi}",
-            "alternateIdentifierType": "DOI",
-        },
         {
             "alternateIdentifier": f"https://identifiers.org/{meta.id}",
             "alternateIdentifierType": "URL",
--- a/dandischema/datacite/__init__.py
+++ b/dandischema/datacite/__init__.py
@@ -172,7 +172,7 @@ def to_datacite(
                 contr_dict["affiliation"] = []
             if getattr(contr_el, "identifier"):
                 orcid_dict = {
-                    "nameIdentifier": contr_el.identifier,
+                    "nameIdentifier": f"https://orcid.org/{contr_el.identifier}",
                     "nameIdentifierScheme": "ORCID",
                     "schemeUri": "https://orcid.org/",
                 }
--- a/dandischema/datacite/__init__.py
+++ b/dandischema/datacite/__init__.py
@@ -255,7 +255,7 @@ def to_datacite(
 
 
 @lru_cache()
-def _get_datacite_schema(version_id: str = "datacite-4.3-17-gaa5db56") -> Any:
+def _get_datacite_schema(version_id: str = "inveniosoftware-4.5-81-g160250d") -> Any:
     """Load datacite schema based on the version id provided."""
     schema_folder = Path(__file__).parent / "schema"
     return json.loads((schema_folder / f"{version_id}.json").read_text())
--- a/dandischema/datacite/__init__.py
+++ b/dandischema/datacite/__init__.py
@@ -101,7 +101,13 @@ def to_datacite(
     attributes["descriptions"] = [
         {"description": meta.description, "descriptionType": "Abstract"}
     ]
-    attributes["publisher"] = "DANDI Archive"
+    attributes["publisher"] = {
+        "name": "DANDI Archive",
+        "schemeUri": "https://scicrunch.org/resolver/",
+        "publisherIdentifier": "https://scicrunch.org/resolver/RRID:SCR_017571",
+        "publisherIdentifierScheme": "RRID",
+        "lang": "en",
+    }
     attributes["publicationYear"] = str(meta.datePublished.year)
     # not sure about it dandi-api had "resourceTypeGeneral": "NWB"
     attributes["types"] = {
@@ -113,7 +119,7 @@ def to_datacite(
     # assuming that all licenses are from SPDX?
     attributes["rightsList"] = [
         {
-            "schemeURI": "https://spdx.org/licenses/",
+            "schemeUri": "https://spdx.org/licenses/",
             "rightsIdentifierScheme": "SPDX",
             "rightsIdentifier": el.name,
         }
@@ -150,7 +156,7 @@ def to_datacite(
         contr_dict: Dict[str, Any] = {
             "name": contr_el.name,
             "contributorName": contr_el.name,
-            "schemeURI": "orcid.org",
+            "schemeUri": "orcid.org",
         }
         if isinstance(contr_el, Person):
             contr_dict["nameType"] = "Personal"
--- a/dandischema/datacite/__init__.py
+++ b/dandischema/datacite/__init__.py
@@ -78,16 +78,19 @@ def to_datacite(
     if publish:
         attributes["event"] = "publish"
 
-    attributes["identifiers"] = [
+    attributes["alternateIdentifiers"] = [
         # TODO: the first element is ignored, not sure how to fix it...
-        {"identifier": f"https://doi.org/{meta.doi}", "identifierType": "DOI"},
         {
-            "identifier": f"https://identifiers.org/{meta.id}",
-            "identifierType": "URL",
+            "alternateIdentifier": f"https://doi.org/{meta.doi}",
+            "alternateIdentifierType": "DOI",
         },
         {
-            "identifier": str(meta.url),
-            "identifierType": "URL",
+            "alternateIdentifier": f"https://identifiers.org/{meta.id}",
+            "alternateIdentifierType": "URL",
+        },
+        {
+            "alternateIdentifier": str(meta.url),
+            "alternateIdentifierType": "URL",
         },
     ]
 
--- a/dandischema/datacite/tests/test_datacite.py
+++ b/dandischema/datacite/tests/test_datacite.py
@@ -452,13 +452,6 @@ def test_datacite_publish(metadata_basic: Dict[str, Any]) -> None:
                 "doi": f"10.80507/dandi.{dandi_id_noprefix}/{version}",
                 "alternateIdentifiers": [
                     {
-                        "alternateIdentifier": (
-                            f"https://doi.org/10.80507"
-                            f"/dandi.{dandi_id_noprefix}/{version}"
-                        ),
-                        "alternateIdentifierType": "DOI",
-                    },
-                    {
                         "alternateIdentifier": f"https://identifiers.org/{dandi_id}/{version}",
                         "alternateIdentifierType": "URL",
                     },
--- a/dandischema/datacite/tests/test_datacite.py
+++ b/dandischema/datacite/tests/test_datacite.py
@@ -150,7 +150,16 @@ def test_datacite(dandi_id: str, schema: Any) -> None:
                     1,
                     {"description": "testing", "descriptionType": "Abstract"},
                 ),
-                "publisher": (None, "DANDI Archive"),
+                "publisher": (
+                    None,
+                    {
+                        "name": "DANDI Archive",
+                        "publisherIdentifier": "https://scicrunch.org/resolver/RRID:SCR_017571",
+                        "publisherIdentifierScheme": "RRID",
+                        "schemeUri": "https://scicrunch.org/resolver/",
+                        "lang": "en",
+                    },
+                ),
                 "rightsList": (
                     1,
                     {"rightsIdentifierScheme": "SPDX", "rightsIdentifier": "CC_BY_40"},
@@ -423,7 +432,7 @@ def test_datacite_publish(metadata_basic: Dict[str, Any]) -> None:
                         "givenName": "A_first",
                         "name": "A_last, A_first",
                         "nameType": "Personal",
-                        "schemeURI": "orcid.org",
+                        "schemeUri": "orcid.org",
                     }
                 ],
                 "creators": [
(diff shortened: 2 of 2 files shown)


dandischema/datacite.py
```diff
@@ -87,7 +90,13 @@ def to_datacite(
     attributes["descriptions"] = [
         {"description": meta.description, "descriptionType": "Abstract"}
     ]
-    attributes["publisher"] = "DANDI Archive"
+    attributes["publisher"] = {
+        "name": "DANDI Archive",
+        "schemeUri": "https://scicrunch.org/resolver/",
+        "publisherIdentifier": "https://scicrunch.org/resolver/RRID:SCR_017571",
```
satra, line 96: is this really the identifier? or just `SCR_017571`

satra (commented):
some memory bank item says something about `alternateIdentifiers` that we implemented and then reverted because the API service did not match the schema requirements. so simply putting a flag here for checking.

dandischema/datacite.py
yarikoptic: in all of the examples https://datacite-metadata-schema.readthedocs.io/en/4.5/properties/publisher/#a-publisheridentifier they used a URL, so I just followed the trend... it is indeed quite an odd schema

yarikoptic:
FWIW -- search comes empty https://github.com/search?q=repo%3Adandi%2Fdandi-schema%20alternateIdentifiers&type=code  :-/    I could be wrong but I think we are unittesting this against jsonschema and test fabric of datacite, right @djarecka ?

djarecka:
I also just want to clarify, that the datacite documentation v4.5 still have `Identifier`: https://datacite-metadata-schema.readthedocs.io/_/downloads/en/4.5/pdf/ (page 11)

yarikoptic:
but there is no `identifiers` (note: plural) defined in the schema, hence switching to `relateIdentifiers` as a list of `relateIdentifier` entities which **is** defined in the schema (`source/meta/kernel-4/metadata.xsd:        <xs:element name="alternateIdentifiers" minOccurs="0">`).

djarecka:
yes, you're right, `identifier` is mandatory but singular, see your point.
But this actually has not changed since v4.3 (i mean the official documentation), so again the validations they used are not fully compatible with their documentation.

yarikoptic:
yes, as I think we ran into before on a number of occasions they have differences between schema, documentation, and jsonschema serializations since documentation and jsonschema are not automatically produced from the schema unfortunately IIRC. Thus all the differences.  I think it would be best to rely on schema (not documentation) as the "ultimate ground truth" with the hope that jsonschema would eventually become autogenerated/fully reflective of the original schema.

FWIW -- tests pass here. That means that we are in compliance with test fabric of datacite and can proceed right?

yarikoptic:
FTR, here  is
- datacite API doc which says that it also accepts `"identifiers"` as alternative to `"alternateIdentifiers"`: https://support.datacite.org/docs/what-is-the-identifiers-attribute-in-the-rest-api 
- "Identifier" mentioned: https://datacite-metadata-schema.readthedocs.io/en/4.5/properties/identifier/

satra:
> datacite API doc which says that it also accepts "identifiers" as alternative to "alternateIdentifiers": https://support.datacite.org/docs/what-is-the-identifiers-attribute-in-the-rest-api

@yarikoptic - so are you going to revert back to identifiers? was there a rationale for changing it?

yarikoptic:
no, I do not see the point of reverting to accommodate API interface for which we do not have a validator.

The rationale for my change is to be able to validate our metadata against jsonschema of datacite record we have access to.  Do you have concerns/reservations?

satra:
as i said, in the past we did this against the schema and they said it's not supported in the api. since we test against the api, that's what i would hold as our functional target. i believe this happened as we were suggesting changes to the schema. at least at that time, almost 5 years ago, they considered their documentation to be the standard, not the schema. and their api reflected the documentation. if things have changed, i would go with that. but i don't see how `alternateIdentifier` replacing `identifier` seems like a reasonable change. 

see the schema here: https://datacite-metadata-schema.readthedocs.io/en/4.5/properties/identifier/ 
`identifer` is mandatory
`alternateidentifier` is optional

they used to consider the schema and the XSD file as what they support. anything json schema was secondary. if they have changed that, go ahead. but i would check the XSD. that's what we did and it did ensure API success.

yarikoptic:
> `identifer` is mandatory

`identifier` is singular and thus `Occurrences: 1`. 
We are talking about `identifiers` (plural) for which there is only indeed that https://datacite-metadata-schema.readthedocs.io/en/4.5/properties/alternateidentifier/ which allows for multiple `Occurrences: 0-n`. 

XSD 4.5
- there is a singular resource / identifier:  https://github.com/datacite/schema/blob/HEAD/source/meta/kernel-4.5/metadata.xsd#L39
- there is the plural alternateIdentifiers https://github.com/datacite/schema/blob/HEAD/source/meta/kernel-4.5/metadata.xsd#L219 seemingly defined at the same level
image

In json schema 
- discussion on dropping "identifiers" in favor of "alternateIdentifiers": https://github.com/inveniosoftware/datacite/issues/81#issuecomment-1921727400 
- seems have no "identifier" whatsoever: https://github.com/inveniosoftware/datacite/issues/102 and anyways the way we used it (as the diff in this PR shows) it was not a singular `"identifier"` but for the `"identifiers"` records of "identifier".  
  - top level "identifier" here is something datacite assigns for the record.

<details>
<summary>Here is the XML of the datacite record we get for a random sample dandiset we have published</summary> 

```shell
❯ curl -LH "Accept: application/vnd.datacite.datacite+xml" https://doi.org/10.48324/DANDI.000897/0.240605.1710 
```

```xml
<?xml version="1.0" encoding="UTF-8"?>
<resource xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://datacite.org/schema/kernel-4" xsi:schemaLocation="http://datacite.org/schema/kernel-4 http://schema.datacite.org/meta/kernel-4/metadata.xsd">
  <identifier identifierType="DOI">10.48324/DANDI.000897/0.240605.1710</identifier>
  <creators>
    <creator>
      <creatorName nameType="Personal">Neupane, Sujaya</creatorName>
      <givenName>Sujaya</givenName>
      <familyName>Neupane</familyName>
      <nameIdentifier nameIdentifierScheme="ORCID" schemeURI="https://orcid.org/">0000-0002-0052-3122</nameIdentifier>
    </creator>
    <creator>
      <creatorName nameType="Personal">Fiete, Ila</creatorName>
      <givenName>Ila</givenName>
      <familyName>Fiete</familyName>
      <nameIdentifier nameIdentifierScheme="ORCID" schemeURI="https://orcid.org/">0000-0003-4738-2539</nameIdentifier>
    </creator>
    <creator>
      <creatorName nameType="Personal">Jazayeri, Mehrdad</creatorName>
      <givenName>Mehrdad</givenName>
      <familyName>Jazayeri</familyName>
      <nameIdentifier nameIdentifierScheme="ORCID" schemeURI="https://orcid.org/">0000-0002-9764-6961</nameIdentifier>
    </creator>
  </creators>
  <titles>
    <title>Neupane_Fiete_Jazayeri_Mental navigation_NHP_EntorhinalCortex</title>
  </titles>
  <publisher>DANDI Archive</publisher>
  <publicationYear>2024</publicationYear>
  <resourceType resourceTypeGeneral="Dataset">Neural Data</resourceType>
  <subjects>
    <subject>entorhinal cortex, cognitive map, mental navigation,</subject>
  </subjects>
  <contributors>
    <contributor contributorType="ContactPerson">
      <contributorName nameType="Personal">Neupane, Sujaya</contributorName>
      <givenName>Sujaya</givenName>
      <familyName>Neupane</familyName>
      <nameIdentifier nameIdentifierScheme="ORCID" schemeURI="https://orcid.org/">0000-0002-0052-3122</nameIdentifier>
    </contributor>
    <contributor contributorType="ContactPerson">
      <contributorName nameType="Personal">Jazayeri, Mehrdad</contributorName>
      <givenName>Mehrdad</givenName>
      <familyName>Jazayeri</familyName>
      <nameIdentifier nameIdentifierScheme="ORCID" schemeURI="https://orcid.org/">0000-0002-9764-6961</nameIdentifier>
    </contributor>
  </contributors>
  <alternateIdentifiers>
    <alternateIdentifier alternateIdentifierType="URL">https://identifiers.org/DANDI:000897/0.240605.1710</alternateIdentifier>
    <alternateIdentifier alternateIdentifierType="URL">https://dandiarchive.org/dandiset/000897/0.240605.1710</alternateIdentifier>
  </alternateIdentifiers>
  <sizes/>
  <formats/>
  <version/>
  <rightsList>
    <rights rightsIdentifier="cc_by_40" rightsIdentifierScheme="SPDX"/>
  </rightsList>
  <descriptions>
    <description descriptionType="Abstract">The dataset contains electrophysiology data recorded from the entorhinal cortex of two NHPs performing a mental navigation task. The recording probes used were V-probe with 32 channels or 64 channels, manufactured by Plexon Inc. </description>
  </descriptions>
  <fundingReferences>
    <fundingReference>
      <funderName>National Institute of Mental Health</funderName>
      <funderIdentifier funderIdentifierType="ROR">https://ror.org/05xj56w78</funderIdentifier>
      <awardNumber>NIMH-MH129046</awardNumber>
    </fundingReference>
    <fundingReference>
      <funderName>Natural Science and Engineering Council of Canada</funderName>
      <funderIdentifier funderIdentifierType="ROR">https://ror.org/01h531d29</funderIdentifier>
      <awardNumber>NSERC PDF-516867-2018</awardNumber>
    </fundingReference>
  </fundingReferences>
</resource>

```
</details> 

so you can see that
- "identifier" is the DOI assigned by datacite
- "alternateIdentifiers" is what we were providing via `identifiers` list of "identifier" records

AFAIK we are testing submission of our datacite records against test fabric, so AFAIK switching to alternateIdentifiers here should work just fine.

satra:
thank you @yarikoptic for the explanation. i'm now aligned.

dandischema/datacite/__init__.py
```diff
@@ -98,7 +101,13 @@ def to_datacite(
     attributes["descriptions"] = [
         {"description": meta.description, "descriptionType": "Abstract"}
     ]
-    attributes["publisher"] = "DANDI Archive"
+    attributes["publisher"] = {
+        "name": "DANDI Archive",
+        "schemeUri": "https://scicrunch.org/resolver/",
+        "publisherIdentifier": "https://scicrunch.org/resolver/RRID:SCR_017571",
+        "publisherIdentifierScheme": "RRID",
```
tmorrell, line 108: RRID technically isn't supported yet, though it should be very soon (when 4.6 is released). It doesn't fail out on the jsonschema because of a bug https://github.com/inveniosoftware/datacite/issues/103

It may work right now because DataCite is rolling out 4.6 support, but is worth double checking it goes through to Fabrica.

tmorrell:
This looks like the right changes to me. From the inveniosoftware/datacite side, we test that all the example json records make it to Fabrica. Around 4.3-4.4 there were issues with parts of the json representations not making it through (it's why we never did a 4.4 jsonschema), but I think all those issues other then GeoLocationPolygon have been resolved. If you find anything weird though just open an issue and we'll fix it. The goal is that anything that validates with the jsonschema goes through to Fabrica.

dandischema/datacite/__init__.py
```diff
@@ -78,16 +78,19 @@ def to_datacite(
...
         attributes["event"] = "publish"
 
-    attributes["identifiers"] = [
+    attributes["alternateIdentifiers"] = [
         # TODO: the first element is ignored, not sure how to fix it...
-        {"identifier": f"https://doi.org/{meta.doi}", "identifierType": "DOI"},
         {
-            "identifier": f"https://identifiers.org/{meta.id}",
-            "identifierType": "URL",
+            "alternateIdentifier": f"https://doi.org/{meta.doi}",
```
tmorrell, line 84: Is this a DataCite DOI that you're minting? If so, it should go in the `doi` field. But it kind of depends on what your workflow is.

yarikoptic: ;-)
Makes sense!  Will remove now!

dandischema/datacite/__init__.py
yarikoptic: We checked, nothing in 4.6 PR seems to suggest for it to become a controlled dictionary, just a "string" ATM, so  I  think we should keep it.

FWIW -- I did open and close
- https://github.com/datacite/schema/issues/150

I guess we could have switched to https://www.re3data.org/repository/r3d100013638  which also points to the RRID.  But it also shows that may be here it could be `publisherIdentifiers` list of records? (I don't want even to suggest that "officially" ;-) )

tmorrell: Ah, you're right. I got messed up by the examples and that RRID is going to be supported as a related identifier type in 4.6. I was just on the ROR call today, where they showed the data in this field is a mess. But that's that happens without a controlled list.

yarikoptic: I think there is nothing more todo here, marking resolved, please reopen/comment if I was wrong

yarikoptic:
windows fails are unrelated (reported to codecov elsewhere).  @satra I am taking your "aligned" [above](https://github.com/dandi/dandi-schema/pull/261#issuecomment-2484498929) as a blessing.

As we are already on 0.10.4-24-g9812ced - I will mark for release.  I will merge in a day or two if there is no objections. In principle we could have updated all DOIs but I would not actively persuade it here

PR #262, opened 2024-11-04
Drop Python 3.8 support (remove typing_extensions from depends)

yarikoptic:


.github/workflows/test.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           - ubuntu-latest
           - macos-12
         python:
-          - 3.8
+          - 3.9
```
candleindark, line 23:
```suggestion
          - 3.9
```
We can just remove the `- 3.8` here since the 3.9 options is already available.

yarikoptic: ooops -- missed that!  we just need to remove this line since followed by 3.9... will do, thanks

yarikoptic: rright, removal suggestion would be 
```suggestion
```

yarikoptic:
whatever had to run seems to have ran and is green!  there is some odd memory of 3.8 and macos-12 which are no more:

```
❯ git grep -e macos-12 -e '3\.8' | grep -v CHANG
❯ 
```
Let's proceed so other PRs could get greener too

PR #271, opened 2024-12-10
Use `title` in place of `examples`

### LINKED ISSUE #270: Avoid using `examples` in `Person` model
The use of the `examples` property in the `Person` model definition

https://github.com/dandi/dandi-schema/blob/9812ced7ef6b74f07654b34242abac337e186ebe/dandischema/models.py#L948-L952

causes VJSF to display the examples in a dropdown-type menu for the name field (as shown in, e.g., https://github.com/dandi/dandi-archive/issues/2090#issue-2720677604) which then causes interaction difficulty as further described in that issue report.

The examples seem to be unnecessary, if their goal is to show the user that names must appear in `<family name>, <given name>` format. Instead, the `title` property of the model could include this information; VJSF renders this property as a placeholder text within an empty field, which would make for an arguably better prompt for the user to enter the name in the format. It would also eliminate the poor UX resulting from the examples dropdown, and simplify the form as well.

waxlamp:
Closes #270; addresses https://github.com/dandi/dandi-archive/issues/2090 indirectly.

The `title` property is rendered by VJSF as dimmed placeholder text inside the field itself, which is a subtle but clear hint to the user to format the name correctly. This is simpler and less obtrusive than using `examples`, which causes VJSF to instead render a dropdown menu containing the examples. This prevents editing of an existing name and implies that the example names are the only suggested names in the system.


dandischema/models.py
```diff
@@ -946,10 +946,10 @@ class Person(Contributor):
         json_schema_extra={"nskey": "schema"},
     )
     name: str = Field(
+        title="Lovelace, Augusta Ada",
```
satra, line 949:
```suggestion
        title="Use Last, First. Example: Lovelace, Augusta Ada",
```

dandischema/consts.py
```diff
@@ -11,6 +11,7 @@
     "0.6.5",
```
candleindark, line 11: I think `DANDI_SCHEMA_VERSION = "0.6.8"` in line 1 should change to `DANDI_SCHEMA_VERSION = "0.6.9" as well since we are publishing a new version of the schema.

candleindark:
I compared the generated JSON schemas with the 0.6.8 version. All differences are expected. Since both the `title` and `examples` attributes in JSON schema are not used for validation. It's unlikely that this change will break anything.

PR #275, opened 2025-01-19
[WIP] updating life cycle of dandiset

djarecka:
- creating test for public dandiset that simulate the life cycle of dandiset and doi updates. Not sure if we should always run the test, so for now I added skip; 
- updates to `to_datacite`  function, so it can also work for `draft`

### DIFF
--- a/dandischema/datacite/tests/test_datacite.py
+++ b/dandischema/datacite/tests/test_datacite.py
@@ -1,17 +1,17 @@
+from copy import deepcopy
 import json
 import os
 from pathlib import Path
 import random
 from typing import Any, Dict, Tuple
-from copy import deepcopy
 
 from jsonschema import Draft7Validator
 import pytest
 import requests
 
 from dandischema.models import (
-    LicenseType,
     Dandiset,
+    LicenseType,
     PublishedDandiset,
     RelationType,
     ResourceType,
@@ -62,7 +62,6 @@ def datacite_update(datacite: dict, doi: str) -> None:
     rg.raise_for_status()
 
 
-
 def _clean_doi(doi: str) -> None:
     """Remove doi. Status code is ignored"""
     rq = requests.delete(
@@ -72,7 +71,10 @@ def _clean_doi(doi: str) -> None:
     print("\n in _clean_doi", doi, rq.status_code)
     return rq.status_code
 
-@pytest.mark.skip(reason="to not produced too many dois, not sure if we want to keep it as a test")
+
+@pytest.mark.skip(
+    reason="to not produced too many dois, not sure if we want to keep it as a test"
+)
 def test_datacite_lifecycle() -> None:
     """testing the lifecycle of a public dandiset and doi (from draft to published)"""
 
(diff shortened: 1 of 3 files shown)


dandischema/datacite/tests/test_datacite.py
```diff
@@ -35,21 +37,178 @@ def datacite_post(datacite: dict, doi: str) -> None:
...
+    rq = requests.delete(
         f"https://api.test.datacite.org/dois/{doi}",
         auth=("DARTLIB.DANDI", os.environ["DATACITE_DEV_PASSWORD"]),
     )
+    print("\n in _clean_doi", doi, rq.status_code)
+    return rq.status_code
+
+
+@pytest.mark.skip(
+    reason="to not produced too many dois, not sure if we want to keep it as a test"
```
yarikoptic, line 76: that is because the issue with DOI deletion was not yet addressed on datacite side, right?
then better be mentioned. in the current skip. 

Otherwise -- test should be diligent about removing any DOI it created in its life time.

dandischema/datacite/tests/test_datacite.py
```diff
@@ -35,21 +37,178 @@ def datacite_post(datacite: dict, doi: str) -> None:
...
+        "license": [LicenseType("spdx:CC-BY-4.0")],
+        "citation": "A_last, A_first 2021",
+        "manifestLocation": [
+            f"https://api.dandiarchive.org/api/dandisets/{dandi_id}/versions/{version}/assets/"
+        ],
+        "assetsSummary": {
+            "schemaKey": "AssetsSummary",
+            "numberOfBytes": 10,
+            "numberOfFiles": 1,
+        },
```
yarikoptic, line 121: hopefully does not matter but whenever dandiset is created, there is no assets summary yet -- it is done async IIRC, hence

```suggestion
```

yarikoptic (changes requested):
Thank you!

I left minor comments/wishes

dandischema/datacite/tests/test_datacite.py
djarecka: no, the bug on the datacite website is for `draft` doi.

The test creates a `Findable` doi, and this can't be deleted

dandischema/datacite/tests/test_datacite.py
djarecka: `assetsSummary` is a mandatory field for `Dandiset`

dandischema/datacite/tests/test_datacite.py
yarikoptic: re DOI deletion issue , now we filed
- https://github.com/datacite/bracco/issues/925
- https://github.com/datacite/bracco/issues/924

and indeed, let's not pollute the DOI space there for now (as with regular automated runs).  We might add option to trigger execution once in a while though (monthly?)

dandischema/datacite/tests/test_datacite.py
yarikoptic: in dandi archive we do following if not provided in metadata

```
        if 'assetsSummary' not in metadata:
            metadata['assetsSummary'] = {
                'schemaKey': 'AssetsSummary',
                'numberOfBytes': 0,
                'numberOfFiles': 0,
            }
```

http://github.com/dandi/dandi-archive/blob/HEAD/dandiapi/api/models/version.py

PR #278, opened 2025-01-29
Improve (better arguments validation, avoiding repeated creation of validator objects, etc) `_validate_obj_json()` in `metadata.py` and supporting funcs

candleindark:
This PR make the following improvement relating to validating metadata instances against a DANDI model's JSON schema.

1. Redefine the interface of `_validate_obj_json()` so that it take a jsonschema validator instead of a JSON schema as argument to avoid recreation of validators for the same schema.
2. Define a utility function, `jsonschema_validator(), for creating an appropriate JSON schema validator given a schema.
    1. Ensure validity of the schema is [checked](https://python-jsonschema.readthedocs.io/en/stable/validate/#the-validator-protocol) to prevent undefined behavior. (A bug fix)
    2. Utilized the `"$schema"` property in a provide schema to determine the appropriate validator class
2. Define a utility function, `validate_json()`, to validate a data instance in general, without the `missing_ok` parameter.
3. Define a utility function, `dandi_jsonschema_validator`, that decides the default/fallback validator class to used based on the `"schemaVersion"` property of the given schema instead of the `"schemaVersion"` field of a given instance.  (A bug fix)
4. Define utility functions, `_get_jsonschema_validator()`and `_get_jsonschema_validator_local()`, with cache to avoid repeated creation of jsonschema validators for the same schemas. **Note**: `functools.cache()` instead of `functools.lru_cache()` is used, for improved performance, in these functions because valid inputs to these functions have been limited by validation, so there is no worry about the cache growing without bound.
5. Reimplement `_validate_obj_json()` using the aforementioned utility functions
6. Rename `schema_map` in `metadata.py` to `SCHEMA_MAP` for the reference dictionary is a constant.
7. Add extensive tests to functions covered in this PR.

### DIFF
--- a/dandischema/metadata.py
+++ b/dandischema/metadata.py
@@ -1,11 +1,12 @@
 from copy import deepcopy
 from enum import Enum
-from functools import lru_cache
+from functools import cache
 from inspect import isclass
 import json
 from pathlib import Path
 from typing import Any, Dict, Iterable, Optional, TypeVar, Union, cast, get_args
 
+from jsonschema.protocols import Validator as JsonschemaValidator
 import pydantic
 import requests
 
@@ -20,13 +21,16 @@ from . import models
 from .utils import (
     TransitionalGenerateJsonSchema,
     _ensure_newline,
+    dandi_jsonschema_validator,
+    json_object_adapter,
     sanitize_value,
     strip_top_level_optional,
     validate_json,
     version2tuple,
 )
 
-schema_map = {
+# A mapping of the schema keys of DANDI models to the names of their JSON schema files
+SCHEMA_MAP = {
     "Dandiset": "dandiset.json",
     "PublishedDandiset": "published-dandiset.json",
     "Asset": "asset.json",
@@ -130,7 +134,7 @@ def publish_model_schemata(releasedir: Union[str, Path]) -> Path:
     version = models.get_schema_version()
     vdir = Path(releasedir, version)
     vdir.mkdir(exist_ok=True, parents=True)
-    for class_, filename in schema_map.items():
+    for class_, filename in SCHEMA_MAP.items():
         (vdir / filename).write_text(
             _ensure_newline(
                 json.dumps(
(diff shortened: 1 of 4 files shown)


dandischema/utils.py
```diff
@@ -136,3 +141,91 @@ def sanitize_value(value: str, field: str = "non-extension", sub: str = "-") ->
...
+    """
+    # Retrieve appropriate validator class for validating the given schema
+    validator_cls: type[JsonschemaValidator] = (
+        validator_for(schema, default_cls)
+        if default_cls is not None
+        else validator_for(schema)
+    )
+
+    # Ensure the schema is valid
+    validator_cls.check_schema(schema)
```
yarikoptic, line 174: wouldn't that be called for the same schema over and over again for every instance?

dandischema/metadata.py
```diff
@@ -250,7 +258,7 @@ def validate(
                     "using json schema for older versions"
                 )
             schema = _get_schema(schema_version, schema_map[schema_key])
-        _validate_obj_json(obj, schema, missing_ok)
+        _validate_obj_json(obj, schema, missing_ok=missing_ok)
```
yarikoptic, line 261: let's RF further to avoid carrying the `schema` around , and rather pass `schema_validator`, and have


```python
@lru_cache
def _get_schema_validator(schema_version: str, schema_name: str) -> Any:
```

where we would do all necessary job of loading from the web, creating validator, validating the schema.

It would make code more performant (not re-creating the same validator over and over again), and would not add overhead of the validating the schema itself.

candleindark: Done. [`_get_jsonschema_validator()`](https://github.com/dandi/dandi-schema/blob/e32f32d5619860b5daa27eb866ec60e8f69b8e55/dandischema/metadata.py#L198) and [`_get_jsonschema_validator_local()`](https://github.com/dandi/dandi-schema/blob/e32f32d5619860b5daa27eb866ec60e8f69b8e55/dandischema/metadata.py#L251) are defined (docstrings for info). **Note**: functools.cache() instead of functools.lru_cache() is used, for improved performance, in these functions because valid inputs to these functions have been limited by validation, so there is no worry about the cache growing without bound.

candleindark:
@yarikoptic I think the code is good now. However, I did quite a bit of validation of function args. I could have used `@validate_call` from Pydantic to remove some code duplications and boilerplates. Let me know if you want me to do that as well, but I think I spent enough time on this PR already.

PR #282, opened 2025-02-05
Use `[]` as default affiliation value

waxlamp:
This provides `[]` as a default value for `Person.affiliation`, which is typed as a list of `Affiliation`. The current default value of `null` causes problems when trying to process an empty affiliation value in the DANDI Archive meditor (see https://github.com/dandi/dandi-archive/issues/2150).

### DIFF
--- a/dandischema/consts.py
+++ b/dandischema/consts.py
@@ -1,4 +1,4 @@
-DANDI_SCHEMA_VERSION = "0.6.9"
+DANDI_SCHEMA_VERSION = "0.6.10"
 ALLOWED_INPUT_SCHEMAS = [
     "0.4.4",
     "0.5.1",
@@ -12,6 +12,7 @@ ALLOWED_INPUT_SCHEMAS = [
     "0.6.6",
     "0.6.7",
     "0.6.8",
+    "0.6.9",
     DANDI_SCHEMA_VERSION,
 ]
 
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -952,7 +952,7 @@ class Person(Contributor):
         json_schema_extra={"nskey": "schema"},
     )
     affiliation: Optional[List[Affiliation]] = Field(
-        None,
+        default=[],
         description="An organization that this person is affiliated with.",
         json_schema_extra={"nskey": "schema"},
     )


dandischema/consts.py
```diff
@@ -1,4 +1,4 @@
-DANDI_SCHEMA_VERSION = "0.6.9"
+DANDI_SCHEMA_VERSION = "0.6.10"
```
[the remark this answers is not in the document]
candleindark: I think it is most likely to be `0.6.10`. @yarikoptic ?

dandischema/models.py
```diff
@@ -952,7 +952,7 @@ class Person(Contributor):
         json_schema_extra={"nskey": "schema"},
     )
     affiliation: Optional[List[Affiliation]] = Field(
-        None,
+        default=[],
```
candleindark, line 955: I am OK with the proposed change. I will be a bit inconsistent with many definition though, such as https://github.com/dandi/dandi-schema/blob/c3007768e002c9f51ea37b5e6b3628f7f7f20943/dandischema/models.py#L1522-L1526.

candleindark (approved):

dandischema/models.py
waxlamp: I think, to the extent possible, those other definitions should be updated to also use `[]`. This is because (1) we already ran into a UX bug due to the affiliation problem I'm addressing in this PR and (2) the [jsonschema spec](https://json-schema.org/draft/2020-12/json-schema-validation#name-default) itself recommends that default values be valid according to the schema. I asked ChatGPT about reasons not to have a valid default value and it provided some [examples](https://chatgpt.com/share/67a3dce7-ec40-800d-985f-08ac6beabaac) (none of which seem to apply to us).

candleindark: I agree with you on that if we only consider the Pydantic models.

In fact, I was thinking changing `affiliation: Optional[List[Affiliation]]` to just `affiliation: List[Affiliation]`, building on your proposed change, since we don't need both `None` and `[]` to indicate having no affiliation. However, doing so has at least two downsides.

1. We will have to migrate the existing metadata
2. There is no default value of an empty list in LinkML. We will be able to get a more precise translation of the models to LinkML if we keep the current definition.

waxlamp: I don't know how to solve the LinkML issue, but I agree with your point about changing the type to `List[Affiliation]`. It's true that it would require a data migration, but in my experience maintaining even slightly wrong types in systems like this tends to have compounding effects into the future (in our case, one of those effects would be the crash I'm trying to fix in the meditor). If we can figure out our approach to automatic migration of metadata, that may become less painful.

But, for now, you're thinking to not make that change in order to keep the blast radius from this PR smaller, is that right?

waxlamp:
@candleindark, I see that you approved this PR (thank you for the quick turnaround) but we still have an open question about the schema version, and there's a failing coverage test. Is there anything I should do to help advance this? Will you merge this when ready, or is that up to me?

candleindark:
@waxlamp Everything looks fine for me. "All modified and coverable lines are covered by tests." I don't do merge on this repo. @yarikoptic and @satra are the ones with the final call on this repo.

dandischema/models.py
candleindark: Yes, I agree with you but let's keep it small for now. If it is not be cause of the LinkML translation, I would suggest changing to all similar type declaration to something like `List[Affiliation]` already.

waxlamp:
In light of [this discussion](https://github.com/dandi/dandi-archive/issues/2150#issuecomment-2644188856), I'll close this PR. We can revisit this later if needed.

PR #293, opened 2025-03-13
Add support for conversion to Google dataset search metadata

satra:
This adds a function to convert a dandiset metadata schema to Google dataset search form. 

However to embed the metadata on a webpage a few additional steps need to happen:

- expand the metadata using jsonld expand
- compact the metadata using a new context: `{"@context": "http://schema.org/"}`

this would either require adding pyld (which works but hasn't seen an update in a while), or doing it in javascript.

### DIFF
--- a/dandischema/tests/test_google_dataset_metadata.py
+++ b/dandischema/tests/test_google_dataset_metadata.py
@@ -1,0 +1,194 @@
+import copy
+from typing import Dict, Any
+
+import pytest
+
+from dandischema.utils import google_dataset_metadata
+
+
+@pytest.fixture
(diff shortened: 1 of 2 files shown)


satra:
i'm going to close this in favor of simply using datacite for now. i'll send a PR to dandiarchive for that. 

`curl -LH "Accept: application/ld+json" https://doi.org/10.48324/dandi.000617/0.250312.0130`

what they generate doesn't pass the validator fully, but for our purposes it's a better route at the moment.

satra:
PR submitted: https://github.com/dandi/dandi-archive/pull/2212

PR #294, opened 2025-04-04
Devendorize: Replace DANDI specific identifier with more generic regexes

yarikoptic:
to accomodate LINC and EMBER. We are yet to workout proper vendorization where a particular model, ideally not entire run time as we could do via env var, could be parametrized with a specific vendor information. I think that relaxing those regular expressions altogether would help to adopt dandi for custom deployments meanwhile.

@satra -- WDYT for a quick workaround like this? any side effects I am not foreseeing besides "looser" validation?

Also attn @aaronkanzer and @kabilar -- for LINC did you just keep DANDI identifiers?

TODOs:
- [ ] actually may be I should add vendoring via a few env vars, then at least a specific deployment of dandi-archive could restrict more heavily for a specific instance
- [ ] add test cases to ensure that works for alternative identifiers


satra (commented):
while this would work (there are several tests failing), how about get them from a config to be specific? also take a look at the jsonld context generator to see which ones map on specifically to dandiarchive. those would also need to be devendored.

yarikoptic:
> how about get them from a config to be specific?

what config do you have in mind since so far we have none -- typically we just resort to some env vars (that's what I was thinking we should do in the short run).  The point is that it should be "instance" (not runtime) specific which is ATM not supported at all, as e.g. imagine within the same session uploading to both EMBER and DANDI, so then having either validation or ideally entire model per each such different instance type.

satra:
perhaps we are saying the same thing in slightly different ways. dandischema can be devendored and instance agnostic (perhaps and hopefully). anything using it (server, cli, etc.,.) can make an instance of it. i.e. dandiset/asset for an instance of dandi with the right config. so introducing an instance config that can be/required to be passed on to initializing a model. 

for example, this could also allow EMBER to have different license considerations than DANDI. 

one possibility is perhaps to still consider a generic core model (e.g., union of licenses, more generic doi/devendored setup like this PR) and then an instance specific validation/schema/constraint/version layer that can be customized. 

for this PR, let's just simplify(expand) for now. i think some of the context adjustment will also need to happen (here: https://github.com/dandi/dandi-schema/blob/782c421e60b5989d3586d9cf89c526b1ac0a6778/dandischema/metadata.py#L46)

```json
        "dandi": "http://schema.dandiarchive.org/",
        "dcite": "http://schema.dandiarchive.org/datacite/",
        "dandiasset": "http://dandiarchive.org/asset/",
        "DANDI": "http://dandiarchive.org/dandiset/",
```

to this (note: the schema ones **should not** change for now at least as the are simple the model)

```json
        "dandi": "http://schema.dandiarchive.org/",
        "dcite": "http://schema.dandiarchive.org/datacite/",
        "dandiasset": "http://dandiarchive.org/asset/",
        "DANDI": "http://dandiarchive.org/dandiset/",
        "emberasset": "http://emberarchive.org/asset/",
        "EMBER": "http://emberarchive.org/dandiset/",
```

a full refactor to dandi + instance concepts would require us to make some breaking changes and different identifiers. we could may be craft such a thing this week when we are together.

aaronkanzer:
@yarikoptic -- for LINC, we are currently not issuing DOIs (not using DANDI either)

I have had placeholders in for the Django env vars expected -- I can certainly implement wherever EMBER goes for their own logic for LINC though (let me know if this makes sense @kabilar )

cc @satra

kabilar:
> @yarikoptic -- for LINC, we are currently not issuing DOIs (not using DANDI either)

Thanks @yarikoptic.  I agree with Aaron.  For LINC, we don't use DOIs or `DANDI:` identifiers, just URLs.  And we don't currently have a need for either.

dandischema/models.py
```diff
@@ -58,8 +58,19 @@
 )
 ASSET_UUID_PATTERN = r"^dandiasset:" + UUID_PATTERN
 VERSION_PATTERN = r"\d{6}/\d+\.\d+\.\d+"
-DANDI_DOI_PATTERN = rf"^10.(48324|80507)/dandi\.{VERSION_PATTERN}"
-DANDI_PUBID_PATTERN = rf"^DANDI:{VERSION_PATTERN}"
+# Vendored
+# DANDI_DOI_PATTERN = rf"^10.(48324|80507)/dandi\.{VERSION_PATTERN}"
+# Unvendored, only with 10. prefix, as likely all would have for datacite
+DANDI_DOI_PATTERN = rf"^10.\d+/[a-z]\.{VERSION_PATTERN}"
```
candleindark, line 64:
```suggestion
DANDI_DOI_PATTERN = rf"^10.\d+/[a-z]+\.{VERSION_PATTERN}"
```

candleindark:
I think there is away to take a config and vendorize the models and enums accordingly even with the technologies we are currently using. Pydantic has a [dynamic model creation syntax](https://docs.pydantic.dev/latest/concepts/models/#dynamic-model-creation) and `Enum` has [a functional API](https://docs.python.org/3.9/library/enum.html#functional-api). However, using these syntaxes/APIs to define all the models and enums (mostly likely within functions) are going to be messy.

dandischema/models.py
```diff
@@ -58,8 +58,19 @@
...
+# Vendored:
+# DANDI_PUBID_PATTERN = rf"^DANDI:{VERSION_PATTERN}"
+# Unvendored:
+DANDI_PUBID_PATTERN = rf"^[A-Z]+:{VERSION_PATTERN}"
+# Vendored:
+#  potentially could be set to alternative namespace
+# Unvendored: use "dandi" by default
+DANDI_NSKEY = "dandi"
```
candleindark, line 72: I don't think I get this one completely. How are you going to vendorize it if this is hardcoded? Is the user supposed to alter this line for different instances of dandi?

yarikoptic: I was thinking about just ad-hoc env variables handling here, but @candleindark reminded about https://docs.pydantic.dev/latest/concepts/pydantic_settings/ which  he also used in datalad-registry, so may be we should go for it here... 
benefits:
- validation
- flexibility in import (could be entire config in json)

then, to minimize changes we would just assign constants here.  It is a separate package https://pypi.org/project/pydantic-settings/

dandischema/models.py
```diff
@@ -58,8 +58,19 @@
 )
 ASSET_UUID_PATTERN = r"^dandiasset:" + UUID_PATTERN
 VERSION_PATTERN = r"\d{6}/\d+\.\d+\.\d+"
-DANDI_DOI_PATTERN = rf"^10.(48324|80507)/dandi\.{VERSION_PATTERN}"
-DANDI_PUBID_PATTERN = rf"^DANDI:{VERSION_PATTERN}"
```
yarikoptic, line 62: TODO: below in the code we also have `DANDI:`, should probably be moved here into a dedicated variable...

dandischema/models.py
```diff
@@ -1605,14 +1624,20 @@ def contributor_musthave_contact(
...
     )
 
     identifier: DANDI = Field(
         title="Dandiset identifier",
         description="A Dandiset identifier that can be resolved by identifiers.org.",
-        pattern=r"^DANDI:\d{6}$",
+        # Vendored:
+        # pattern=r"^DANDI:\d{6}$",
+        # Unvendored:
+        pattern=r"^[A-Z]+:\d{6}$",
```
yarikoptic, line 1640: TODO: left a comment on top where we define all of such things.  We might need to see if "IDENTIFIERS_PREFIX" is defined, and if so -- use it as is, and if not defined -- make it into such unvendored regex.

dandischema/models.py
```diff
@@ -58,8 +64,12 @@
 )
 ASSET_UUID_PATTERN = r"^dandiasset:" + UUID_PATTERN
 VERSION_PATTERN = r"\d{6}/\d+\.\d+\.\d+"
-DANDI_DOI_PATTERN = rf"^10.(48324|80507)/dandi\.{VERSION_PATTERN}"
-DANDI_PUBID_PATTERN = rf"^DANDI:{VERSION_PATTERN}"
+# Vendored
+# DANDI_DOI_PATTERN = rf"^10\.(48324|80507)/dandi\.{VERSION_PATTERN}"
+# Unvendored, only with 10. prefix, as likely all would have for datacite
+DANDI_DOI_PATTERN = rf"^10\.\d+/[a-z]+\.{VERSION_PATTERN}"
```
candleindark, line 70: @yarikoptic I think this PR is complete except you may want to go further with this line by leveraging the configuration in `conf.py` like what is done for `DANDI_PUBID_PATTERN`. There are two parts in this pattern you may want get through the config (I don't know how appropriately to name them). The `\d+` (or `(48324|80507)` in the vendored version) and the `[a-z]+` (or `dandi` in the vendor version).

candleindark:
@yarikoptic I left a final comment at https://github.com/dandi/dandi-schema/pull/294#discussion_r2069432537. Otherwise, this PR is good to go. Feel free to squash the commits. There are small and serving as breadcrumbs for review purposes.

dandischema/models.py
candleindark: done

candleindark:
The two remaining tests failures must be unrelated to the current PR. The last time the two test succeeded was three months ago at https://github.com/dandi/dandi-schema/actions/runs/13537569388. They are worthy of further investigation though.

dandischema/models.py
yarikoptic: so we are limited to datacite ATM. I guess (didn't check), all datacite DOIs will start with `10.` (TODO: check)

so I feel like we need some `datacite_doi_id` which e.g. would be that 48324 for one of the dandi instances, and then compose it together with `f"{id_prefix_pattern.lower()}"`,  so then  here indeed it would become similar to `DANDI_PUBID_PATTERN`:

```suggestion
DANDI_DOI_PATTERN = rf"^10\.{DATACITE_DOI_ID}/{ID_PREFIX_PATTERN.lower()}\.{VERSION_PATTERN}"
```

dandischema/conf.py
```diff
@@ -0,0 +1,28 @@
...
+        same name prefixed with the prefix defined in `model_config` with the name
+        of the environment variable interpreted case-insensitively.
+        For details, see https://docs.pydantic.dev/latest/concepts/pydantic_settings/
+    """
+
+    model_config = SettingsConfigDict(env_prefix="dandi_schema_")
+
+    nskey: str = "dandi"
+    id_prefix_pattern: str = r"[A-Z]+"
+    """Regex pattern for the prefix of identifiers"""
```
yarikoptic, line 25: please accompany this PR with configs for 

- dandi
- dandi staging (AKA sandbox)
- ember  (for EMBER -- talk to Eric & Sandy)

for `dandi` config, I think you could provide it as one of the CI matrix runs  (or may be we could do for all)

dandischema/conf.py
```diff
@@ -0,0 +1,28 @@
...
+        instance of this class can be populated from an environment variable of the
+        same name prefixed with the prefix defined in `model_config` with the name
+        of the environment variable interpreted case-insensitively.
+        For details, see https://docs.pydantic.dev/latest/concepts/pydantic_settings/
+    """
+
+    model_config = SettingsConfigDict(env_prefix="dandi_schema_")
+
+    nskey: str = "dandi"
+    id_prefix_pattern: str = r"[A-Z]+"
```
yarikoptic, line 24: only 1 to add: for `DATACITE_DOI_ID` (see below) and I think we would have minimal set needed. I might even remove `nskey` from being configurable here (although in the code still use nskey variable).

yarikoptic: and let's shorten to `id_pattern`

dandischema/models.py
candleindark: > so we are limited to datacite ATM. I guess (didn't check), all datacite DOIs will start with 10. (TODO: check)

It turns out that the `10.` start is not unique to datacite DOIs. It's the beginning for all DOIs. (See https://support.datacite.org/docs/doi-basics and https://en.wikipedia.org/wiki/Digital_object_identifier).

The number sequence that follows `10.` is called "registrant code" in DOI terminology, so I will use the variable name `DOI_REGISTRANT_CODE` instead of `DATACITE_DOI_ID`.

yarikoptic: thanks! I think there is no need for `DOI_REGISTRANT_CODE` since as you say all DOIs would have it, and thus all datacite DOIs, and since we support only datacite, no need for anything else . Just add `DATACITE_DOI_ID_PATTERN` or alike, since we will use it as pattern right?

dandischema/conf.py
candleindark: done. `Config.id_prefix_pattern` is renamed to `Config.id_pattern, and ``Config.nskey` is drop since `"dandi"` is the namespace of DANDI ontology and doesn't change.

dandischema/models.py
candleindark: requested change is applied.

dandischema/conf.py
candleindark: I don't think I get this one exactly. With `config.Config`, dandi-schema now can be "vendorized", configured, for a particular vendor, such as dandi or ember. Shouldn't the most direct tests be the testing of validation of a vendorized `Dandiset` models against expected valid metadata instances from respective particular vendors? However, this route may requires a bit maintenance down the road since each time the model changes, we may have to update the metadata instances used in the tests.

yarikoptic:
Let's add testing for 3 scenarios in CI.  

1. "generic" - should handle records for any vendor
2. "dandi" - settings specific to DANDI
3. "ember" - settings specific to EMBER (different prefix - EMBER, and different DOI number, whatever it is)

records we have which are not vendor specific which we have in `dandischema/tests/data/metadata` should pass for all scenarios, and if they are specific to DANDI - only for generic and DANDI; and when we add some for EMBER - only for generic and EMBER.

For now let's just do

- what @candleindark says he would do
- merge this PR

and then we will reconsider...

dandischema/models.py
```diff
@@ -58,8 +64,12 @@
 )
 ASSET_UUID_PATTERN = r"^dandiasset:" + UUID_PATTERN
 VERSION_PATTERN = r"\d{6}/\d+\.\d+\.\d+"
-DANDI_DOI_PATTERN = rf"^10.(48324|80507)/dandi\.{VERSION_PATTERN}"
-DANDI_PUBID_PATTERN = rf"^DANDI:{VERSION_PATTERN}"
+DANDI_DOI_PATTERN = (
+    rf"^10\.{DATACITE_DOI_ID_PATTERN}/{ID_PATTERN.lower()}\.{VERSION_PATTERN}"
+)
+DANDI_PUBID_PATTERN = rf"^{ID_PATTERN}:{VERSION_PATTERN}"
+DANDI_NSKEY = "dandi"  # Namespace for DANDI ontology
```
asmacdo, line 71: Should be "dandi" no matter what? or would ID_PATTERN.lower() work here too?

candleindark: Are you talking about `DANDI_NSKEY`? If you are the answer is yes, according to @yarikoptic, the namespace for DANDI ontology is not going to change. `ID_PATTERN.lower()` can have a different value than `"dandi"`, so it will not work.

dandischema/conf.py
yarikoptic: here is the invocation I did
```
DANDI_SCHEMA_id_pattern=EMBER DANDI_SCHEMA_DATACITE_DOI_ID_PATTERN='(60533|82754)' python -m pytest -s -v dandischema/tests
```

which failed only a few tests... I pushed a commit to potentially fixup one or two tests

yarikoptic:
here you could grab a sample from EMBER  https://api-dandi.emberarchive.org/api/dandisets/000004/versions/draft/ but would still need to replace with EMBER: prefix since not vendorized version was used.

also think on how to do migration now within EMBER to adjust all those IDs etc...

dandischema/models.py
```diff
@@ -58,8 +64,12 @@
 )
 ASSET_UUID_PATTERN = r"^dandiasset:" + UUID_PATTERN
 VERSION_PATTERN = r"\d{6}/\d+\.\d+\.\d+"
-DANDI_DOI_PATTERN = rf"^10.(48324|80507)/dandi\.{VERSION_PATTERN}"
-DANDI_PUBID_PATTERN = rf"^DANDI:{VERSION_PATTERN}"
+DANDI_DOI_PATTERN = (
+    rf"^10\.{DATACITE_DOI_ID_PATTERN}/{ID_PATTERN.lower()}\.{VERSION_PATTERN}"
+)
```
yarikoptic, line 69: for @asmacdo and 
- https://github.com/dandi/dandi-archive/pull/2012
we might right away mandate either "empty" or "any" DOI, ie.

```suggestion
DANDI_DOI_PATTERN = (
    rf"^10\.{DATACITE_DOI_ID_PATTERN}/{ID_PATTERN.lower()}\.{VERSION_PATTERN}"
    if DATACITE_DOI_ID_PATTERN else "^$" # or ".*" for "any"
)
```

but we need to decide which one in that design PR.  Might be easier when implemented to see the manifestations

candleindark:
@yarikoptic The tests picked up a problem with the definition for the `DANDI_DOI_PATTERN`. https://github.com/dandi/dandi-schema/blob/a22076c930245b59b6784b6e48a5b06672801a9a/dandischema/models.py#L67-L69

This definition will not work with EMBER DOIs regardless of the values for the current configuration setup. `10.60533/ember-2025-g8j9` is an example of EMBER DOIs which points to https://dandi.emberarchive.org/dandiset/000005/0.250404.1839. The primary problem has to do with the definition of the DOI [suffix](https://support.datacite.org/docs/doi-basics#suffix).

I will define `DANDI_DOI_PATTERN` based on suffix pattern to be provided through configuration.

To be continue tomorrow.

P.S. I have saved the initiating commit you put in for testing. I will apply it when existing problems/conflicts are resolved.

yarikoptic:
re DOI pattern, I think EMBER folks could be convinced as you have initiated the dialog.  I think it would be best to stick to consistent DOI pattern across instances due to @asmacdo work on "Dandiset DOI", since otherwise we need to define two separate DOI patterns etc.

asmacdo:
I think we will need config values that are different between deployments or there will be DOI collisions. 

Even if the DOI prefix is different

10.12345/dandi.000001
10.11111/dandi.000001 might be confusing. IMO

10.12345/ember.000001 would be better

candleindark:
> I think we will need config values that are different between deployments or there will be DOI collisions.
> 
> Even if the DOI prefix is different
> 
> 10.12345/dandi.000001 10.11111/dandi.000001 might be confusing. IMO
> 
> 10.12345/ember.000001 would be better

That's a valid concern, but I think our current setup is doing that already. The `{ID_PATTERN.lower()}` in

https://github.com/dandi/dandi-schema/blob/a22076c930245b59b6784b6e48a5b06672801a9a/dandischema/models.py#L67-L69

provides the beginning of the suffix of the DOI. If the user launches the EMBER DANDI instance with `ID_PATTERN` set to `"EMBER"` and the DANDI DANDI instance with `ID_PATTERN` set to `"DANDI"`, they will achieve the desired result you mentioned.

asmacdo:
@candleindark I think I may have misinterpreted:

> re DOI pattern, I think EMBER folks could be convinced as you have initiated the dialog. I think it would be best to stick to consistent DOI pattern across instances due to @asmacdo work on "Dandiset DOI", since otherwise we need to define two separate DOI pattern config settings etc.

I thought this was suggesting not to use `{ID_PATTERN.lower}` and keep that consistent across deployments. If thats not true, I think we are all good!

candleindark:
> @candleindark I think I may have misinterpreted:
> 
> > re DOI pattern, I think EMBER folks could be convinced as you have initiated the dialog. I think it would be best to stick to consistent DOI pattern across instances due to @asmacdo work on "Dandiset DOI", since otherwise we need to define two separate DOI pattern config settings etc.
> 
> I thought this was suggesting not to use `{ID_PATTERN.lower}` and keep that consistent across deployments. If thats not true, I think we are all good!

My interpretation is to use it consistently across different instances though each instance may have a different value for `ID_PATTERN`.

yarikoptic:
@candleindark great to see more (vendorized) tests going, but I wonder if we should limit vendorized (`ember` and `dandi`) to just a few, like only smallest python version across all OSes.

candleindark:
> @candleindark great to see more (vendorized) tests going, but I wonder if we should limit vendorized (`ember` and `dandi`) to just a few, like only smallest python version across all OSes.

Great suggestion. I was wondering what I need to do about the following error caused by having too many tests.

```shell
.tox/py/lib/python3.11/site-packages/requests/models.py:1024: in raise_for_status
    raise HTTPError(http_error_msg, response=self)
E   requests.exceptions.HTTPError: 429 Client Error: Too Many Requests for url: https://raw.githubusercontent.com/dandi/schema/master/releases/0.4.4/dandiset.json
```

yarikoptic:
we had similar one before I believe at some level -- do we cache (as simple as lru_cache) fetching of those?

candleindark:
@yarikoptic In two remaining failed tests, [`test_dandimeta_datacite`](https://github.com/dandi/dandi-schema/blob/e1ca5af14ee515aaf318b46d0c9d8c53dd8c4be3/dandischema/datacite/tests/test_datacite.py#L371-L413) and [`test_datacite`](https://github.com/dandi/dandi-schema/blob/e1ca5af14ee515aaf318b46d0c9d8c53dd8c4be3/dandischema/datacite/tests/test_datacite.py#L114-L139), the failure was caused by lack of access to create DOIs for the EMBER-DANDI instance. For the DANDI-DANDI instance, we dedicate credential for testing purposes

https://github.com/dandi/dandi-schema/blob/e1ca5af14ee515aaf318b46d0c9d8c53dd8c4be3/dandischema/datacite/tests/test_datacite.py#L35-L40

Do you think that we need to test the creation DOIs for the EMBER-DANDI instance specifically as well? If not, I can just skip the call to `datacite_post()` in those tests.

candleindark:
@NEStock Do you have the DataCite credential for testing purposes in the EMBER project that you can share? If you do,  I can use them in the tests mentioned in https://github.com/dandi/dandi-schema/pull/294#issuecomment-2895816747.

.github/workflows/test.yml
```diff
@@ -9,21 +9,36 @@ on:
...
+          # 3.11
+          - python: '3.11'
+            vendored_env: dandi
+          - python: '3.11'
+            vendored_env: ember-dandi
+          # 3.12
+          - python: '3.12'
+            vendored_env: dandi
+          - python: '3.12'
+            vendored_env: ember-dandi
```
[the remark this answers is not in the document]
yarikoptic: for now may be make `datacite_doi_id` empty for ember-dandi, and make test conditioned on that value of not being empty

NEStock:
> @NEStock Do you have the DataCite credential for testing purposes in the EMBER project that you can share? If you do, I can use them in the tests mentioned in [#294 (comment)](https://github.com/dandi/dandi-schema/pull/294#issuecomment-2895816747).

@sandyhider is going to send them over!

.github/workflows/test.yml
```diff
@@ -42,7 +57,17 @@ jobs:
         python -m pip install --upgrade pip wheel
         python -m pip install --upgrade tox
 
+    - name: Set vendor environment
+      if: ${{ matrix.vendored_env != 'unvendored' }}
+      shell: bash
+      run: |
+        echo "DANDI_INSTANCE_NAME=${{ matrix.instance_name }}" >> "$GITHUB_ENV"
+        echo "DANDI_DATACITE_DOI_ID=${{ matrix.datacite_doi_id }}" >> "$GITHUB_ENV"
```
yarikoptic, line 65: that's where to not `echo` if not defined, or defined empty

.github/workflows/test.yml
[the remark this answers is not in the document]
yarikoptic: or just comment out ember-dandi for now and create a dedicated `adhoc-nodoi` one which will have name but no datacite_doi_id

.github/workflows/test.yml
```diff
@@ -42,7 +57,17 @@ jobs:
...
+
     - name: Run all tests
+      env:
+        DATACITE_DEV_PASSWORD: ${{ secrets.DATACITE_DEV_PASSWORD }}
+        NO_ET: 1
```
yarikoptic, line 70: if ember folks would provide password, we would need to make this specific to each vendored instance since DANDI and EMBER passwords would be different.

Also add variable `DATACITE_DEV_LOGIN` which is currently hardcoded to be `DARTLIB.DANDI` (DANDI specific), just for testing

sandyhider: I am sending your our datacite information in an email @yarikoptic 
use **dataciteinfo** to unencrypt it.

dandischema/models.py
yarikoptic: I still think that we should do above suggestion so we allow for instances which are not connected to dandicite (e.g. local ones), but still should be allowed to "publish" versions of dandisets on them.

dandischema/tests/utils.py
```diff
@@ -1,16 +1,46 @@
...
+
+INSTANCE_NAME = CONFIG.instance_name
+METADATA_DIR = Path(__file__).with_name("data") / "metadata"
+DANDISET_METADATA_DIR = METADATA_DIR / INSTANCE_NAME
+
+if CONFIG.instance_name == "DANDI" or CONFIG.instance_name == "DANDI-ADHOC":
+    DOI_PREFIX = "10.80507"
+elif CONFIG.instance_name == "EMBER-DANDI":
+    DOI_PREFIX = "10.60533"
+else:
```
yarikoptic, line 25:
```suggestion
if CONFIG.instance_name not in ("DANDI", "DANDI-ADHOC", "EMBER-DANDI"):
```

.github/workflows/test.yml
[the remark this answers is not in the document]
candleindark: > wouldn't this accomplish the same?
> 
> ```suggestion
>         os: [windows-2019, ubuntu-latest, macos-latest]
>         python: ['3.9', '3.10', '3.11', '3.12']
>         vendored_env: [unvendored]
>         include:
>           - vendored_env: dandi
>             instance_name: DANDI
>             datacite_doi_id: '80507'
>             python: '3.9'
>           - vendored_env: ember-dandi
>             instance_name: EMBER-DANDI
>             datacite_doi_id: '60533'
>             python: '3.9'
> ```

No, it doesn't, and since the entries in the `include` block don't have the `os` value defined, it is problematic as well.  To test vendorization with "smallest python version across all OSes" as you indicated in https://github.com/dandi/dandi-schema/pull/294#issuecomment-2891589558, and to express the testing matrix without an `exclude` block. We will need the following instead.

```yml
    os: [windows-2019, ubuntu-latest, macos-latest]
    python: ['3.9', '3.10', '3.11', '3.12']
    vendored_env: [unvendored]          # base jobs
    include:                            # explicit vendored jobs
      - os: windows-2019
        python: '3.9'
        vendored_env: dandi
        instance_name: DANDI
        datacite_doi_id: '80507'
      - os: ubuntu-latest
        python: '3.9'
        vendored_env: dandi
        instance_name: DANDI
        datacite_doi_id: '80507'
      - os: macos-latest
        python: '3.9'
        vendored_env: dandi
        instance_name: DANDI
        datacite_doi_id: '80507'
      - os: windows-2019
        python: '3.9'
        vendored_env: ember-dandi
        instance_name: EMBER-DANDI
        datacite_doi_id: '60533'
      - os: ubuntu-latest
        python: '3.9'
        vendored_env: ember-dandi
        instance_name: EMBER-DANDI
        datacite_doi_id: '60533'
      - os: macos-latest
        python: '3.9'
        vendored_env: ember-dandi
        instance_name: EMBER-DANDI
        datacite_doi_id: '60533'
```

which is, in my opinion, even more difficult to interpret than the solution in place.

yarikoptic:
while working on 
- https://github.com/dandi/dandi-archive/pull/2386
I was reminded of us having

```python
    repository: Optional[AnyHttpUrl] = Field(
        # mypy doesn't like using a string as the default for an AnyHttpUrl
        # attribute, so we have to convert it to an AnyHttpUrl:
        (
            TypeAdapter(AnyHttpUrl).validate_python(DANDI_INSTANCE_URL)
            if DANDI_INSTANCE_URL is not None
            else None
        ),
```
so here instances have not `None` but a specific value for the DANDI_INSTANCE_URL for the default value...  worth checking history of this change and seeing what we should do about that!

candleindark:
> while working on
> 
> * [Local runtime schema serialization endpoints dandi-archive#2386](https://github.com/dandi/dandi-archive/pull/2386)
>   I was reminded of us having
> 
> ```python
>     repository: Optional[AnyHttpUrl] = Field(
>         # mypy doesn't like using a string as the default for an AnyHttpUrl
>         # attribute, so we have to convert it to an AnyHttpUrl:
>         (
>             TypeAdapter(AnyHttpUrl).validate_python(DANDI_INSTANCE_URL)
>             if DANDI_INSTANCE_URL is not None
>             else None
>         ),
> ```
> 
> so here instances have not `None` but a specific value for the DANDI_INSTANCE_URL for the default value... worth checking history of this change and seeing what we should do about that!

That setup originates from https://github.com/dandi/dandi-schema/pull/100#issuecomment-1434810951, and it is committed at https://github.com/dandi/dandi-schema/pull/100/commits/cd0962fd23ff5f4fcf97c4501e07ecf8b098e3f1.

@yarikoptic What is your concern about this setup?

dandischema/models.py
```diff
@@ -1840,7 +1862,7 @@ class PublishedDandiset(Dandiset, Publishable):
     doi: str = Field(
```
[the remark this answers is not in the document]
candleindark: I think I can make this field go away completely if `DATACITE_DOI_ID` is not defined in the config. I.e. `PublishedDandiset` in that case will not include this field.  I think doing so is better solution.

.github/workflows/test.yml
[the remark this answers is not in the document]
yarikoptic: we do not need to test vendorization across all OSes per each vendor IMHO since vendorization is not OS-specific.  Shrink to a choice of 1 OS per each vendor, randomly.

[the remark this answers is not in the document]
candleindark: I reduced the workflow matrix so that one vendor is only associated with one OS. However, I use the same OS, ubuntu-latest, across different vendors, since I want the difference in vendor to be the only variable.

dandischema/models.py
candleindark: This done, as discussed in the meeting, the `doi` field in `PublishedDandiset` is defined conditionally.

candleindark:
> we had similar one before I believe at some level -- do we cache (as simple as lru_cache) fetching of those?

The caching doesn't work across different CI jobs. Now, with the number of CI jobs reduced, the problem no longer exists.

dandischema/metadata.py
```diff
@@ -123,8 +123,8 @@ def generate_context() -> dict:
 
     for item in models.DigestType:
         fields[item.value] = {"@id": item.value, "@nest": "digest"}
-    fields["Dandiset"] = "dandi:Dandiset"
-    fields["Asset"] = "dandi:Asset"
+    fields["Dandiset"] = f"{models.DANDI_NSKEY}:Dandiset"
+    fields["Asset"] = f"{models.DANDI_NSKEY}:Asset"
```
satra, line 127: these should not change as those prefixes are not urls just keys in jsonld - they won't resolve. the lower case `dandi` is simply a uri.

dandischema/tests/utils.py
```diff
@@ -35,6 +59,6 @@ def _basic_publishmeta(
             "schemaKey": "PublishActivity",
         },
         "version": version,
-        "doi": f"{prefix}/dandi.{dandi_id}/{version}",
+        "doi": f"{prefix}/{instance_name.lower()}.{dandi_id}/{version}",
```
satra, line 62: shouldn't `dandi_id` change here?

satra: it was deleted in the lines above.

satra: never mind. i did not read well.

dandischema/models.py
```diff
@@ -58,8 +64,14 @@
 )
 ASSET_UUID_PATTERN = r"^dandiasset:" + UUID_PATTERN
 VERSION_PATTERN = r"\d{6}/\d+\.\d+\.\d+"
-DANDI_DOI_PATTERN = rf"^10.(48324|80507)/dandi\.{VERSION_PATTERN}"
-DANDI_PUBID_PATTERN = rf"^DANDI:{VERSION_PATTERN}"
+DANDI_DOI_PATTERN = (
+    (rf"^10\.{DATACITE_DOI_ID_PATTERN}/{ID_PATTERN.lower()}\.{VERSION_PATTERN}$")
```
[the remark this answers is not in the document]
candleindark: @asmacdo Please see post below.

dandischema/conf.py
```diff
@@ -0,0 +1,68 @@
...
+        Annotated[
+            str, StringConstraints(pattern=rf"^{_UNVENDORED_DATACITE_DOI_ID_PATTERN}$")
+        ]
+    ] = None
+    """
+    The registrant code of the DOI prefix at DataCite
+
+    The number sequence that follows "10." within the DOI prefix as documented
+    at https://support.datacite.org/docs/prefixes.
+    """
```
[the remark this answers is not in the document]
candleindark: @yarikoptic The only use of `datacite_doi_id` is to create DOI prefix currently.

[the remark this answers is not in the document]
yarikoptic: fine with me.

dandischema/tests/utils.py
```diff
@@ -1,16 +1,49 @@
...
+DANDISET_METADATA_DIR = METADATA_DIR / INSTANCE_NAME
+
+DOI_PREFIX = f"10.{DATACITE_DOI_ID}" if DATACITE_DOI_ID is not None else None
+
+
+skipif_no_datacite_auth = pytest.mark.skipif(
+    os.getenv("DATACITE_DEV_LOGIN") is None
+    or os.getenv("DATACITE_DEV_PASSWORD") is None,
+    reason="no datacite login or password set in environment variables",
+)
```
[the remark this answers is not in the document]
yarikoptic: to be more friendly to contributors let's proceed with this, BUT do it in a separate PR since unrelated to this PR which comes from the branch in this repo.

dandischema/conf.py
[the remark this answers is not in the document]
candleindark: @asmacdo FYI: This change is now part of this PR.

dandischema/tests/utils.py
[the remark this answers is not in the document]
yarikoptic: ok, let's do it here then.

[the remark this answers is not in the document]
candleindark: done

dandischema/conf.py
```diff
@@ -0,0 +1,66 @@
...
+
+    @property
+    def id_pattern(self) -> str:
+        """Regex pattern for the prefix of identifiers"""
+        if "instance_name" in self.model_fields_set:
+            return self.instance_name
+
+        # If the instance name is not set,
+        #   we use a pattern for unvendored DANDI instances
+        return _UNVENDORED_ID_PATTERN
```
candleindark, line 52:
```suggestion
    @property
    def id_pattern(self) -> str:
        """Regex pattern for the prefix of identifiers"""
        return self.instance_name
```

I think we should simplify `id_pattern` to just the above. Otherwise, vendorizing  the CLI according to the JSON serialization of the `Config` object on the server can be overly complicated or inconsistent with the vendorization on the server.

yarikoptic: please remind me why we need `id_pattern` at all, or in other words -- how is it used?

It seems that current (might have forgot) course of thinking is that every instance has an `instance_name`, which should pass `_UNVENDORED_ID_PATTERN` check at "config specification time". And then we would just base everything against that `instance_name`, as all further schema regexes which use name in them would just use it, right?

candleindark: Your description is what the proposed change above entails.

The current solution, the one before the proposed change,

```py
    @property
    def id_pattern(self) -> str:
        """Regex pattern for the prefix of identifiers"""
        if "instance_name" in self.model_fields_set:
            return self.instance_name
        # If the instance name is not set,
        #   we use a pattern for unvendored DANDI instances
        return _UNVENDORED_ID_PATTERN
```

sets the property `id_pattern` to `_UNVENDORED_ID_PATTERN` if the field `instance_name` is not set, i.e., defaulted to `"DANDI-ADHOC"`, and only sets the property `id_pattern` to `self.instance_name` when the field `instance_name` is explicitly set.

Because the currently definition of `id_pattern` bases not only on the value of `self.instance_name` but also on how the value of `self.instance_name` comes to be, set by user code or defaulted, it creates complications in achieving the same behavior in both the dandi-archive and dandi-cli.

In the proposed change, one may argue that `id_pattern` is unnecessary since it is exactly the same as `self.instance_name`. However, I prefer to keep it because `id_pattern` is not the same as `instance_name` semantically, and it would be nice to have `Config` class to decide what the ID pattern is without relying on some external code.

yarikoptic: well, I would argue that if we do not have/use pattern here -- we better not pretend that we do!  then all the regexes where such `id_pattern` is used would make more sense to people since they would explicitly just list the `instance_name`.

yarikoptic: Per our discussion -- Let's move this out to be logic directly within models.py where we have all other logic which is based on config, and here have only what to be user-facing config variables.

dandischema/models.py
```diff
@@ -36,11 +36,18 @@
 from pydantic_core import CoreSchema
 from zarr_checksum.checksum import InvalidZarrChecksum, ZarrDirectoryDigest
 
+from dandischema.conf import get_instance_config
+
```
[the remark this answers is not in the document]
candleindark: What about making the instance specific definitions in to `instance_moddels.py` instead? The name is more to the point that way.

Regardless of which new file name to use, my concern is that by doing so, we are introducing many more moving parts in this vendorization effort, which though useful but may not be necessary at this point, making integration with downstream components, dandi-cli and dandi-archive, much more complex.

dandischema/models.py
```diff
@@ -36,11 +36,26 @@
...
 from .consts import DANDI_SCHEMA_VERSION
 from .digests.dandietag import DandiETag
 from .types import ByteSizeJsonSchema
 from .utils import name2title
 
+# Load needed configurations into constants
+_INSTANCE_CONFIG = get_instance_config()
```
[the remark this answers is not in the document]
jjnesbitt: > The fact that configuration needs to happen before import is somewhat orthogonal given current design since `config` is anyways is not an argument to any of those model classes.

I don't believe it's orthogonal at all, it's actually the reason I'm proposing this change. While `Config` is not an "argument" to those model classes, they use values constructed from it, like `DOI_PREFIX_PATTERN`. As it stands those derived values are instantiated once the file is imported, which means they are static after that. 

Currently, if the file is imported, and then `set_instance_config` is called afterwards, it has no effect on any of the models, since the string values used in the models have already been constructed at the top of the file. If what I'm proposing is implemented, you could import the file, call `set_instance_config` afterwards, and the changed config would be correctly propagated to the models.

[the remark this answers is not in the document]
candleindark: > we used such a pattern for `id_pattern` property at some point, but I insisted to switch from that IIRC so it doesn't become a part of serialized configuration itself.
> 
> The fact that configuration needs to happen before import is somewhat orthogonal given current design since `config` is anyways is not an argument to any of those model classes. @candleindark might recall more of reasoning behind not going properties way

We removed the `id_pattern` not because that it would become part of the JSON serialization of the instance config (properties in Pydantic models are not serialized by default). The rationale behind was that pattern entities such `id_pattern`, the ID patterns in constructing the models, are pertinent to the models and belongs to `dandischema/models.py`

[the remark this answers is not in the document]
yarikoptic: As there is  work towards integrating 
- https://github.com/dandi/dandi-schema/pull/316

I would consider this discussion resolved here.

candleindark:
@yarikoptic Merge conflicts resolved again. Please take a final look and merge and release.

PR #296, opened 2025-04-11
Increase `Description` field max character length

### LINKED ISSUE #295: Allow for more than 3,000 characters in the DLP `Description` field

kabilar:
Fix #295 

10,000 characters is about 2-3 single-spaced pages.  Based on the OpenScope use case, the dataset descriptions (https://openscopedatafront.web.app/datasets) will fit within 10,000 characters.  See here for the context: https://github.com/dandi/dandi-archive/issues/2297.

### DIFF
--- a/dandischema/consts.py
+++ b/dandischema/consts.py
@@ -1,4 +1,4 @@
-DANDI_SCHEMA_VERSION = "0.6.10"
+DANDI_SCHEMA_VERSION = "0.6.11"
 ALLOWED_INPUT_SCHEMAS = [
     "0.4.4",
     "0.5.1",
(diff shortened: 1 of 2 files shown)


kabilar:
@yarikoptic Not sure why the tests are failing but it appears to be unrelated.

candleindark (approved):
This change should be fine. However, I think we will get a new JSON schema because of this change. In `consts.py`, you may want to change `DANDI_SCHEMA_VERSION` 
 to `"0.6.11"` and add `"0.6.10"` to `ALLOWED_INPUT_SCHEMAS`

kabilar:
> This change should be fine. However, I think we will get a new JSON schema because of this change. In `consts.py`, you may want to change `DANDI_SCHEMA_VERSION` to `"0.6.11"` and add `"0.6.10"` to `ALLOWED_INPUT_SCHEMAS`

Thanks.  I have now updated the `DANDI_SCHEMA_VERSION`.

kabilar:
Looks like these CI failures are unrelated.  Merging.

PR #297, opened 2025-04-22
enh: allow creation of dandiset dois (contrasted to a version doi)

asmacdo:


dandischema/datacite/__init__.py
```diff
@@ -67,16 +74,23 @@
 
 def to_datacite(
     meta: Union[dict, PublishedDandiset],
+    event: str = None,
     validate: bool = False,
-    publish: bool = False,
```
yarikoptic, line 71: Let's ease transition via a deprecation cycle

- leave `publish` in (so code in dandi-archive could still use it as it does ATM)
  - default it to `None`. If `bool` type provided -- issue DeprecationWarning informing to use `event`, treat bool accordingly to set value of `event` while first checking if not set in parallel -- then `ValueError` -- only one should be provided
- add `event` **after** publish, so anything using it as positional arg, would not collide (we might want to add `*`)

asmacdo: Actually I got a little ahead of the other PR here, this change isn't used as it is currently.

This change does open the door to "hide" findable DOIs (converting to "Registered"), so depending on how the design doc changes this might just get removed.

asmacdo:
https://github.com/dandi/dandi-archive/pull/2350/files#r2054748163

If version_doi is False, we should point to the DLP instead of a specific version, which will prevent needing additional updates.

dandischema/models.py
```diff
@@ -57,12 +57,22 @@
...
+
+PUBLISHED_DANDISET_URL_PATTERN = (
+    rf"^{DANDI_INSTANCE_URL_PATTERN}/dandiset/{DANDI_ID_PATTERN}"
+)
 PUBLISHED_VERSION_URL_PATTERN = (
     rf"^{DANDI_INSTANCE_URL_PATTERN}/dandiset/{VERSION_PATTERN}$"
 )
+PUBLISHED_URL_PATTERN = (
+    rf"{PUBLISHED_VERSION_URL_PATTERN}|{PUBLISHED_DANDISET_URL_PATTERN}"
+)
```
asmacdo, line 75: TODO: this has gotten really messy, it can be cleaned up (i just hacked it out to work for now)

candleindark: Some of those will be outdated soon. We are modifying some of them in https://github.com/dandi/dandi-schema/pull/294. The goal for that PR is not to clean the mess though.

asmacdo: @candleindark thanks for lookin out, nice catch!

dandischema/datacite/__init__.py
```diff
@@ -66,16 +79,57 @@
 
 
 def to_datacite(
-    meta: Union[dict, PublishedDandiset],
+    meta: Union[dict, PublishedDandiset, DraftDandiset],
```
yarikoptic, line 82:
```suggestion
    meta: Union[dict, Dandiset],
```

dandischema/datacite/__init__.py
```diff
@@ -66,16 +79,57 @@
...
+            meta = PublishedDandiset(**meta)
+        except ValidationError as e:
+            try:
+                # If that fails, try DraftDandiset
+                logger.info("Falling back to DraftDandiset validation")
+                # Create a copy to avoid modifying the original dict
+                draft_meta = meta.copy()
+                # Set the correct schemaKey for DraftDandiset
+                draft_meta["schemaKey"] = "DraftDandiset"
+                meta = DraftDandiset(**draft_meta)
```
yarikoptic, line 107:
```suggestion
    (PublishedDandiset or Dandiset) based on the metadata state.
    """
    # Try to convert dict to model if needed
    if isinstance(meta, dict):
        try:
            # First try PublishedDandiset
            meta = PublishedDandiset(**meta)
        except ValidationError as e:
            try:
                # If that fails, try DraftDandiset
                logger.info("Falling back to Dandiset validation")
                meta = Dandiset.model_construct(**meta)
```
and so on...

yarikoptic (changes requested):
avoid creating new classes if feasible

asmacdo:
@candleindark I dont think its as bad to be unvalidated here as it might seem, but it does seem like we are bypassing the value of using the Pydantic models for this. Normal validation is still going on under the hood for Dandi itself, this is just executed *after* the data has been saved to the db, just validating the data we are sending to Datacite. If that doesnt conform to our models, thats ok, and if it doesnt conform to their spec, we will just log and move on.

dandischema/models.py
asmacdo: I wonder if the model changes are still necessary at all? I've gone through several implementations that were trying to pass all the data through Pydantic validation first, and we've also made the decision to store the dandiset-wide doi on the draft version. If we aren't going to soften the Dandiset pydantic model so we can use that for a draft dandiset, I'm not sure if the draft could ever pass PublishedDandiset validation. So there isnt really a need to change the regex to accept that other format of doi...

dandischema/datacite/__init__.py
```diff
@@ -65,17 +80,102 @@
...
+    to build the model without validation but with properly handled nested types.
+    """
+    if isinstance(meta, dict):
+        meta = deepcopy(meta)
+        try:
+            meta = PublishedDandiset(**meta)
+        except ValidationError:
+            # mypy can't track that meta is still dict after failed PublishedDandiset(**meta)
+            assert isinstance(meta, dict)
+            if meta.get("version") == "draft":
```
asmacdo, line 150: if the version is a draft, it wont have fields like datePublished. However this can also happen when we are creating a Dandiset DOI from a published version-- in this case, the metadata is the published version, but the doi and the url fields wont pass validation (they both won't include the version). 

Previously I modified the PublishedDandiset to accept either format for url and doi, but I dont think that really makes sense-- those aren't valid for a published dandiset, and we wouldnt want our output schema to reflect flexibility thats not really there.

asmacdo:
This work may still be useful if https://github.com/dandi/dandi-archive/pull/2799 goes forward, but its so far out now its probably best to close and use this as a reference to create a new branch if needed.

PR #301, opened 2025-05-20
Add version sorting helper using packaging.version

yarikoptic:
- Add sort_versions utility function to properly sort version strings
- Update error messages to use this function for better readability
- Add packaging as a dependency
- Add tests for the new function

🤖 Generated with [Claude Code](https://claude.ai/code)

@yarikoptic observed in some CI run 

```
"message": "Metadata version 0.6.10 is not allowed. Allowed are: 0.6.9, 0.4.4, 0.5.1, 0.5.2, 0.6.0, 0.6.1, 0.6.2, 0.6.3, 0.6.4, 0.6.5, 0.6.6, 0.6.7, 0.6.8.",
```
and got confused since missed that 0.6.9 was listed at all ... 

With this change we should get more consistent reporting

the only possible gotcha I am afraid off, whenever we manage to sneak in non 'x.y.z' version, but that should not happen AFAIK.

### DIFF
--- a/dandischema/consts.py
+++ b/dandischema/consts.py
@@ -23,6 +23,8 @@ ALLOWED_TARGET_SCHEMAS = [DANDI_SCHEMA_VERSION]
 
 # This allows multiple schemas for validation, whereas target schemas focus on
 # migration.
+# We use simple sorted() here to avoid circular imports, but use
+# utils.sort_versions() if to output to the user
 ALLOWED_VALIDATION_SCHEMAS = sorted(
     set(ALLOWED_INPUT_SCHEMAS).union(ALLOWED_TARGET_SCHEMAS)
 )
--- a/dandischema/metadata.py
+++ b/dandischema/metadata.py
@@ -24,6 +24,7 @@ from .utils import (
     dandi_jsonschema_validator,
     json_object_adapter,
     sanitize_value,
+    sort_versions,
     strip_top_level_optional,
     validate_json,
     version2tuple,
@@ -215,7 +216,7 @@ def _get_jsonschema_validator(
     if schema_version not in ALLOWED_VALIDATION_SCHEMAS:
         raise ValueError(
             f"DANDI schema version {schema_version} is not allowed. "
-            f"Allowed are: {', '.join(ALLOWED_VALIDATION_SCHEMAS)}."
+            f"Allowed are: {', '.join(sort_versions(ALLOWED_VALIDATION_SCHEMAS))}."
         )
     if schema_key not in SCHEMA_MAP:
         raise ValueError(
@@ -312,7 +313,7 @@ def validate(
     if schema_version not in ALLOWED_VALIDATION_SCHEMAS and schema_key in SCHEMA_MAP:
         raise ValueError(
             f"Metadata version {schema_version} is not allowed. "
-            f"Allowed are: {', '.join(ALLOWED_VALIDATION_SCHEMAS)}."
+            f"Allowed are: {', '.join(sort_versions(ALLOWED_VALIDATION_SCHEMAS))}."
         )
     if json_validation:
         if schema_version == DANDI_SCHEMA_VERSION:
@@ -488,7 +489,7 @@ def _add_asset_to_stats(assetmeta: Dict[str, Any], stats: _stats_type) -> None:
     if schema_version not in ALLOWED_INPUT_SCHEMAS:
         raise ValueError(
             f"Metadata version {schema_version} is not allowed. "
-            f"Allowed are: {', '.join(ALLOWED_INPUT_SCHEMAS)}."
+            f"Allowed are: {', '.join(sort_versions(ALLOWED_INPUT_SCHEMAS))}."
         )
 
     stats["numberOfBytes"] = stats.get("numberOfBytes", 0)
--- a/dandischema/tests/test_utils.py
+++ b/dandischema/tests/test_utils.py
@@ -14,6 +14,7 @@ from dandischema.utils import (
     jsonschema_validator,
     name2title,
     sanitize_value,
+    sort_versions,
     strip_top_level_optional,
     validate_json,
     version2tuple,
@@ -97,6 +98,21 @@ def test_sanitize_value() -> None:
     assert sanitize_value("A;B") == "A-B"
     assert sanitize_value("A\\/B") == "A--B"
     assert sanitize_value("A\"'B") == "A--B"
+
+
+def test_sort_versions() -> None:
+    """Test that sort_versions correctly sorts version strings semantically."""
+    # Test for correct semantic version sorting
+    unsorted_versions = ["0.6.2", "0.6.10", "0.6.1", "0.5.1", "0.6.0", "0.4.4"]
+    expected_sort = ["0.4.4", "0.5.1", "0.6.0", "0.6.1", "0.6.2", "0.6.10"]
+
+    assert sort_versions(unsorted_versions) == expected_sort
+
+    # Test with empty list
+    assert sort_versions([]) == []
+
+    # Test with single element
+    assert sort_versions(["0.6.0"]) == ["0.6.0"]
 
 
 @pytest.fixture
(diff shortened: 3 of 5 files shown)


setup.cfg
```diff
@@ -29,6 +29,7 @@ project_urls =
 python_requires = >=3.9
 install_requires =
     jsonschema[format]
+    packaging
```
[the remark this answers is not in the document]
yarikoptic: aren't you ... detailed little thing, ok

```suggestion
    packaging>=14.0
```

where `Version` was added:

```
❯ python
Python 3.12.10 (main, Apr  9 2025, 04:44:59) [GCC 14.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from packaging.version import Version
*>>> Version("0.10") > Version("0.2")
True
>>> import packaging
>>> packaging.__version__
'14.0'
```

dandischema/utils.py
```diff
@@ -1,11 +1,12 @@
 from __future__ import annotations
 
 import re
-from typing import Any, Iterator, List, Union, cast, get_args, get_origin
+from typing import Any, Iterator, List, Sequence, Union, cast, get_args, get_origin
```
candleindark, line 4: This is deprecated since py3.9. The prefer one is  `collections.abc.Sequence`

dandischema/consts.py
```diff
@@ -23,6 +23,8 @@
 
 # This allows multiple schemas for validation, whereas target schemas focus on
 # migration.
+# We use simple sorted() here to avoid circular imports, but use
+# utils.sort_versions() if to output to the user
 ALLOWED_VALIDATION_SCHEMAS = sorted(
     set(ALLOWED_INPUT_SCHEMAS).union(ALLOWED_TARGET_SCHEMAS)
 )
```
candleindark, line 30:
```suggestion
ALLOWED_VALIDATION_SCHEMAS = sorted(
    set(ALLOWED_INPUT_SCHEMAS).union(ALLOWED_TARGET_SCHEMAS), key=Version
)
```

I understand the need to avoid circular definition, and not use `sort_versions` here. However, I would like to have `ALLOWED_VALIDATION_SCHEMAS` sorted correctly in the first place. I think it will be hard to remember or enforce everyone to remember to resorting `ALLOWED_VALIDATION_SCHEMAS` when it is output to the user.

With this solution, we don't even define the `sort_versions()` function, but the function is a one line function, and  the behavior of `packaging.version.Version` is well documented.

The other less preferable solution to me is to define the `sort_versions()` function in this file, `consts.py`. However, that will contradict with the name of this file. We can change the name to `_sort_versions()` though.

dandischema/tests/test_utils.py
```diff
@@ -99,6 +100,21 @@ def test_sanitize_value() -> None:
     assert sanitize_value("A\"'B") == "A--B"
 
 
+def test_sort_versions() -> None:
```
candleindark, line 103: The use `@pytest.mark.parametrize` can help reduce duplication in logic and make the logic more apparent.

dandischema/consts.py
yarikoptic: cute, let's do, but need to import `from packaging.version import Version as _Version` so we do not pollute the namespace with imports.

dandischema/utils.py
yarikoptic: hm, indeed it is as many others https://docs.python.org/3/library/typing.html#typing.Sequence as [List](https://docs.python.org/3/library/typing.html#typing.List) and others... we better eventually uniformly adjust such imports

yarikoptic: ok, did that for now

dandischema/metadata.py
```diff
@@ -24,6 +24,7 @@
     dandi_jsonschema_validator,
     json_object_adapter,
     sanitize_value,
+    sort_versions,
```
candleindark, line 27:
```suggestion
```

dandischema/metadata.py
```diff
@@ -215,7 +216,7 @@ def _get_jsonschema_validator(
     if schema_version not in ALLOWED_VALIDATION_SCHEMAS:
         raise ValueError(
             f"DANDI schema version {schema_version} is not allowed. "
-            f"Allowed are: {', '.join(ALLOWED_VALIDATION_SCHEMAS)}."
+            f"Allowed are: {', '.join(sort_versions(ALLOWED_VALIDATION_SCHEMAS))}."
```
candleindark, line 219:
```suggestion
            f"Allowed are: {', '.join(ALLOWED_VALIDATION_SCHEMAS)}."
```

dandischema/metadata.py
```diff
@@ -312,7 +313,7 @@ def validate(
     if schema_version not in ALLOWED_VALIDATION_SCHEMAS and schema_key in SCHEMA_MAP:
         raise ValueError(
             f"Metadata version {schema_version} is not allowed. "
-            f"Allowed are: {', '.join(ALLOWED_VALIDATION_SCHEMAS)}."
+            f"Allowed are: {', '.join(sort_versions(ALLOWED_VALIDATION_SCHEMAS))}."
```
candleindark, line 316:
```suggestion
            f"Allowed are: {', '.join(ALLOWED_VALIDATION_SCHEMAS)}."
```

dandischema/metadata.py
```diff
@@ -488,7 +489,7 @@ def _add_asset_to_stats(assetmeta: Dict[str, Any], stats: _stats_type) -> None:
     if schema_version not in ALLOWED_INPUT_SCHEMAS:
         raise ValueError(
             f"Metadata version {schema_version} is not allowed. "
-            f"Allowed are: {', '.join(ALLOWED_INPUT_SCHEMAS)}."
+            f"Allowed are: {', '.join(sort_versions(ALLOWED_INPUT_SCHEMAS))}."
```
candleindark, line 492:
```suggestion
            f"Allowed are: {', '.join(ALLOWED_INPUT_SCHEMAS)}."
```

dandischema/tests/test_utils.py
```diff
@@ -14,6 +14,7 @@
     jsonschema_validator,
     name2title,
     sanitize_value,
+    sort_versions,
```
candleindark, line 17:
```suggestion
```

dandischema/tests/test_utils.py
```diff
@@ -99,6 +100,21 @@ def test_sanitize_value() -> None:
...
+    expected_sort = ["0.4.4", "0.5.1", "0.6.0", "0.6.1", "0.6.2", "0.6.10"]
+
+    assert sort_versions(unsorted_versions) == expected_sort
+
+    # Test with empty list
+    assert sort_versions([]) == []
+
+    # Test with single element
+    assert sort_versions(["0.6.0"]) == ["0.6.0"]
+
```
candleindark, line 116: Since we no longer need `sort_versions()`, we don't need its tests.

dandischema/utils.py
```diff
@@ -1,11 +1,13 @@
 from __future__ import annotations
 
+from collections.abc import Sequence
```
candleindark, line 3:
```suggestion
```

dandischema/utils.py
```diff
@@ -1,11 +1,13 @@
...
 import re
 from typing import Any, Iterator, List, Union, cast, get_args, get_origin
 
 from jsonschema import Draft7Validator, Draft202012Validator
 from jsonschema.protocols import Validator as JsonschemaValidator
 from jsonschema.validators import validator_for
+from packaging.version import Version
```
candleindark, line 10:
```suggestion
```

candleindark (commented):
I have removed `sort_versions()`, it uses, and its tests. As we discussed earlier, it is not needed if `ALLOWED_VALIDATION_SCHEMAS` is sorted properly when it is initialized.

dandischema/tests/test_utils.py
```diff
@@ -99,6 +100,21 @@ def test_sanitize_value() -> None:
...
+
```
yarikoptic, line 117:
```suggestion
```

PR #305, opened 2025-06-04
Replace GH runner image `windows-2019` with `windows-latest`

candleindark:
The `windows-2019` image is deprecated. For more details, checkout https://github.com/actions/runner-images/issues/12045

### DIFF
--- a/.github/workflows/test-dandi-cli.yml
+++ b/.github/workflows/test-dandi-cli.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - windows-2019
+          - windows-latest
           - ubuntu-latest
           - macos-latest
         python:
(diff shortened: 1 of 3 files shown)


candleindark:
The Windows runner being used for this PR is still `windows-2019`. I guess for the changes in this PR to be applied, this PR has to be merged first.

yarikoptic (changes requested):
FWIW, I checked git log and it seems that we just evolved from specific versions to `-latest` for linux and osx but never for windows. (wanted to make sure we didn't limit it for some specific reason). FWIW -- as you still see it above you can see that we have it used in multiple workflows:

```
❯ git grep windows-2019
.github/workflows/test-dandi-cli.yml:          - windows-2019
.github/workflows/test-dandi-cli.yml:          - os: windows-2019
.github/workflows/test-nonetwork.yml:          - windows-2019
.github/workflows/test.yml:        os: [windows-2019, ubuntu-latest, macos-latest]

```

and better be replaced in all of them together here.

yarikoptic:
note:
image
is likely a fluke of sending follow up quickly with renamed one, so then it still waits for prior one. PR modified 3 files, so all should be addressed:

```
❯ git grep windows-2019
❯ 
```

PR #306, opened 2025-06-05
Provide mechanism to reset instance configuration by function arguments, not env vars

candleindark:
This is a replacement of https://github.com/dandi/dandi-schema/pull/304, which has been marked as merged to the `devendorize` branch however changes in `devendorize` was later written.

This PR provide a mechanism to reset DANDI instance configuration by function arguments, not env vars. This future is a step in supporting `dandi-cli` to initialize DANDI schema models to a particular vendor in order to communicate with a DANDI service of a particular vendor.

**Breaking change**
1. `migrate()` and `validate()` are not available at `dandischema` level. To use them, one must do `from dandischema.metadata import migrate, validate` instead of `from dandischema import migrate, validate`.

### DIFF
--- a/dandischema/__init__.py
+++ b/dandischema/__init__.py
@@ -1,4 +1,3 @@
-__all__ = ["__version__", "migrate", "validate"]
+__all__ = ["__version__"]
 
 from ._version import __version__
-from .metadata import migrate, validate
--- a/dandischema/conf.py
+++ b/dandischema/conf.py
@@ -2,13 +2,19 @@
 
 from __future__ import annotations
 
+import logging
 from typing import Annotated, Any, Optional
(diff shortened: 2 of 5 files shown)


yarikoptic:
Re breaking change -- please check if with `auto` there is some what to enhance changelog message with some info from within PR description?  since it would be hard to fit everything into the PR title and we do not use `scriv` or alike here like we do in datalad.

candleindark:
> well, note that this PR is to be merged to `devendorize`, so for `auto` what would likely matter is just such section in that #294 where we have all the gory changes

I just copied the "Release Notes" section in this PR to #294 in case `auto` is not smart enough.

PR #309, opened 2025-07-09
Escape `<` and `>` properly in `description`

mvandenburgh:
Fixes dandi/dandi-archive#2432 

Currently, the `<` and `>` characters in the description field of the help tooltip are not escaped, which can causes a rendering issue where every character in between the `<` and `>` is not rendered.

Screenshot of meditor after this change -

Screenshot from 2025-07-09 12-41-52

### DIFF
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -1048,7 +1048,7 @@ class Resource(DandiBaseModel):
     relation: RelationType = Field(
         title="Resource relation",
         description="Indicates how the resource is related to the dataset. "
-        "This relation should satisfy: dandiset <relation> resource.",
+        r"This relation should satisfy: dandiset \<relation\> resource.",
         json_schema_extra={"nskey": "dandi"},
     )
     resourceType: Optional[ResourceType] = Field(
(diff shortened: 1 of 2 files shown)


mvandenburgh:
The failing type check in CI seems to be unrelated, and [is also happening on `master`](https://github.com/dandi/dandi-schema/actions/runs/16136891563/job/45535026456)

satra:
@mvandenburgh - should this be done here or in the meditor? it seems the fix is specific to a particular rendition of the ui rather than the schema itself or other contexts where the field is displayed.

dandischema/models.py
```diff
@@ -1048,7 +1048,7 @@ class Resource(DandiBaseModel):
     relation: RelationType = Field(
         title="Resource relation",
         description="Indicates how the resource is related to the dataset. "
-        "This relation should satisfy: dandiset <relation> resource.",
+        r"This relation should satisfy: dandiset \<relation\> resource.",
```
candleindark, line 1051: I agree with @satra that this is about rendering of the text in the UI and should be done in the UI level. Making modification here also, unfortunately, changes the text itself, not just its rendering.

PR #316, opened 2025-07-31
Support dynamic configuration

jjnesbitt:
This allows support for setting `Config` dynamically (vendorization, for example), without needing to rely on module import order.

@yarikoptic I believe this removes the need for the `clear_dandischema_modules_and_set_env_vars` conftest function, as config no longer relies on module import, but I'm still looking into that, which is why this is draft.


dandischema/models.py
```diff
@@ -1863,25 +1847,16 @@ class Publishable(DandiBaseModel):
...
     id: str = Field(
         description="Uniform resource identifier.",
-        pattern=DANDI_PUBID_PATTERN,
         json_schema_extra={"readOnly": True},
     )
-    doi: str = Field(**_doi_field_kwargs)
+    doi: str = Field(
+        title="DOI",
+        json_schema_extra={"readOnly": True, "nskey": DANDI_NSKEY},
+        default="",
```
candleindark, line 1858: We require `doi` value when DANDI DOI pattern is not available. Defaulting to `"""` will fail the `check_id` method below, but the feedback the user getting will be different.

candleindark (commented):
At this point, I haven't check all the details for correctness, but I can see that with this approach, the calling order restriction of `dandischema.conf.set_instance_config()` in relation to the import of `dandischema.models` has indeed been lifted. However, to achieve that, the validations affected by the value of the instance config in `dandischema.conf` have been moved into custom Pydantic validators, i.e., the `field_validator`s. The validation behaviors of those validators exist only in Python runtime and are not encoded in the corresponding JSON Schema schemas of the models.

I think, in general, we want to move validation out of custom Pydantic validators into validation that can be encoded into JSON Schema schemas. @yarikoptic, you input?

yarikoptic:
re 

> I think, in general, we want to move validation out of custom Pydantic validators into validation that can be encoded into JSON Schema schemas. @yarikoptic, your input?

and related

> that will make the eventual transitioning to LinkML more complex, and from now on we have to manage the serialization of those involved elements instead of being done by Pydantic.
> ...
> I like the idea of separation of model from code, so I am hesitant in moving more of the model specs to the code by using those custom validators and respective serializers.

Although I am with you on the ultimate desires/design, an immediate target is to provide support for multiple instances with current existing setup of pydantic + jsonschema, with minimal amount of "user visible changes" (i.e. not changing much if anything in "DANDI vendorized schema"). So I would say -- we can go and move into a few more python validations and serializers *for now*. It would also help to identify such points better for when we re-approach expressing it in linkml again.

dandischema/models.py
jjnesbitt: Ah I see, that's an oversight on my part.

jjnesbitt:
@candleindark I unfortunately don't have time to get this PR to 100%. However, it is almost there (90%). As far as I can tell, the big remaining item is [this](https://github.com/dandi/dandi-schema/pull/316#discussion_r2246058093) comment you left, regarding the doi default value. I believe a default could be supplied in the `check_doi` method, but the issue is that [other parts](https://github.com/dandi/dandi-schema/blob/05694fbe6650d1286e34495290bf3a3cf7b60890/dandischema/tests/utils.py#L10-L13) of the testing code make use of a statically defined instance config. In order to properly fix how DOI is handled, I think those constants (and anywhere they're made use of) need to be updated to use a dynamic instance config.

Do you think you can take this on?

dandischema/models.py
```diff
@@ -118,7 +70,10 @@ class LicenseType(Enum):
 else:
     LicenseType = Enum(
         "LicenseType",
-        [(license_.name, license_.value) for license_ in _INSTANCE_CONFIG.licenses],
+        [
+            (license_.name, license_.value)
+            for license_ in get_instance_config().licenses
```
candleindark, line 75: This line prevents `dandischema` to become truly dynamic, the purpose of this PR, and there is no way around it as long as `LicenseType` is defined as an `Enum` at the module level. Once this definition of `LicenseType` is executed, changes in the value returned by `get_instance_config()` do not alter the value of `LicenseType`.

A way to make `LicenseType` dynamic, as [suggested by ChatGPT](https://chatgpt.com/share/68a3f190-0288-800a-9820-7f5b9f954fb7), is to define it as custom type with hooks for Pydantic to validation and JSON schema generation, such as the following.

```py
# types.py (or nearby)
from pydantic import GetJsonSchemaHandler
from pydantic.json_schema import JsonSchemaValue
from pydantic_core import core_schema, PydanticCustomError
from dandischema.conf import get_instance_config

def _current_license_values() -> list[str]:
    # your config objects already have .value equal to "scheme:id"
    return [lic.value for lic in get_instance_config().licenses]

class DynamicLicense(str):
    @classmethod
    def __get_pydantic_core_schema__(cls, _source, _handler):
        def validate(v):
            s = str(v)
            allowed = _current_license_values()
            if s not in allowed:
                raise PydanticCustomError(
                    'license_value',
                    'Invalid license: {val}. Allowed: {allowed}',
                    {'val': s, 'allowed': allowed},
                )
            return s
        return core_schema.no_info_after_validator_function(
            validate, core_schema.str_schema()
        )

    @classmethod
    def __get_pydantic_json_schema__(cls, core_schema, handler: GetJsonSchemaHandler) -> JsonSchemaValue:
        schema = handler(core_schema)
        schema['enum'] = _current_license_values()
        schema['title'] = 'License'
        return schema
```

Though I have yet to test this approach fully, it looks viable to me but very messy and opaque. However, the more crucial question for me is if we should redefine `LicenseType` as a custom type in order achieve the goal of this PR, or should we just accept the current state of https://github.com/dandi/dandi-schema/pull/294 which has restriction on import order? If we take this approach, and there are other enum classes need to be made dynamic in the further, they will have to be redefined the same way as well. Should we take this approach, @yarikoptic?

candleindark: The other point I want to bring up, after realizing it in handling this `LicenseType` issue, is that in making `models.py` fully dynamic, we have rendered the use of `set_instance_config()` a potential pitfall. Now, one has to be extremely careful with the use of `set_instance_config()`. Using `set_instance_config()` to define the models, one has to ensure that it is always called in a function that will be executed each time that a model entity is evaluated, and any failure in doing so will lead to a `models.py` that is only partially dynamic and holds definitions that are inconsistent with the instant config. Case in point, the error in the current definition of `LicenseType` was overlook.

jjnesbitt: The way that I can think to solve this problem in the framework of this PR is to change the type of `license` to a `str`, and define the following field validator for `license`:

```python
    @field_validator("license")
    @staticmethod
    def check_license(value: str) -> str:
        license_values = [x.value for x in get_instance_config().licenses]
        if value not in license_values:
            raise ValueError(f"License {value} not valid")

        return value
```

and then add this in [`DandiBaseModel.__get_pydantic_json_schema__`](https://github.com/dandi/dandi-schema/blob/b78da9ea2dd41c617b4699ba6f7b34ad9a0aeb6e/dandischema/models.py#L607):

```python
            if prop == "License":
                value["items"]["enum"] = [
                    x.value for x in get_instance_config().licenses
                ]
```

This solution is messy and has its issues, so if you'd rather go with a static import approach, you're free to do so. However, I'll just point out that there's already loads of arbitrary logic in the `__get_pydantic_json_schema__` method, as well as many other places. 

The larger issue is that you're trying to take a 2000 line file filled with pydantic type definitions, and make it configurable with minimal changes. As far as I can tell, this is not something that's really done in pydantic. You'd be better off creating a function that returns the properly configured classes at runtime, but that has it's own issues.

:shrug:

yarikoptic: Wouldn't it also effect meditor so we loose drop down and require users to enter that string?

Indeed the task we are trying to do has big "footprint" but IMHO it is quite easy to achieve with import time customization. After we achieve desired effects of allowing multiple instances and making backend and frontend configured with just a few new settings, we will look into overhauling this setup, likely with a switch to linkml as the source of the model. Potentially, again, just keeping it a singular version customized at import time.

candleindark: I think that solution wouldn't effect the meditor since the JSON schema for the field is customized at `DandiBaseModel.__get_pydantic_json_schema__`. However, this is less desirable to the automated JSON schema generation if we were to define `LicenseType` as an `Enum`.

PR #322, opened 2025-08-22
Provide environment variable aliases for setting instance config

candleindark:
This PR facilitates dandi-archive, a Django app, to use environment variables with more consistent names to set the instance config to address the issue discussed at https://github.com/dandi/dandi-infrastructure/pull/224/files#r2205622059

**Note**: Population of fields in an instance of `dandischema.conf.Config` can no longer be done by field names due to the limitation imposed by the bug documented at https://github.com/pydantic/pydantic-settings/issues/689. However, this shouldn't have much negative effect since population by aliases remains available.

### DIFF
--- a/dandischema/tests/test_conf.py
+++ b/dandischema/tests/test_conf.py
@@ -18,14 +18,17 @@ def test_get_instance_config() -> None:
     ), "`get_instance_config` should return a copy of the instance config"
 
 
-FOO_CONFIG_DICT = {
+_FOO_CONFIG_DICT_BY_FIELD_NAME = {
     "instance_name": "FOO",
     "doi_prefix": "10.1234",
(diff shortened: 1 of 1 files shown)


dandischema/tests/test_conf.py
```diff
@@ -40,7 +43,7 @@ def test_valid_instance_name(self, instance_name: str) -> None:
         """
         from dandischema.conf import Config
 
-        Config(instance_name=instance_name)
+        Config(dandi_instance_name=instance_name)
```
yarikoptic, line 46: so we are loosing convenience of automated `dandi_` `env_prefix` entirely and need to duplicate it every time we would instantiate `Config`?

yarikoptic (approved):

PR #324, opened 2025-09-02
dandi-archive switched to use uv

yarikoptic:
This is for now just a blind attempt to install

### DIFF
--- a/tools/api-with-schema.Dockerfile
+++ b/tools/api-with-schema.Dockerfile
@@ -1,3 +1,3 @@
 FROM dandiarchive/dandiarchive-api
 COPY . /opt/dandischema/
-RUN cd /opt/dandischema && pip install -e .[dev]
+RUN cd /opt/dandischema && uv pip install -e .[dev]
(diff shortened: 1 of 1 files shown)


yarikoptic:
@jjnesbitt @mvandenburgh with the switch to `uv` of the `dandiarchive/dandiarchive-api` -- what is the best way to build on top of it to install a new library since `pip` is no longer there and `uv pip install` didn't work for us either (including with `--system`).

jjnesbitt:
@yarikoptic I've pushed a commit which seems to fix the issue. I can verify in the built image that I can import `dandischema` with the following command:

```bash
uv run python -c "import dandischema"
```

yarikoptic:
Thank you @jjnesbitt ! So, it finds the `venv` in default path (`/.venv`) and that is why it didn't work whenever I did keep it going away under `/opt` before invoking `uv pip ...`:
```
❯ podman run -it --rm --entrypoint bash dandiarchive/dandiarchive-api -c 'ls -ld $PWD/.venv'
drwxr-xr-x 1 root root 106 Sep  2 15:03 /opt/django/.venv
```

PR #266, opened 2024-11-18
Disallow extra fields other than `"@context"`

candleindark:
This PR closes #75. It addresses #75 in the follow manner.

1. Make the models disallow data instances that have extra fields other than the `"@context"` field in both Pydantic and JSON level.
2. The Pydantic models continue to not have a context field.
3. Data instances with an `"@context"` field can be validated against a Pydantic model but ignored.

<details>
<summary>The solution implemented in this PR is based on the following script.
</summary> 

```py
from typing import Any
import json

from pydantic import BaseModel, ConfigDict, model_validator
from pydantic.json_schema import JsonDict, JsonValue
from jsonschema import validate, Draft202012Validator
import jsonschema


def get_dict_without_context(d: Any) -> Any:
    """
    If a given object is a dictionary, return a copy of it without the
    `@context` key. Otherwise, return the input object as is.

    :param d: The given object
    :return: If the object is a dictionary, a copy of it without the `@context` key;
             otherwise, the input object as is.
    """
    if isinstance(d, dict):
        return {k: v for k, v in d.items() if k != "@context"}
    return d


def add_context(json_schema: JsonDict) -> None:
    """
    Add the `@context` key to the given JSON schema

    :param json_schema: The dictionary representing the JSON schema

    raises: ValueError if the `@context` key is already present in the given dictionary
    """
    context_key = "@context"
    context_key_title = "@Context"
    properties: JsonDict = json_schema.get("properties", {})
    required: list[JsonValue] = json_schema.get("required", [])

    if context_key in properties or context_key in required:
        msg = f"The '{context_key}' key is already present in the given JSON schema."
        raise ValueError(msg)

    properties[context_key] = {
        "format": "uri",
        "minLength": 1,
        "title": context_key_title,
        "type": "string",
    }
    # required.append(context_key)  # Uncomment this line to make `@context` required

    # Update the schema
    # This is needed to handle the case in which the keys are newly created
    json_schema["properties"] = properties
    json_schema["required"] = required


class Foo(BaseModel):
    x: int

    # Model validator to remove the `"@context"` key from data instance before
    # "base" validation is performed.
    _remove_context_key = model_validator(mode="before")(get_dict_without_context)

    model_config = ConfigDict(extra="forbid", json_schema_extra=add_context)


json_schema_ = Foo.model_json_schema()
print(json.dumps(json_schema_, indent=2))
"""
{
  "additionalProperties": false,
  "properties": {
    "x": {
      "title": "X",
      "type": "integer"
    },
[...]

PR #342, opened 2025-10-27
Add ability to downgrade schema all the way to 0.6.10

yarikoptic:
Reminder to myself: we run checks on no changes to schema only on PRs with "release" tag. Thus we did not run checks when we actually adjusted the schema in 
- https://github.com/dandi/dandi-schema/pull/185 by @bendichter.  

Then I rushed merging https://github.com/dandi/dandi-schema/pull/341
as I was not expecting surprises from adjusting labels (forgot about above).
The moral again, do not rush merging - let CI finish! ;)

I think the motivation/aspects are best captured in
- https://github.com/dandi/dandi-schema/issues/343
but overall -- to facilitate upload to "older" server which might not be yet aware of a new(er) schema.

~~Boosting "patch" for both schema and library since just a "minor addition of a new feature" (mostly to schema).~~

TODOs
- [x] release dandi-cli with https://github.com/dandi/dandi-cli/pull/1732
- [x] lower boost here into .patch which would also trigger retesting, 
- [x] full CI, dandi-cli included, must come green

### DIFF
--- a/dandischema/consts.py
+++ b/dandischema/consts.py
@@ -1,6 +1,6 @@
 from packaging.version import Version as _Version
 
-DANDI_SCHEMA_VERSION = "0.8.0"
+DANDI_SCHEMA_VERSION = "0.7.1"
 ALLOWED_INPUT_SCHEMAS = [
     "0.4.4",
     "0.5.1",
--- a/dandischema/consts.py
+++ b/dandischema/consts.py
@@ -22,7 +22,7 @@ ALLOWED_INPUT_SCHEMAS = [
 
 # We establish migrations (back) to only a few recent versions.
 # When adding changes, please consider whether a migration path should be added.
-ALLOWED_TARGET_SCHEMAS = ["0.6.10", DANDI_SCHEMA_VERSION]
+ALLOWED_TARGET_SCHEMAS = ["0.6.10", "0.7.0", DANDI_SCHEMA_VERSION]
 
 # This allows multiple schemas for validation, whereas target schemas focus on
 # migration.
--- a/dandischema/consts.py
+++ b/dandischema/consts.py
@@ -1,6 +1,6 @@
 from packaging.version import Version as _Version
 
-DANDI_SCHEMA_VERSION = "0.7.1"
+DANDI_SCHEMA_VERSION = "0.8.0"
 ALLOWED_INPUT_SCHEMAS = [
     "0.4.4",
     "0.5.1",
--- a/dandischema/consts.py
+++ b/dandischema/consts.py
@@ -20,10 +20,9 @@ ALLOWED_INPUT_SCHEMAS = [
     DANDI_SCHEMA_VERSION,
 ]
 
-# ATM we allow only for a single target version which is current
-# migrate has a guard now for this since it cannot migrate to anything but current
-# version
-ALLOWED_TARGET_SCHEMAS = [DANDI_SCHEMA_VERSION]
+# We establish migrations (back) to only a few recent versions.
+# When adding changes, please consider whether a migration path should be added.
+ALLOWED_TARGET_SCHEMAS = ["0.6.10", DANDI_SCHEMA_VERSION]
 
 # This allows multiple schemas for validation, whereas target schemas focus on
 # migration.
--- a/dandischema/metadata.py
+++ b/dandischema/metadata.py
@@ -414,20 +414,28 @@ def migrate(
     # Downgrades
 
     # Simple downgrades that just require removing fields, which is totally fine
-    # if they are empty
+    # if they are empty, as they are None or empty containers (list, tuple, etc).
+    # List only those for which such notion of "empty" applies.
     SIMPLE_DOWNGRADES = [
         # version added, fields to remove
         ("0.6.11", ["releaseNotes"]),
     ]
     for ver_added, fields in SIMPLE_DOWNGRADES:
         # additional guards are via ALLOWED_TARGET_SCHEMAS
-        if (to_version_tuple < version2tuple(ver_added) <= obj_version_tuple):
+        if to_version_tuple < version2tuple(ver_added) <= obj_version_tuple:
             for field in fields:
                 if field in obj_migrated:
-                    if val := obj_migrated.get(field):
-                        raise ValueError(f"Cannot downgrade to {to_version} from "
-                                         f"{obj_version} with {field}={val!r} present")
-                    del obj_migrated[field]
+                    value = obj_migrated.get(field)
+                    # Explicit check for "empty" value per above description.
+                    if value is None or (
+                        not value and isinstance(value, (list, tuple, dict, set))
+                    ):
+                        del obj_migrated[field]
+                    else:
+                        raise ValueError(
+                            f"Cannot downgrade to {to_version} from "
+                            f"{obj_version} with {field}={value!r} present"
+                        )
 
     # Always update schemaVersion when migrating
     obj_migrated["schemaVersion"] = to_version
--- a/dandischema/metadata.py
+++ b/dandischema/metadata.py
@@ -420,7 +420,7 @@ def migrate(
     SIMPLE_DOWNGRADES = [
         # version added, fields to remove
         ("0.7.0", ["releaseNotes"]),
-        ("0.8.0", ["sameAs"]),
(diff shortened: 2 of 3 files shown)


waxlamp:
> @waxlamp here we are to get into that unpleasant situation to need to coordinate releasing since otherwise if client has this release more recent dandi-schema installed (possible with or without us pinning in client): we would get
> 
> ```
> ERROR tests/test_upload.py::test_upload_sync[False-False] - dandi.exceptions.SchemaVersionError: Server requires schema version 0.6.10; client only supports 0.7.0.  You may need to upgrade dandi and/or dandischema.
> ```
> 

How could this happen if you are pinning `dandischema` in the client (other than a user going out of their way to forcibly upgrade `dandischema` in their virtual environment after installing `dandi-cli`)?

(For the record, it seems that the CLI does indeed [pin `dandischema` to a minor version](https://github.com/dandi/dandi-cli/blob/a79fffc8e8cc32637cf592a8c0bdd809eacd64e9/pyproject.toml#L52).)

> so the question(s)
> 
> * do you see any problem with us aiming to update versioned depends/release dandi-archive ATM?

No problem, given that both the archive and the CLI are pinning to a specific version of `dandischema` already.

This does mean we need to coordinate releasing versions of both that do use the new version, but I'm not sure how we can avoid that without solving the bigger problem of enabling automatic migration between schema versions.

> * should we provide PR or you will?

I imagine it would be part of https://github.com/dandi/dandi-archive/pull/2606 (see [this comment](https://github.com/dandi/dandi-archive/pull/2606/commits/5017f27194849516ce2f7028a4996303bb00d19a#r2469728935)).

yarikoptic:
> > @waxlamp here we are to get into that unpleasant situation to need to coordinate releasing since otherwise if client has this release more recent dandi-schema installed (possible with or without us pinning in client): we would get
> > ```
> > ERROR tests/test_upload.py::test_upload_sync[False-False] - dandi.exceptions.SchemaVersionError: Server requires schema version 0.6.10; client only supports 0.7.0.  You may need to upgrade dandi and/or dandischema.
> > ```
> 
> How could this happen if you are pinning `dandischema` in the client (other than a user going out of their way to forcibly upgrade `dandischema` in their virtual environment after installing `dandi-cli`)?

Simply because we are testing against these changes, which will give new version 0.7.0 -- that's the whole point. And it just reveals fragility of our tight binding of versions here and there and that pinning just a workaround for "user deployment" while developers, who need to test components where one is ahead of release, would get out of sync!  What is critical here is that actually this new version would be just fine for dandi-cli...  so I might reconsider and

- relax in dandi-cli to require the same MAJOR.MINOR only, not MAJOR.MINOR.PATCH (potentially with some hardcoded minimal we need since we were not consistent all the time before in 0.6 series)
- make this to release 0.6.11 not 0.7.0 schema, so we absorb within MAJOR.MINOR even backward compatible enhancements . If we need server to demand newer client (e.g. we know client extracts new important metadata), we can demand via `/server-info` but IMHO there is no way to demand matching schema version

then this would 
- pass testing as long as we do not introduce backward compatible changes! and that is where cli might need to account for them

> This does mean we need to coordinate releasing versions of both that do use the new version, but I'm not sure how we can avoid that without solving the bigger problem of enabling automatic migration between schema versions.

well, I think there is overall misunderstanding as we actually/apparently do not auto migrate metadata at all ATM!  Server does not really care much, and client started to demand exact match to avoid missing recently added new metadata to be extracted/uploaded.  Since here we talk about releaseNotes which client is not concerned about, it should not be concerned about being 'outdated' from client's perspective, so no boost of `minimal client version` is needed.

> > * should we provide PR or you will?
> 
> I imagine it would be part of [dandi/dandi-archive#2606](https://github.com/dandi/dandi-archive/pull/2606) (see [this comment](https://github.com/dandi/dandi-archive/pull/2606/commits/5017f27194849516ce2f7028a4996303bb00d19a#r2469728935)).

thanks for the pointers!  Indeed, would make sense to boost in that PR to this new version simply to allow for this new feature (since you hard-pin).

dandischema/consts.py
```diff
@@ -1,6 +1,6 @@
 from packaging.version import Version as _Version
 
-DANDI_SCHEMA_VERSION = "0.6.10"
+DANDI_SCHEMA_VERSION = "0.6.11"
```
candleindark: The patch is boosted instead of minor here which contradicts with the top post, https://github.com/dandi/dandi-schema/pull/342#issue-3558387078, which indicates booting of minor.  However, "minor" can possible mean patch there.

candleindark (commented):
Left a comment. Will provide more.

dandischema/metadata.py
```diff
@@ -424,22 +418,17 @@ def migrate(
...
-    # Ensure the target DANDI schema version is at least the DANDI schema version
-    # of the provided instance
-    if obj_ver_tuple > target_ver_tuple:
-        raise ValueError(f"Cannot migrate from {obj_ver} to lower {to_version}.")
-
     # Optionally validate the instance against the DANDI schema it specifies
     # before migration
     if not skip_validation:
-        _validate_obj_json(obj, _get_jsonschema_validator(obj_ver, "Dandiset"))
+        _validate_obj_json(obj, _get_jsonschema_validator(obj_version, "Dandiset"))
```
candleindark: This migration function seems to expect the metadata instance to be a `Dandiset` instance.

candleindark: Validating `PublishedDandiset` instance may create a problem if https://github.com/dandi/dandi-schema/pull/266 is put in. However, I guess we can have it specifically validate against the `PublishDandiset` schema if a `PublishedDandiset` instance is provided.

candleindark (commented):
Left a few more comments.

dandischema/metadata.py
```diff
@@ -358,24 +358,18 @@ def migrate(
...
     # --------------------------------------------------------------
     # Validate DANDI schema version provided in the metadata instance
     # --------------------------------------------------------------
     # DANDI schema version of the provided instance
-    obj_ver = obj.get("schemaVersion")
-    if obj_ver is None:
+    obj_version = obj.get("schemaVersion")
+    if obj_version is None:
         msg = (
-            "The provided Dandiset metadata instance does not have a "
```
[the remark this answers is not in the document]
yarikoptic: Because this is not model specific per se AFAIK, as it could take any of them

dandischema/metadata.py
```diff
@@ -458,7 +447,28 @@ def migrate(
...
+    SIMPLE_DOWNGRADES = [
+        # version added, fields to remove
+        ("0.6.11", ["releaseNotes"]),
+    ]
+    for ver_added, fields in SIMPLE_DOWNGRADES:
+        # additional guards are via ALLOWED_TARGET_SCHEMAS
+        if to_version_tuple < version2tuple(ver_added) <= obj_version_tuple:
+            for field in fields:
+                if field in obj_migrated:
+                    if val := obj_migrated.get(field):
```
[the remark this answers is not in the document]
yarikoptic: good point! I was aiming for an "empty value" check but indeed what is needed is to check if it matches the default... or may be more relaxed would be just to check if field there at all, and hence remove this check, but then it would become dependent on it being optional and people do not fill up optional field.

Overall, given current distraction at sfn 2025, I will postpone the "downgrade" portion of this PR. Since you provided good comments, I will craft a separate short one for releaseNotes so we could proceed with merging both that + devendorization in one go. I will deal with downgrades later!

yarikoptic:
rright, not ready to release 
>  ERROR tests/test_upload.py::test_upload_nonzarr_to_zarr_path - dandi.exceptions.SchemaVersionError: Server uses older incompatible schema version 0.7.0; client supports 0.8.0.

as client then should introduce use of this downgrading functionality...  I guess we might need to go for a patch jump in schema version which should make it "legit" in the eye of the client which has a more relaxed logic now.

yarikoptic:
It was all good besides the tests of dandi-cli which we excercised in the 

- https://github.com/dandi/dandi-cli/pull/1898

which is instrumented to use this migration functionality.  So I rebased and if we have it all still 'kosher' I will merge/release new dandi-schema library (minor release since IMHO this is an important addition).  I will instruct that newer dandi-cli to skip prior release since it had no migration support, and we will prep dandi-cli for release

PR #344, opened 2025-11-17
Release new schema 0.7.0 for addition of the releaseNotes

yarikoptic:
This is a cut version of

- #342

which added a wishful component for allowing downgrading model versions.  That would take more work, so will just progress now with releasing both new added releaseNotes and devendorization (https://github.com/dandi/dandi-schema/pull/294), hence "minor" component boost for both schema and library

note: `dandi-cli` tests would fail since dandi-archive component would not be ready to publish with a new version of the schema ATM.

### DIFF
--- a/dandischema/consts.py
+++ b/dandischema/consts.py
@@ -1,6 +1,6 @@
 from packaging.version import Version as _Version
 
-DANDI_SCHEMA_VERSION = "0.6.10"
+DANDI_SCHEMA_VERSION = "0.7.0"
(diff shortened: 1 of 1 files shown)


yarikoptic:
dandi-cli is to fail since server can't validate new version of the schema as it doesn't know it.  But otherwise I think it is ready and we could just release if dandi-archive is ready to accept devendorized schema version. (attn @dandi/archive-maintainers). @candleindark last look?

PR #348, opened 2025-11-20
Revert change to delegate building package and upload it to PyPi using `auto` hook but add "released" `auto` plugin

candleindark:
The title of this PR provides sufficient description already.

If this PR is to be targeted to the master branch instead, make sure that it has the `release` label and an label that bumps the version.


yarikoptic:
I meant to just make a separate PR for a release not PR to pr

PR #374, opened 2026-03-02
maint: drop Python 3.9 support, require Python >=3.10

candleindark:
## Summary

- Remove Python 3.9 from the test matrix in all CI workflows
- Update `python_requires = >=3.10` in `setup.cfg`
- Drop the `Programming Language :: Python :: 3.9` classifier

Python 3.9 reached end-of-life in October 2024. Additionally,
`black>=26.x` (pinned in `.pre-commit-config.yaml`) requires Python>=3.10, making 3.9 incompatible with the current pre-commit setup.

## Test plan

- [x] All CI jobs pass with Python 3.10 as the new minimum
- [x] Confirm `black` pre-commit hook installs and runs correctly in the updated environment

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -16,7 +16,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v6
       with:
-        python-version: '3.9'
+        python-version: '3.10'
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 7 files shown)


yarikoptic:
may be @CodyCBakerPhD would remember when he comes back on where it is required -- I did not see any branch protection rules here, just ignore it

yarikoptic (approved):

PR #381, opened 2026-03-10
WiP: Branch with auto converted linkml model

yarikoptic:
### To preview and test results, actual changes should go to

- #387

PR #387, opened 2026-03-20
Linkml conversion tooling

yarikoptic:
This is an extract with amends from
- #381

which (branch [`linkml-auto-converted`](TODO)) would keep merging this branch into itself while reflecting on changes in the branch which could be rebased or gain merges from the master, and also can accumulate or drop "patch branches" from within its script defining what to patch with.

This way `linkml-auto-converted` would represent reflection of current state of conversion

TODO/PLAN
- [x] Establish branch [linkml-auto-converted]() -- that one in #381
- [x] Made ‘hatch’ script (you could add pydantic2linkml as dependency there) to convert [orig_models.py]() into dandischema/models.yaml : `hatch ... TODO`
- [x] Translated the original `models.py` into [`dandischema/models.yaml`]() and overlaid with an [dandischema/models_overlay.yaml] overlay file.
- [x] script `tools/linkml_conversion` to convert into ‘linkml-auto-converted’
- [ ] define `model_instances.yaml` (or alike) which would define pre-populated records such as standards (bids, nwb, ...). aim for potentially multiple classes there.
- [ ] add a github workflow here which would react to changes into 'master' and this branch and with manual dispatch, which would first merge master into this branch, then run the script, and push results to linkml-auto-converted branch. This way we would always have 'up to date' and automatically updated state of that branch.
- [ ] address "notes" about failed conversions one way (changing current dandi-schema pydantic model) or another (pydantic2linkml) or !
  - [ ] we can add a custom script to "enhance" auto generate linkml model to address any changes needed programmatically! 
  - [ ] we can have a branch (or just a .patch file) with changes to perform **on top** of converted linkml
- [ ] ... 
- [ ] There you produce pydantic model out of this patched model sufficient (although potentially more relaxed) to replace current pydantic model.


pyproject.toml
```diff
@@ -62,6 +62,19 @@ all = [
...
+#   `hatch run linkml-auto-converted:2linkml`    # Convert dandischema.models -> dandischema/models.yaml
+#   `hatch run linkml-auto-converted:2pydantic`  # Convert dandischema/models.yaml -> dandischema/models_linkml.py
+[tool.hatch.envs.linkml-auto-converted]
+dependencies = [
+    "black",  # This allows `gen-pydantic` to format the generated Pydantic models with Black formatting
+    "pydantic2linkml @ git+https://github.com/dandi/pydantic2linkml.git"
+]
+[tool.hatch.envs.linkml-auto-converted.scripts]
+2linkml = "pydantic2linkml -l INFO -M dandischema/models_merge.yaml -O dandischema/models_overlay.yaml dandischema.models | sed -e 's,0x[0-9a-ef]*,0xADDRESS,g' -E -e 's,([^[:space:]]+):[0-9]+,\\1:NUMBER,g' > dandischema/models.yaml"
+2pydantic = "gen-pydantic --black dandischema/models.yaml | sed -e 's,dciteCOLON,,g' > dandischema/models_linkml.py"
```
yarikoptic: so the idea to replace `dandiCOLON` etc as well

```suggestion
2pydantic = "gen-pydantic --black dandischema/models.yaml | sed -e 's,[a-z]\+COLON,,g' > dandischema/models_linkml.py"
```

candleindark: done but with -E for extended regex (with plain +) for portability.

tools/linkml_conversion
```diff
@@ -0,0 +1,100 @@
...
+    if [ "$status" -ne 0 ]; then
+        if grep -qi 'failed' <<<"$out"; then
+            echo "patch FAILED for branch $b — see rejects" >&2
+            exit 1
+        fi
+        if ! grep -qiE 'previously applied|reversed|skipping patch' <<<"$out"; then
+            echo "patch exited $status for branch $b without a recognized skip indicator; aborting" >&2
+            exit 1
+        fi
+    fi
```
yarikoptic, line 73:
```suggestion
    if [ "$status" -ne 0 ]; then
        if grep -qiE 'previously applied|reversed|skipping patch' <<<"$out"; then
            echo "patch exited $status for branch $b but it was about applied patch, ignoring"
        else
            echo "patch FAILED for branch $b with exit $status — see rejects" >&2
            exit $status
        fi
    fi
```

PR #398, opened 2026-04-15
Include contributors with includeInCitation=True as DataCite creators

yarikoptic:
Previously, only contributors with dcite:Author role were placed into DataCite creators. Organizations (e.g., "Allen Institute for Brain Science" in dandiset 000020) that had includeInCitation=True but no dcite:Author role ended up only as DataCite contributors with contributorType "Other" -- despite appearing in the citation string.

Now any contributor with includeInCitation=True OR dcite:Author role becomes a DataCite creator.

Tests: folded new includeInCitation test cases into the existing test_dandimeta_datacite parametric test. Removed @skipif_no_doi_prefix and @skipif_no_datacite_auth decorators so that local validation (schema + field checks) always runs; posting to DataCite is now gated at runtime on credential availability. This required adapting the publisher field check to be instance-config-aware (via _EXPECTED_PUBLISHER) since the RRID fields are only present for the DANDI instance, not DANDI-ADHOC.

This is initially triggered by curation work on https://github.com/bids-dandisets/curations/ and running into empty Authors within https://github.com/bids-dandisets/000020/blob/curation/dataset_description.json#L7 .

Related issues/PRs now
- https://github.com/bids-standard/bids-specification/pull/2397
- https://github.com/con/nwb2bids/issues/385

### DIFF
--- a/dandischema/datacite/__init__.py
+++ b/dandischema/datacite/__init__.py
@@ -227,11 +227,14 @@ def to_datacite(
         elif isinstance(contr_el, Organization):
             contr_dict["nameType"] = "Organizational"
 
-        if contr_el.roleName and RoleType("dcite:Author") in contr_el.roleName:
+        is_author = contr_el.roleName and RoleType("dcite:Author") in contr_el.roleName
+        if is_author:
+            contr_el.roleName.remove(RoleType("dcite:Author"))
+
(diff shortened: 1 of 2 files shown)


candleindark:
Some tests are still failing, particularly ones that have vendor specific information.

Issue #292, opened 2025-03-06
Expand metadata model (summaryStatistics?) and provide more metadata for datacite


yarikoptic:
@satra pointed in slack

> https://metadata-health-ui.vercel.app/providers/dartlib?view=summary&client=dartlib.dandi (
[@Yaroslav (AKA Yarik)](https://dandiarchive.slack.com/team/W0120Q1E26T)
 and 
[@dorota](https://dandiarchive.slack.com/team/W012FMSEF7V)
 - as you rethink the doi generation, this report may be helpful in thinking about updating the datacite metadata for recommended and optional mapping).

<details>
<summary>Currently  Mandatory: 100%, Recommended: 48%, Optional: 29%</summary> 

![Image](https://github.com/user-attachments/assets/552f5c08-3f54-4ed1-959e-75e1bfb9689c)

</details>

satra:
RRID has been introduced to `relatedIdentifierType` in datacite schema 4.6

yarikoptic:
interesting! related brief issue was https://github.com/datacite/schema/issues/150

Issue #389, opened 2026-03-27
Handle LinkML migration issue of `pydantic2linkml: Impossible to generate slot usage entry for the`


candleindark:
pydantic2link has the complains about not able to generate some slot usage entries  in generation of some class definitions

such as in the translation of the `Dandiset` class

<img width="988" height="102" alt="Image" src="https://github.com/user-attachments/assets/cb532142-fa6e-4068-9b49-c52875fd0a41" />

## Stats of such complains

Total: 54 count
Ones about the `schemaKey` usage slots at different class defintions: 41 counts

## Cause
pydantic2link unable to generate the usage slot cause by the fact that LinkML schema are [monotonic](https://linkml.io/linkml/schemas/slots.html#slot-usage). Restrictions from a slot usage definition are only additive and can not override the restrictions in the corresponding slot definition.

## TODOs:
- [ ] Come up with a custom patch to each of these complains in the generated LinkML schema (the patching can be done with the `--merge-file` option in pydantic2linkml
    - [ ] The ones about  `schemaKey` should be able to be handled the same way (do these first)
        - [x] make the `schemaKey` slot be a type designator (using the _merge.yaml)
        - [x] remove the `ifabsent` value in the  `schemaKey` slot (this value is not relevant in a type designator; using the he _merge.yaml))
        - [x] remove all the existing notes of `Impossible to generate slot usage entry for the SchemaKey` appended by pydantic2linkml
        - [x] make `schemaKey` required  (using the `_merge.yaml`). Doing this ensure new data coming in have the `schemaKey` field.
        -  [x] note: if `_merge.yaml` was merged "first" while processing subclasses, and if default value differs from the class name - we wanted to add `notes:` about that divergence
            - [x] workaround: add manually to `_merge.yaml` file those notes for all those classes where we overload `schemaKey` with the value that is not the same as class name  
        - [ ] Figure out **why** or **what for** we had schemaKey 'Dandiset' for `PublishedDandiset` and so on
            - [ ] check what class loading of a published dandiset manifest record results in (I guess should be `PublishedDandiset`)
            - [ ] check with @dandi/dandi-archive people how schemaKey is used in frontend (e.g. meditor) and whether it must remain e.g. "Dandiset" for "PublishedDandiset"

candleindark:
**Note:** Inline definition of a slot via the `attributes` meta slot does
not change its behavior under inheritance. All slots are inherited the
same way regardless of whether they are defined inline within a class
[...]

PR #401, opened 2026-04-27
Add concept DOI model, dates field, IsVersionOf relations, DANDI identifier

satra:
## Summary

- Add optional `doi` field to `Dandiset` model for concept DOI support (version-independent DOI per dandiset), with `DANDI_CONCEPT_DOI_PATTERN` using dynamic `Config.doi_prefix` to preserve vendorization
- Add `dates` field to `to_datacite()` output mapping `datePublished` to DataCite `dateType: "Issued"`, bringing recommended field coverage from 4/6 to 5/6 (83%)
- Add `concept_doi` parameter to `to_datacite()` that emits an `IsVersionOf` `relatedIdentifier` when provided (backward compatible, defaults to `None`)
- Add DANDI identifier (e.g., `DANDI:000485`) to DataCite `alternateIdentifiers` with type `"DANDI"`

All changes are backward compatible. Existing callers of `to_datacite()` continue to work without modification.

**Related issues**: dandi/dandi-archive#1319, dandi/dandi-schema#259, dandi/dandi-schema#292

## Test plan

- [x] 9 new TDD tests in `test_doi_improvements.py` covering all 4 changes
- [x] Updated `test_datacite_publish` expected output for new fields
- [x] 105 total tests passing (including 79 model tests), 0 failures
- [x] Verified no regressions with `DANDI_DOI_PREFIX=10.80507 DANDI_INSTANCE_NAME=DANDI DANDI_INSTANCE_IDENTIFIER="RRID:SCR_017571"`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


satra:
@yarikoptic - you can review this and we will likely want to release this before the PRs (to archive) will be merged for doi process (coming soon).

dandischema/datacite/tests/test_datacite.py
```diff
@@ -603,8 +603,13 @@ def test_datacite_publish(metadata_basic: Dict[str, Any]) -> None:
                         ),
                         "alternateIdentifierType": "URL",
                     },
+                    {
+                        "alternateIdentifier": dandi_id,
+                        "alternateIdentifierType": _INSTANCE_CONFIG.instance_name,
```
yarikoptic: what will be exact value here? Are we "reserving" now the "DANDI" and "EMBER-DANDI" etc as "alternateIdentifierType"?

note that we already have (above) 

```
                    {
                        "alternateIdentifier": f"https://identifiers.org/{dandi_id}/{version}",
                        "alternateIdentifierType": "URL",
                    },
```

so for concept could simply be

```
                    {
                        "alternateIdentifier": f"https://identifiers.org/{dandi_id}",
                        "alternateIdentifierType": "URL",
                    },
```
I think and be consistent with already existing support

dandischema/datacite/__init__.py
```diff
@@ -138,6 +139,10 @@ def to_datacite(
             "alternateIdentifier": str(meta.url),
             "alternateIdentifierType": "URL",
         },
+        {
+            "alternateIdentifier": meta.identifier,
+            "alternateIdentifierType": instance_config.instance_name,
```
yarikoptic: I do not see how this fits here well...

uv.lock
```diff
@@ -0,0 +1,1400 @@
...
+version = "0.4.7"
+source = { registry = "https://pypi.org/simple" }
+dependencies = [
+    { name = "click" },
+    { name = "tqdm" },
+]
+sdist = { url = "https://files.pythonhosted.org/packages/b4/af/eb603e6d7f02d16a60b05335ed97a08de4c71f5b109c2a65f3a571d2c4be/zarr_checksum-0.4.7.tar.gz", hash = "sha256:081f163d52533a286bc8249a241da5614b5dab0c06377b208980dc205e272da9", size = 12414, upload-time = "2025-04-01T15:54:48.548Z" }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/e3/6b/a57376f2713b79e5f2eb8003ebc569aa3201a36779ee1472404f183e3c87/zarr_checksum-0.4.7-py3-none-any.whl", hash = "sha256:97cbffe2288338f9083f92a2312052f045fb2154a6f4ddbbcb1bde109b6fd22e", size = 15292, upload-time = "2025-04-01T15:54:47.27Z" },
+]
```
yarikoptic: let's not introduce those (for now at least) -- we do not have any canonical deployment of dandischema, so I do not want to taint any use of `uv` to be limited to some specific deployment scenario

yarikoptic (changes requested):
This is an alternative (if merged, should close)
- #297 

hence invited @asmacdo to review since we might have some memories on that effort back in his mind.

It is not clear to me on the value for the type in the otherIdentifiers and why not to use already reserved one based on identifiers.org URL type?

- remove uv.lock
- annotate AI produced tests with a dedicated mark (I am yet to get juice to review and potentially ask to condense or parametrize(

dandischema/datacite/tests/test_datacite.py
satra: Resolved in cc07c42 — the instance-name `alternateIdentifier` was removed entirely. Only the two `URL`-typed entries remain (`https://identifiers.org/{dandi_id}/{version}` and `meta.url`), so search/lookup by DANDI ID via identifiers.org continues to resolve. `alternateIdentifierType` is a free-form string in DataCite (no controlled vocabulary), so emitting `"DANDI"`/`"EMBER-DANDI"` would have been a custom convention with no lookup story. If a structured DANDI ID is later useful, the right path is a typed `relatedIdentifier` rather than re-introducing this entry.

dandischema/datacite/__init__.py
satra: Same — addressed in cc07c42 by removing the instance-name alternateIdentifier from the DataCite emit. The remaining `alternateIdentifiers` are only the two `URL`-typed entries that pre-dated this PR.

uv.lock
satra: Apologies — cc07c42's commit message claimed `uv.lock` was removed but the file was still tracked. Properly removed now in 2077205 and added to `.gitignore`.

satra:
**Update — addressing review (commit 2077205):**

- ✅ `uv.lock` properly removed and added to `.gitignore` (cc07c42's commit message claimed removal but the file was still tracked — apologies for the confusion).
- ✅ Instance-name `alternateIdentifier` removed entirely (cc07c42); only the two URL-typed entries remain.
- ✅ Bumped `DANDI_SCHEMA_VERSION` 0.7.0 → 0.7.1 — the optional `doi` field on `Dandiset` and the changed DataCite output shape (added `dates`, removed instance-name `alternateIdentifier`) is a schema change. Without a bump, downstream validators caching by version wouldn't notice the new optional field.
- ✅ Added `test_dandiset_doi_rejects_malformed` (5 parametrized cases) to exercise the new `DANDI_CONCEPT_DOI_PATTERN`. Closes the Codecov gap on `models.py:1710`.

@yarikoptic @asmacdo — ready for re-review.

dandischema/datacite/tests/test_doi_improvements.py
asmacdo, line 1: `test_doi_improvenents.py` wont make sense after the merge. Instead I think it makes more sense to incorporate this into `test_datacite`, or split into a more expressive file name.

dandischema/datacite/__init__.py
```diff
@@ -163,6 +164,10 @@ def to_datacite(
         )
 
     attributes["publicationYear"] = str(meta.datePublished.year)
+    # T002: Add dates field with Issued dateType
```
asmacdo, line 167: If this is done with spec-kit, it might be worth reviewing the spec itself.

yarikoptic: +1 -- we would need the "sources"! ;)

dandischema/models.py
```diff
@@ -1700,6 +1707,14 @@ def contributor_musthave_contact(
         ),
     ]
 
+    doi: Optional[str] = Field(
+        default=None,
+        title="Concept DOI",
+        description="A version-independent DOI for the Dandiset as a whole.",
+        pattern=DANDI_CONCEPT_DOI_PATTERN,
```
yarikoptic, line 1714: @candleindark I believe this would be one more case where subclass would overload the constraint, and thus we would have difficulty expressing in linkml right?

I feel that in such cases we just need to define slot without any constraint and in specific class uses of the slots define their specific constraints thus making it all work without warning... can we do that?

NB this is just a note/question to @candleindark , not request to change anything here

dandischema/datacite/tests/test_doi_improvements.py
```diff
@@ -0,0 +1,253 @@
...
+
+from dandischema.conf import get_instance_config
+from dandischema.models import (
+    Dandiset,
+    LicenseType,
+    RoleType,
+)
+from dandischema.tests.utils import (
+    DOI_PREFIX,
+    INSTANCE_NAME,
```
asmacdo, line 25: IIRC, settting the instance name is one possible strategy for "safe testing" against a real datacite instance, ie instead of `DANDI`, we might use `DANDI_TEST_2026_05_07` 

So if we set up some e2e test where real findable DOIs get created, they would be namespaced and not interfere with future tests.

dandischema/models.py
candleindark: > [@candleindark](https://github.com/candleindark) I believe this would be one more case where subclass would overload the constraint, and thus we would have difficulty expressing in linkml right?

Yes, but not exactly right here though this will cause it. With this, we will have difficult of translating the `doi` `slot_usage` entry in `PublishedDandiset`. This is because LinkML has monotonic behavior per https://linkml.io/linkml/schemas/advanced.html#unions-as-ranges, and `pattern` is a constraint metadata slot in LinkML to which the monotonic behavior applies to. (However, we can actually do it currently because of a [bug](https://github.com/linkml/linkml/issues/1508#issuecomment-2292431671), but I wouldn't count on it.)

candleindark: > I feel that in such cases we just need to define slot without any constraint and in specific class uses of the slots define their specific constraints thus making it all work without warning... can we do that?

No. In the schema level slot definitions in our LinkML schema of dandischema, a definitions already contains just the ~~minimum~~ maximum set of properties shared by all uses of the slot. The problem originates from inheritance. Since `PublishedDandiset` inherits from `Dandiset`. The `doi` slot in  `PublishedDandiset` has all the properties of the `doi` slot in `Dandiset`, and overriding the value of `pattern` in the `doi` slot in `PublishedDandiset` is not possible per the monotonic behavior.

candleindark: > NB this is just a note/question to [@candleindark](https://github.com/candleindark) , not request to change anything here

I agree. This is problem adding to a category that we already have. I will just have to find a way to solve the problem category not any particular problem.

yarikoptic: > The `doi` slot in `PublishedDandiset` has all the properties of the `doi` slot in `Dandiset`, and overriding the value of `pattern` in the `doi` slot in `PublishedDandiset` is not possible per the monotonic behavior.

that's what I am saying -- it will not override but only add it where actually used, i.e. not defined in the `slot` definition but only in the classes which would use that slot.

anyways -- remind where the issue on this is, we could continue there

candleindark: > that's what I am saying -- it will not override but only add it where actually used, i.e. not defined in the `slot` definition but only in the classes which would use that slot.

I don't think that will work conceptually. Where a slot is defined doesn't change the slot inheritance behavior in LinkML. A more detailed post is in the issue you are looking for https://github.com/dandi/dandi-schema/issues/389#issuecomment-4403504733.

However, overriding still works in practice because of the [bug](https://github.com/linkml/linkml/issues/1508#issuecomment-2292431671).

PR #402, opened 2026-04-28
DOC: Adopt from dandi-cli - rudimentary CLAUDE.md  and pytest_plugin.py (summary of depends, ai_generated marker, better diff for DandiBaseModel)

yarikoptic:
adopted from dandi-cli

@satra rebase your #401 and ask your claude to reflect upon details provided here for tests annotation


CLAUDE.md
```diff
@@ -0,0 +1,44 @@
...
+- DataCite-dependent tests require `DATACITE_DEV_LOGIN` / `DATACITE_DEV_PASSWORD`; set `NO_ET=1` to disable etelemetry. Both are passed through by tox.
+- There is no `docker-compose` in this repo and no `hatch run test:run` env — don't suggest those (the dandi-cli pattern doesn't apply here).
+
+## Pre-commit
+- Install hooks if missing (no `.git/hooks/pre-commit` present): `pre-commit install`.
+- Hooks run: trailing-whitespace, end-of-file-fixer, check-yaml, check-added-large-files, black, isort, codespell, flake8.
+- If a commit fails because hooks auto-fixed files, just re-run `git commit` — the second attempt usually succeeds. Investigate further only if it fails twice.
+
+## Test Markers
+- Mark AI-generated tests with `@pytest.mark.ai_generated`.
```
candleindark, line 33: I made a comment about this in another PR https://github.com/dandi/dandi-cli/pull/1827/changes/BASE..0f8867bd1b98dddd281f9720f3597acc86e2f6f1#r3156553817.

candleindark: If you insist in using this mark, you should [register](https://docs.pytest.org/en/stable/how-to/mark.html#registering-marks) it, preferable in `pyproject.toml`. Otherwise, the use of the marker will trigger a warning, which will turn into an error according our setup in `tox.ini`.

CLAUDE.md
```diff
@@ -0,0 +1,44 @@
...
+- Class names: CamelCase; functions/variables: snake_case.
+- Exception names end with `Error` (see `dandischema/exceptions.py`).
+- Docstrings in NumPy style for public APIs; keep them in sync with signature changes.
+- Prefer specific exceptions over generic ones.
+- Imports organized stdlib / third-party / local, alphabetical within groups (isort handles ordering).
+- All imports at module top — no function-local imports (per global rule), unless guarding an optional dep with `try/except ImportError`.
+
+## Schema / Data Files
+- `dandischema/*.yaml` (e.g. `standards.yaml`, `model_standards.yaml`, `personinfo.yaml`) are tracked schema/data inputs — don't reformat or rename casually; downstream consumers and `tools/pubschemata.py` rely on them.
+- `dandischema/_version.py` is generated by `versioningit` (see `[tool.versioningit.write]`); don't edit by hand.
```
candleindark, line 48: Suggestion: add a "Documentation Freshness" section at the end of `CLAUDE.md`. Accept this suggestion on the last line (line 44, the `_version.py` bullet) — it preserves that line and appends the new section after it.

```suggestion
- `dandischema/_version.py` is generated by `versioningit` (see `[tool.versioningit.write]`); don't edit by hand.

## Documentation Freshness
- When preparing a PR, check whether the documentation — both human-facing (e.g. `README.md`, docstrings, comments) and machine-facing (e.g. this `CLAUDE.md`, schema descriptions) — is still accurate with respect to the changes the PR introduces. If anything has drifted, update it as part of the same PR.
```

Rationale: this `CLAUDE.md` itself encodes facts that live in `pyproject.toml`/`tox.ini`/`.flake8` (line length, mypy flags, env names, marker policy). Without an explicit "update the docs" reminder, those facts will rot the first time someone changes a tool config without touching `CLAUDE.md`, and future Claude sessions will be primed with stale guidance.

candleindark (changes requested):

CLAUDE.md
yarikoptic: see next item?

yarikoptic: but indeed we could just register them in this PR... will do. dandi-cli has pytest_plugin.py... may be I should just introduce it here too

yarikoptic: we even have some nice visualizer there for dependencies and DandiBaseModel

candleindark: I haven't get your response to https://github.com/dandi/dandi-cli/pull/1827/changes/BASE..0f8867bd1b98dddd281f9720f3597acc86e2f6f1#r3156553817 yet, but it seems that you are moving forward with using the mark. I really don't know how you can enforce it to produce a meaningful outcome. What it will eventually end up is that all tests submitted through an Ai, even with human intervention will end up that mark? My prediction is that the majority of the tests will eventual have that mark in the future, making that mark mostly meaningless in my opinion.

candleindark: > see next item?

What do you mean by next item?

yarikoptic:
odd, the [test-dandi-cli (ubuntu-latest, 3.11, release, client, default)](https://github.com/dandi/dandi-schema/actions/runs/25075609920/job/73467251030?pr=402#logs) seems got stuck on

```
tests/test_delete.py::test_delete_zarr_path {'size': 21}
....
{'status': 'uploading', 'progress': 100.0}
{'status': 'server calculating checksum'}
```
canceled, reran

CLAUDE.md
yarikoptic: it was the next item in the text; but nevermind now -- I addressed this by introducing an entire pytest_plugin.py and that marker is included now. Please re-review

yarikoptic:
stuck again, on `{'status': 'server calculating checksum'}` can't be coincidence, yet to check on master since should not relate AFAIK but I could be wrong. @copilot wdyt?

yarikoptic:
doh -- it is all that `rabbitmq:management` "goodness". Released since then dandi-cli should not have that issue -- restarting again!

yarikoptic:
if no further comments -- will merge tomorrow.

PR #409, opened 2026-05-14
feat: add numberOfSessions to AssetsSummary

bendichter:
## Summary
- Add optional `numberOfSessions: int` field to `AssetsSummary`, with a schema description documenting the counting semantics
- Bump `DANDI_SCHEMA_VERSION` to 0.7.1, since 0.7.0 is already published in dandi/schema and the release workflow rejects modifications to published schemata
- Aggregator counts unique `(subject, session)` pairs parsed from BIDS-style `sub-*` / `ses-*` tokens in asset paths. Keying on the pair avoids the undercount when two subjects share a session id like `ses-1`
- Filename is parsed first (matching existing `sub-`/`sample-` handling); directory components are scanned as a fallback so `ses-*` tokens that appear only in the path (e.g. `sub-X/ses-Y/foo_acq-Z_bold.nii.gz`) are still counted

## Caveats
- Assets whose paths carry no `sub-`/`ses-` tokens won't contribute. Note that `dandi organize` produces `sub-X/sub-X_ses-Y.nwb` layouts, so organized NWB dandisets with session ids are counted; only datasets that follow neither BIDS nor the organized layout are missed.

## Test plan
- [x] New `test_aggregate_number_of_sessions` covers: same subject across sessions, two subjects sharing a session id, `ses-` only in directory, `ses-` without any `sub-`, and the no-`ses-` case
- [x] Existing `test_aggregate` fixtures updated with expected `numberOfSessions` values
- [x] Existing `test_aggregation_bids` now asserts `numberOfSessions == 2`
- [x] Full suite: `268 passed, 17 skipped`; `tox -e lint,typing` clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dandischema/metadata.py
```diff
@@ -539,6 +540,31 @@ def _add_asset_to_stats(assetmeta: Dict[str, Any], stats: _stats_type) -> None:
...
+                if dir_session is None and part.startswith("ses-"):
+                    dir_session = part.split("ses-", 1)[1]
+        if dir_subject is not None and dir_session is not None:
+            found.setdefault("subject", dir_subject)
+            found["session"] = dir_session
+    stats["sessions"] = stats.get("sessions", [])
+    if found.get("subject") and found.get("session"):
+        pair = (found["subject"], found["session"])
+        if pair not in stats["sessions"]:
+            stats["sessions"].append(pair)
```
yarikoptic: too aisloppy -- easier to HI compose the logic but first potentially even generalize 2 prior blocks into a 'for' loop, then if no session was found in filename, go through path's parts and split on ses- and be done if any matches.

yarikoptic (changes requested):
THANK YOU for the PR.  Needs a little a human touch ;)

dandischema/metadata.py
bendichter: Done in cdb37ee — collapsed the `sub-`/`sample-`/`ses-` filename parsing into a single loop over an `entities` list, and simplified the session fallback to scanning `asset_path.parts[:-1]` for a `ses-X` directory and stopping at the first match. Dropped the directory-only subject fallback since BIDS files always carry `sub-` in the filename.

candleindark:
Is the `minor` label a bit too much? Since we don't have a major in our version number, patch has been carrying the semantics of a minor.

bendichter:
yes, I think patch makes sense for this while major is 0

dandischema/models.py
```diff
@@ -1201,6 +1201,9 @@ class AssetsSummary(DandiBaseModel):
         None, json_schema_extra={"readOnly": True}
     )  # more of NWB
     numberOfCells: Optional[int] = Field(None, json_schema_extra={"readOnly": True})
+    numberOfSessions: Optional[int] = Field(
+        None, json_schema_extra={"readOnly": True}
+    )  # BIDS ses-* tokens, counted as unique (subject, session) pairs
```
candleindark: Possibly add to the comment that this count assumes the dataset is valid BIDS dataset. (For example, a session without subject will not be counted)

candleindark: Possibly moving some of the info in the comment into the `description` argument of `Field`. That way, the info will be part of the model. However, other fields, e.g. `numberOfCells` is not currently doing that.

candleindark (commented):
LGTM as the targeted datasets are valid BIDS datasets. (See other comments for more details).

dandischema/models.py
bendichter: I would like to keep this information as comments rather than description. If we were to add a new data structure, then we could add a new way to compute numberOfSessions, which could still be a valid field. If we don't make notes about how the data is derived, we can support new data structure without making any changes to the schema. That's why I think the field should contain information about what the data is about, and comments should be used to give hints about how it was derived.

yarikoptic: well, that's what I wonder -- since we do have `sub-..._ses-...` even in DANDI names, and we do count on those -- we better count sessions as well in non-BIDS names as well. but "AI logic" above is wrong for that... will comment there

yarikoptic: actually my initial read was wrong -- there is a generic handling in the filename at https://github.com/dandi/dandi-schema/pull/409/changes#diff-196e8d7f9869422a0dba629b1f58a9972a300d67308aec783b2847cdb50ef5ffR541 

but wouldn't it work for DANDI names (no folder name) as well and thus just a matter of


```suggestion
    )  # BIDS or DANDI ses-* tokens (in file name or path), counted as unique (subject, session) pairs
```

?

bendichter: Right, it works on DANDI names too. Changed the description to say BIDS or DANDI ses- tokens in file names or paths.

Issue #406, opened 2026-05-12
Handle `Impossible to generate slot usage entry` complaints about changes in the `range` metadata slot


candleindark:
Sub-issue of #389 — tracks handling `Impossible to generate slot usage entry` complaints about changes in the `range` metadata slot.

Remaining instances in `dandischema/models.yaml` on `linkml-auto-converted`:

- `Dandiset.wasGeneratedBy` (parent range `Activity` narrowed to `Project`)
- `BareAsset.wasGeneratedBy` (parent range narrowed in subclass)

## Tasks

- [x] Formulate a solution to handle the complaints.
- [x] Add a CI workflow that checks/ensures the LinkML behavior which the solution depends on in the spirit of https://github.com/dandi/dandi-schema/issues/389#issuecomment-4410248860.

candleindark:
This issue can be closed. Both tasks are done, on `linkml-conversion`.

**The wording changed, as in #407.** The complaint is no longer `Impossible to generate slot usage entry`. After https://github.com/dandi/pydantic2linkml/pull/71 allowed partial `slot_usage` entry generation, both instances read ``Cannot express in a slot_usage entry a value for the `range` ``.

**Fix.** Both ranges are now specified explicitly through `dandischema/models_merge.yaml`, and the notes they made obsolete are stripped during sanitization:

- 6c7f0e77: specify the range of `Dandiset.wasGeneratedBy` as `Project`
- a786022a: specify the range of `BareAsset.wasGeneratedBy` as `Any`, alongside its existing `any_of`
- 75341f30: allow note removal to be scoped to locations in the document, so the removal targets only these two `notes`
- 5b50b3c5: drop the obsolete `range` notes on the two `wasGeneratedBy` slot usages

**Second task, the LinkML behavior check.** `tests/linkml_behavior/range_refinement/` pins both respecifications the fix depends on (narrowing a range to a subclass, and widening it to `Any` constrained by an `any_of`) against all three validators the pipeline uses: `linkml-validate`, `gen-json-schema` and `gen-pydantic`.

- 5d603ae4: add the `range_refinement` topic
- e259d12b: make the topic directories packages so sibling topics can coexist

These run in CI under the existing `LinkML behavior tests` workflow (`hatch run linkml-behavior-test:test`).

[...]

PR #417, opened 2026-05-29
feat: let `wasGeneratedBy` in `BareAsset` and `Dandiset` to have range of `Activity`

candleindark:
This PR is a possible solution to https://github.com/dandi/dandi-schema/issues/406

With `schemaKey` being the designator
type, instances of subclasses of
`Activity` should be able to validate
into the corresponding subclasses. Though
this solution would relex the schema a bit
for `BareAsset` and `Dandiset`

### DIFF
--- a/dandischema/models_merge.yaml
+++ b/dandischema/models_merge.yaml
@@ -14,6 +14,14 @@ classes:
       - "MANUAL_NOTE: The default of the `schemaKey` field in the corresponding Pydantic
       model in `dandischema.models` is not the model's name. Adjustment to the inherited
       `schemaKey` slot may be needed."
+    slot_usage:
+      wasGeneratedBy:
+        notes: null
+        any_of: null
+  Dandiset:
+    slot_usage:
+      wasGeneratedBy:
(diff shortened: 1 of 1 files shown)


candleindark:
As pointed out by @yarikoptic, we should be able to override the range of `Dandiset.wasGeneratedBy` to `Project` in the LinkML schema without violating the monotonicity rule to be in alignment with the definition in the Pydantic model, https://github.com/dandi/dandi-schema/blob/d75273828bb7742be74f23d9f8cf8a941f4af8e2/dandischema/models.py#L1755.

However, some tests against LinkML behavior should be added as well to check against regression.

yarikoptic:
do we still need this @candleindark ?

candleindark:
> do we still need this @candleindark ?

No. Let's close this. We don't need this for the LinkML translation anymore.

However, the type annotations for `wasGeneratedBy` in `CommonModel` and `BareAsset` are problematic in and of themselves, regardless of LinkML translation. I will submit a PR with a solution soon.

Issue #343, opened 2025-10-27
Provide "downgrade" functionality to `migrate` and make dandi-cli use `migrate`!


yarikoptic:
While trying to release new dandi-schema and reviewing dandi-cli tests fails in

- https://github.com/dandi/dandi-schema/pull/342

decided to review what we have ATM:

- way to "upgrade" in [`dandischema::migrate`](https://github.com/dandi/dandi-schema/blob/master/dandischema/metadata.py#L340) where we provide necessary transformations on a dict to go to current (higher only!) schema version from a prior.
  - we refuse to "downgrade"
- client refuses to upload if current dandischema is **not identical**:
```
ERROR tests/test_upload.py::test_upload_sync[False-False] - dandi.exceptions.SchemaVersionError: Server requires schema version 0.6.10; client only supports 0.7.0.  You may need to upgrade dandi and/or dandischema.
```
   - that was added in https://github.com/dandi/dandi-cli/pull/724 in 2021 to address
   - https://github.com/dandi/dandi-cli/issues/722 which was in likeliness to address missing metadata (older schema did not have it) as in
   - https://github.com/dandi/dandi-cli/issues/719. 
- I am not actually 100% certain why client refuses as I think it should be a server errorring out instead if schema is no good!
- I also fail to identify at what stage `migrate(` is actually used ATM besides some management commands in dandi-archive

```shell
❯ git grep -p 'migrate('
dandiapi/api/management/commands/migrate_published_version_metadata.py=def migrate_published_version_metadata(*, dandiset: str, published_version: str, to_version: str):
dandiapi/api/management/commands/migrate_published_version_metadata.py:        metanew = migrate(metadata, to_version=to_version, skip_validation=False)
dandiapi/api/management/commands/migrate_version_metadata.py=def migrate_version_metadata(*, to_version: str):
dandiapi/api/management/commands/migrate_version_metadata.py:            metanew = migrate(metadata, to_version=to_version, skip_validation=True)
web/src/migrate.ts=const __dirname = path.dirname(__filename);
web/src/migrate.ts:async function migrate(version: string) {
web/src/migrate.ts=if (process.argv.length !== 3) {
web/src/migrate.ts:migrate(process.argv[2]);
```
@dandi/dandi-archive  - when are those migrations triggered? (IIRC we might have done some migrations early in the days. but not recently) 

[...]

PR #419, opened 2026-06-08
Consolidate Asset/Dandiset models; gate publish validation on datePublished

candleindark:
## Summary

Make every model class's `schemaKey` value equal its class name, so that an eventual LinkML translation can use `schemaKey` as its type designator (`designates_type`). Three classes violated this (`BareAsset` → `"Asset"`, `PublishedAsset` → `"Asset"`, `PublishedDandiset` → `"Dandiset"`), and fixing them takes two distinct changes:

1. **Consolidate the publication-specific variants into their base classes:** `PublishedDandiset` merges into `Dandiset` and `PublishedAsset` into `Asset`. Publication requirements now fire via a `datePublished`-gated validator instead of separate classes.
2. **Align `BareAsset.schemaKey` to `"BareAsset"`**

Stored `schemaKey` values in archive data are unchanged, and the old names remain as deprecated aliases so `dandi-cli`/`dandi-archive` imports keep working. Adopting this release does, however, require one lockstep `dandi-cli` change (see Prerequisites).

`DANDI_SCHEMA_VERSION` is also bumped to `0.8.0`, which is due independently of the model consolidation: the schemas generated from `master` already differ from the published `0.7.0`, and this PR changes all four of them further. A summary of the diff against the published `0.7.0`, and of why none of it is breaking for existing metadata, is in https://github.com/dandi/dandi-schema/pull/419#issuecomment-5086656768.

## Prerequisites (merge first)

- **dandi/dandi-schema#422** — removes the `missing_ok` parameter of `metadata.validate`. That parameter is never really used, and its feature never actually worked as intended (see #422 for details); merging it first frees this PR from having to keep that behavior "working" while it reshapes the validation errors.
- **dandi/dandi-cli#1886** — set `schemaKey = "Asset"` on bare asset metadata at upload, since `BareAsset.schemaKey` is now `"BareAsset"` and the server does not normalize it.

<details>
<summary>What changed in <code>dandischema/models.py</code></summary>

- **Merge `PublishedDandiset` → `Dandiset`, `PublishedAsset` → `Asset`.** Publication-only fields (`doi`, `publishedBy`, `datePublished`, `releaseNotes` on `Dandiset`; `publishedBy`, `datePublished` on `Asset`) become optional, and the publication requirements move into a `datePublished`-gated `check_publication_status` model validator:
  - `datePublished` is `None` (draft) → publication-only fields must be absent;
  - `datePublished` set (published) → enforce the former `Published*` rules (`publishedBy`/`url`/`doi` presence, stricter `id`/`url` patterns, `check_filesbytes`, `digest_sha256check`).

  dandi-archive injects `datePublished` before validating, so the gated checks fire exactly as the `Published*` classes did.
- **Per-field validation errors.** The publication validator accumulates all violations and raises them via `pydantic_core.ValidationError.from_exception_data(...)`, so each error carries its own field `loc` (e.g. `('assetsSummary',)`) plus the usual `"Value error, …"` message — rather than one model-level error with an empty `loc`.
- **Keep `BareAsset`** (`Asset` still inherits it) with `schemaKey: Literal["BareAsset"]`; `Asset` overrides it to `Literal["Asset"]` (a localized `# type: ignore[assignment]` covers the narrowed override), and `ensure_schemakey` pins each instance's value to its class name.
- **Remove the `Publishable` mixin**; add `PublishedDandiset`/`PublishedAsset` as deprecated aliases.
- **Simplify `ensure_schemakey`** to a plain `schemaKey == class-name` check.
- `to_datacite` asserts the published precondition (`datePublished` set) the `PublishedDandiset` type used to guarantee.
</details>

<details>
<summary>What changed outside <code>dandischema/models.py</code></summary>

- **`dandischema/consts.py`:** `DANDI_SCHEMA_VERSION` becomes `0.8.0`, and `"0.7.0"` joins `ALLOWED_INPUT_SCHEMAS` so that metadata stamped with it remains valid input for `migrate()`.

The `tools/api-with-schema.Dockerfile` change that the `test-dandi-cli` results below depend on (adding `ENV UV_NO_SYNC=1` and dropping the undeclared `dev` extra) was split out into dandi/dandi-schema#435 and merged to `master`, so it is no longer part of this PR's diff; this branch inherits it after being rebased onto `master`.
</details>

<details>
<summary>Why gate on <code>datePublished</code> rather than an explicit (context-triggered) validator</summary>

The publish-only checks could instead be triggered explicitly by the caller (e.g. a Pydantic validation-context flag passed by the publish flow) rather than by the presence of `datePublished`. We chose data-driven gating because:

- **It can be expressed in LinkML; an explicit context cannot.** Gating on a real field is a conditional constraint on the data, which is exactly what a LinkML rule expresses (`gen-json-schema` emits it as `if`/`then`). A caller-supplied validation mode has no LinkML equivalent — LinkML describes only the data instance, not how validation is invoked — so it would have to live as hand-written Python forever, outside the source of truth, working against the migration this change is meant to enable.
- **It travels with the data.** The gated checks fire through both `validate()` and direct construction (`Dandiset(**data)` / `model_validate`). A context flag only rides `model_validate(..., context=...)`: the terse `PublishedDandiset(**meta)` form (used by `to_datacite` and in tests) cannot pass one, so it would silently skip the publish checks. Beyond the call-site rewrites that would force, a constructor whose enforcement quietly depends on how it is invoked is a lasting programming hazard.
- **It matches the domain.** `datePublished` present ⟺ published is a real invariant: a draft never carries it, and the archive injects it only when building a publishable version. "Published" is a state of the record, not a mode a caller opts into.
- **Published-ness stays introspectable.** With gating, "is this published?" is just `obj.datePublished is not None`; with a context flag, whether an object was validated as published is ephemeral to the call and recorded nowhere on the object.
</details>

<details>
<summary>Compatibility</summary>

- Generated JSON Schema (verified by before/after diff): draft `Dandiset`/`Asset` gain **only optional, `readOnly`** publication properties — nothing newly required, no pattern tightened — so the dandi-archive Meditor is unaffected (it hides `readOnly` props). The `published-*.json` schemas become the relaxed versions, with publication strictness enforced by the gated validator.
- `published-dandiset.json` and `published-asset.json` are now byte-identical to `dandiset.json` and `asset.json`, because `publish_model_schemata` generates each file from the class named in `SCHEMA_MAP` and `PublishedDandiset`/`PublishedAsset` are now aliases of `Dandiset`/`Asset`.
- The publication requirements no longer appear in the generated JSON Schema at all, since they are conditional on `datePublished` and a flat `required` list cannot express that; the gated validator enforces them instead. That is an acceptable gap for now: the plan is to re-encode these requirements as LinkML rules in the LinkML translation, where `gen-json-schema` emits them as `if`/`then`.
- `PublishedDandiset`/`PublishedAsset`/`BareAsset` imports continue to work in `dandi-cli`/`dandi-archive`.
</details>

## Follow-ups (separate PRs)

- Bump the `dandischema` pin in dandi-archive and dandi-cli once this change is merged and released. Both pin `dandischema`, and `DANDI_SCHEMA_VERSION` must match across the two services, so they need to move to the same release together. This is also what clears the expected `test-dandi-cli` failures described under Test plan.
  - The bump of  `dandischema` in dandi-cli has been implemented in https://github.com/dandi/dandi-cli/pull/1898
  - The bump of  `dandischema` in dandi-archive is being tracked with https://github.com/dandi/dandi-archive/issues/2876
- [x] dandi-archive `_encode_pydantic_error` robustness (dandi/dandi-archive#2851). This PR's per-field errors avoid the empty-`loc` case at the source, but that fix is still worth having as defense-in-depth.
- https://github.com/dandi/dandi-schema/issues/440
## Test plan

- [x] `tox -e py3` (262 passed, 17 skipped — skips are environment-only: no `DOI_PREFIX` / instance name not "DANDI")
- [x] `tox -e lint,typing`
- [x] Generated JSON Schema before/after diff reviewed, and diffed against the published `0.7.0` release in `dandi/schema` (see the comment linked from the Summary)
- [x] Added tests for the `datePublished`-gated publish requirements and the coherence invariant
- [x] Verified against dandi-archive's own CI via the throwaway PR dandi/dandi-archive#2866, which pins `dandischema` to this branch. After this PR was rebased onto `master` (following the dandi/dandi-schema#435 split), that pin was re-resolved to the current tip (`227cdd01`) and the run re-triggered; this box will be checked once the refreshed run passes. The earlier run, against the now-dropped pre-split tip, had the archive's publication-validation suite (`backend-ci`) passing against the consolidated models and both `cli-integration` variants (consolidated-model server against a released-`dandischema` client, on dandi-cli `master` and `release`) passing.

### Expected `test-dandi-cli` failures

Bumping `DANDI_SCHEMA_VERSION` ahead of dandi-archive and dandi-cli leaves integration jobs red: three of them today, and two once dandi/dandi-cli#1894 merges, since the `server` job runs dandi-cli `master` and that PR drops the incidental requirement that both sides carry the same `DANDI_SCHEMA_VERSION`. Each is version skew between the two sides of the test rather than a regression from the model changes, and none can be fixed from this PR; the remaining two clear when both services move to this release.

<details>
<summary>Which jobs fail, and why</summary>

| job | client schema | server schema | outcome |
| --- | --- | --- | --- |
| ubuntu, `client`, dandi-cli `master` | 0.8.0 | 0.7.0 | fail |
| ubuntu, `client`, dandi-cli `release` | 0.8.0 | 0.7.0 | fail |
| ubuntu, `server` | 0.7.0 | 0.8.0 | fail, until dandi/dandi-cli#1894 merges |
| ubuntu, `both` (default and ember-dandi) | 0.8.0 | 0.8.0 | pass |
| macOS and Windows, `client` | 0.8.0 | n/a | pass, no Docker so the server-backed tests are skipped |

- The two `client` jobs install this branch into the client only, so a `0.8.0` client meets the released `0.7.0` server. dandi-cli treats a server that is behind by a MINOR schema version as fatal (`dandi/dandiapi.py:740-747`), so nearly every server-backed test errors with `SchemaVersionError: Server uses older incompatible schema version 0.7.0; client supports 0.8.0`.
- The `server` job is the mirror image: a released client against a `0.8.0` server. Only 5 of 1006 tests fail, all `cli/tests/test_service_scripts.py::test_update_dandiset_from_doi`, because that test builds its expected `schemaVersion` and `@context` from the client's own `DANDI_SCHEMA_VERSION` (`test_service_scripts.py:111-118`) and so expects `0.7.0` where the server emits `0.8.0`. dandi/dandi-cli#1894 fixes that expectation, and since this job runs dandi-cli `master`, it goes green on a re-run once that PR merges.
- The `both` jobs install this branch on both sides and pass, which is the signal that the consolidated models and the new schema version work end to end.

Before the `UV_NO_SYNC=1` fix (added in dandi/dandi-schema#435, now in `master` and inherited here via rebase), the `both` jobs failed exactly like the `client` jobs, and the `server` job passed only because the ineffective override left both sides agreeing on `0.7.0`. Its red status here is a consequence of the override finally working.
</details>

### DIFF
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -35,6 +35,7 @@ from pydantic import (
 )
 from pydantic.json_schema import JsonSchemaValue
 from pydantic_core import CoreSchema
+from typing_extensions import Self
 from zarr_checksum.checksum import InvalidZarrChecksum, ZarrDirectoryDigest
 
 from dandischema.conf import (
@@ -574,15 +575,8 @@ class DandiBaseModel(BaseModel):
     @field_validator("schemaKey")
     @classmethod
     def ensure_schemakey(cls, val: str) -> str:
-        tempval = val
-        if "Published" in cls.__name__:
-            tempval = "Published" + tempval
-        elif "BareAsset" == cls.__name__:
-            tempval = "Bare" + tempval
-        if tempval != cls.__name__:
-            raise ValueError(
-                f"schemaKey {tempval} does not match classname {cls.__name__}"
-            )
+        if val != cls.__name__:
+            raise ValueError(f"schemaKey {val} does not match classname {cls.__name__}")
         return val
 
     @classmethod
@@ -1756,6 +1750,40 @@ class Dandiset(CommonModel):
 
     version: str = Field(json_schema_extra={"nskey": "schema", "readOnly": True})
 
+    # Publication-related fields. A non-``None`` ``datePublished`` marks the
+    # record as published; ``check_publication_status`` enforces the cross-field
+    # rules (which of these are then required vs. must be absent).
+    url: Optional[AnyHttpUrl] = Field(
+        None,
+        description="Permalink to the Dandiset.",
+        json_schema_extra={"readOnly": True, "nskey": "schema"},
+    )
+    doi: Optional[str] = Field(
+        None,
+        title="DOI",
+        pattern=DANDI_DOI_PATTERN,
+        json_schema_extra={"readOnly": True, "nskey": DANDI_NSKEY},
+    )
+    """The DOI of the published Dandiset.
+
+    The empty string indicates that there is no DOI for the published Dandiset;
+    it is set automatically for a published Dandiset on a DANDI instance with no
+    configured DOI prefix.
+    """
+    publishedBy: Optional[Union[AnyHttpUrl, PublishActivity]] = Field(
+        None,
+        description="The URL should contain the provenance of the publishing process.",
+        json_schema_extra={"readOnly": True, "nskey": DANDI_NSKEY},
+    )
+    datePublished: Optional[datetime] = Field(
+        None, json_schema_extra={"readOnly": True, "nskey": "schema"}
+    )
+    releaseNotes: Optional[str] = Field(
+        None,
+        description="The description of the release",
+        json_schema_extra={"readOnly": True, "nskey": "schema"},
+    )
+
     wasGeneratedBy: Optional[Sequence[Project]] = Field(
         None,
         title="Associated projects",
@@ -1772,6 +1800,54 @@ class Dandiset(CommonModel):
         "rdfs:label": "Information about the dataset",
         "nskey": DANDI_NSKEY,
     }
+
+    @model_validator(mode="after")
+    def check_publication_status(self) -> Self:
+        """Enforce publication coherence keyed on ``datePublished``.
+
+        A non-``None`` ``datePublished`` marks the record as published and
+        triggers the published-Dandiset requirements; otherwise the publication-only
+        fields must be absent.
+        """
+        errors = []
+        if self.datePublished is None:
+            errors += [
+                f"{name} is not allowed unless datePublished is set"
+                for name in ("doi", "publishedBy", "releaseNotes")
+                if getattr(self, name) is not None
+            ]
+        else:
+            if self.publishedBy is None:
+                errors.append("publishedBy is required for a published Dandiset")
+            if self.url is None:
+                errors.append("url is required for a published Dandiset")
+            elif not re.match(PUBLISHED_VERSION_URL_PATTERN, str(self.url)):
+                errors.append(
+                    f'url does not match regex "{PUBLISHED_VERSION_URL_PATTERN}"'
+                )
+            # Stricter than the ``id`` field pattern, which also accepts
+            # ``/draft`` and a lowercase prefix; only a published version id
+            # passes here.
+            if not re.match(DANDI_PUBID_PATTERN, self.id):
+                errors.append(
+                    f'id "{self.id}" does not match the published-Dandiset pattern '
+                    f'"{DANDI_PUBID_PATTERN}"'
+                )
+            if (
+                self.assetsSummary.numberOfBytes == 0
+                or self.assetsSummary.numberOfFiles == 0
+            ):
+                errors.append(
+                    "A Dandiset containing no files or zero bytes is not publishable"
+                )
+            if self.doi is None:
+                if _INSTANCE_CONFIG.doi_prefix is None:
+                    self.doi = ""
+                else:
+                    errors.append("doi is required for a published Dandiset")
+        if errors:
+            raise ValueError("; ".join(errors))
+        return self
 
 
 class BareAsset(CommonModel):
@@ -1845,9 +1921,8 @@ class BareAsset(CommonModel):
         json_schema_extra={"nskey": "prov"},
     )
 
-    # Bare asset is to be just Asset.
-    schemaKey: Literal["Asset"] = Field(
-        "Asset", validate_default=True, json_schema_extra={"readOnly": True}
+    schemaKey: Literal["BareAsset"] = Field(
+        "BareAsset", validate_default=True, json_schema_extra={"readOnly": True}
     )
 
     _ldmeta = {
@@ -1898,6 +1973,10 @@ class BareAsset(CommonModel):
 class Asset(BareAsset):
     """Metadata used to describe an asset on the server."""
 
+    # A non-``None`` ``datePublished`` marks the asset as published, which
+    # additionally requires ``publishedBy``, a published-asset ``id``, and (for a
+    # non-zarr asset) a ``sha2_256`` digest; see ``check_publication_status``.
+
     # all of the following are set by server
     id: str = Field(
         json_schema_extra={"readOnly": True}, description="Uniform resource identifier."
@@ -1907,102 +1986,64 @@ class Asset(BareAsset):
         json_schema_extra={"readOnly": True, "nskey": "schema"}
     )
 
-
-class Publishable(DandiBaseModel):
-    publishedBy: Union[AnyHttpUrl, PublishActivity] = Field(
+    # Publication-related fields. Present only on a published asset, i.e. when
+    # ``datePublished`` is set.
+    publishedBy: Optional[Union[AnyHttpUrl, PublishActivity]] = Field(
+        None,
         description="The URL should contain the provenance of the publishing process.",
         json_schema_extra={"readOnly": True, "nskey": DANDI_NSKEY},
     )
-    datePublished: datetime = Field(
-        json_schema_extra={"readOnly": True, "nskey": "schema"}
-    )
-    schemaKey: Literal["Publishable", "Dandiset", "Asset"] = Field(
-        "Publishable", validate_default=True, json_schema_extra={"readOnly": True}
+    datePublished: Optional[datetime] = Field(
+        None, json_schema_extra={"readOnly": True, "nskey": "schema"}
     )
 
-
-_doi_field_kwargs: dict[str, Any] = {
-    "title": "DOI",
-    "pattern": DANDI_DOI_PATTERN,
-    "json_schema_extra": {"readOnly": True, "nskey": DANDI_NSKEY},
-}
-if _INSTANCE_CONFIG.doi_prefix is None:
-    _doi_field_kwargs["default"] = ""
-
-
-class PublishedDandiset(Dandiset, Publishable):
-    id: str = Field(
-        description="Uniform resource identifier.",
-        pattern=DANDI_PUBID_PATTERN,
-        json_schema_extra={"readOnly": True},
-    )
-    doi: str = Field(**_doi_field_kwargs)
-    """
-    The DOI of the published Dandiset
-
-    The value of the empty string indicates that there is no DOI for the published
-    Dandiset.
-    """
-
-    url: AnyHttpUrl = Field(
-        description="Permalink to the Dandiset.",
-        json_schema_extra={"readOnly": True, "nskey": "schema"},
-    )
-    releaseNotes: Optional[str] = Field(
-        None,
-        description="The description of the release",
-        json_schema_extra={"readOnly": True, "nskey": "schema"},
-    )
-
-    schemaKey: Literal["Dandiset"] = Field(
-        "Dandiset", validate_default=True, json_schema_extra={"readOnly": True}
-    )
-
-    @field_validator("assetsSummary")
-    @classmethod
-    def check_filesbytes(cls, values: AssetsSummary) -> AssetsSummary:
-        if values.numberOfBytes == 0 or values.numberOfFiles == 0:
-            raise ValueError(
-                "A Dandiset containing no files or zero bytes is not publishable"
-            )
-        return values
-
-    @field_validator("url")
-    @classmethod
-    def check_url(cls, url: AnyHttpUrl) -> AnyHttpUrl:
-        if not re.match(PUBLISHED_VERSION_URL_PATTERN, str(url)):
-            raise ValueError(
-                f'string does not match regex "{PUBLISHED_VERSION_URL_PATTERN}"'
-            )
-        return url
-
-
-class PublishedAsset(Asset, Publishable):
-    id: str = Field(
-        description="Uniform resource identifier.",
-        pattern=ASSET_UUID_PATTERN,
-        json_schema_extra={"readOnly": True},
-    )
-
-    schemaKey: Literal["Asset"] = Field(
+    # mypy flags the narrowed Literal as an incompatible override of
+    # BareAsset's, but pydantic allows it and ensure_schemakey pins each
+    # instance to its class name at runtime.
+    schemaKey: Literal["Asset"] = Field(  # type: ignore[assignment]
         "Asset", validate_default=True, json_schema_extra={"readOnly": True}
     )
 
-    @field_validator("digest")
-    @classmethod
-    def digest_sha256check(
-        cls, v: Dict[DigestType, str], info: ValidationInfo
-    ) -> Dict[DigestType, str]:
-        values = info.data
-        if values.get("encodingFormat") != "application/x-zarr":
-            if DigestType.sha2_256 not in v:
-                raise ValueError("A non-zarr asset must have a sha2_256.")
-            digest = v[DigestType.sha2_256]
-            if not re.fullmatch(SHA256_PATTERN, digest):
-                raise ValueError(
-                    f"Digest must have an appropriate sha2_256 value. Got {digest}"
+    @model_validator(mode="after")
+    def check_publication_status(self) -> Self:
+        """Enforce publication coherence keyed on ``datePublished``.
+
+        A non-``None`` ``datePublished`` marks the asset as published and
+        triggers the published-asset requirements; otherwise ``publishedBy``
+        must be absent.
+        """
+        errors = []
+        if self.datePublished is None:
+            if self.publishedBy is not None:
+                errors.append("publishedBy is not allowed unless datePublished is set")
+        else:
+            if self.publishedBy is None:
+                errors.append("publishedBy is required for a published asset")
+            if not re.match(ASSET_UUID_PATTERN, self.id):
+                errors.append(
+                    f'id "{self.id}" does not match the published-asset pattern '
+                    f'"{ASSET_UUID_PATTERN}"'
                 )
-        return v
+            if self.encodingFormat != "application/x-zarr":
+                if DigestType.sha2_256 not in self.digest:
+                    errors.append("A non-zarr asset must have a sha2_256.")
+                elif not re.fullmatch(SHA256_PATTERN, self.digest[DigestType.sha2_256]):
+                    errors.append(
+                        "Digest must have an appropriate sha2_256 value. "
+                        f"Got {self.digest[DigestType.sha2_256]}"
+                    )
+        if errors:
+            raise ValueError("; ".join(errors))
+        return self
+
+
+# ``PublishedDandiset`` and ``PublishedAsset`` were consolidated into
+# ``Dandiset`` and ``Asset`` respectively: a record is "published" when its
+# ``datePublished`` is set (see each class's ``check_publication_status``).
+# These names are kept as deprecated aliases for backward compatibility and
+# will be removed in a future release.
+PublishedDandiset = Dandiset
+PublishedAsset = Asset
 
 
 def get_schema_version() -> str:
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -34,7 +34,8 @@ from pydantic import (
     model_validator,
 )
 from pydantic.json_schema import JsonSchemaValue
-from pydantic_core import CoreSchema
+from pydantic_core import CoreSchema, InitErrorDetails
+from pydantic_core import ValidationError as _CoreValidationError
 from typing_extensions import Self
 from zarr_checksum.checksum import InvalidZarrChecksum, ZarrDirectoryDigest
 
@@ -1751,8 +1752,7 @@ class Dandiset(CommonModel):
     version: str = Field(json_schema_extra={"nskey": "schema", "readOnly": True})
 
     # Publication-related fields. A non-``None`` ``datePublished`` marks the
-    # record as published; ``check_publication_status`` enforces the cross-field
-    # rules (which of these are then required vs. must be absent).
+    # record as published; ``check_publication_status`` enforces the rules.
     url: Optional[AnyHttpUrl] = Field(
         None,
         description="Permalink to the Dandiset.",
@@ -1807,46 +1807,59 @@ class Dandiset(CommonModel):
 
         A non-``None`` ``datePublished`` marks the record as published and
         triggers the published-Dandiset requirements; otherwise the publication-only
-        fields must be absent.
+        fields must be absent. Each violation is reported against its own field.
         """
-        errors = []
+        line_errors: list[InitErrorDetails] = []
+
+        def err(field: str, msg: str) -> None:
+            line_errors.append(
+                InitErrorDetails(
+                    type="value_error",
+                    loc=(field,),
+                    input=getattr(self, field),
+                    ctx={"error": msg},
+                )
+            )
+
         if self.datePublished is None:
-            errors += [
-                f"{name} is not allowed unless datePublished is set"
-                for name in ("doi", "publishedBy", "releaseNotes")
-                if getattr(self, name) is not None
-            ]
+            for name in ("doi", "publishedBy", "releaseNotes"):
+                if getattr(self, name) is not None:
+                    err(name, f"{name} is not allowed unless datePublished is set")
         else:
             if self.publishedBy is None:
-                errors.append("publishedBy is required for a published Dandiset")
+                err("publishedBy", "publishedBy is required for a published Dandiset")
             if self.url is None:
-                errors.append("url is required for a published Dandiset")
+                err("url", "url is required for a published Dandiset")
             elif not re.match(PUBLISHED_VERSION_URL_PATTERN, str(self.url)):
-                errors.append(
-                    f'url does not match regex "{PUBLISHED_VERSION_URL_PATTERN}"'
+                err(
+                    "url", f'url does not match regex "{PUBLISHED_VERSION_URL_PATTERN}"'
                 )
             # Stricter than the ``id`` field pattern, which also accepts
             # ``/draft`` and a lowercase prefix; only a published version id
             # passes here.
             if not re.match(DANDI_PUBID_PATTERN, self.id):
-                errors.append(
+                err(
+                    "id",
                     f'id "{self.id}" does not match the published-Dandiset pattern '
-                    f'"{DANDI_PUBID_PATTERN}"'
+                    f'"{DANDI_PUBID_PATTERN}"',
                 )
             if (
                 self.assetsSummary.numberOfBytes == 0
                 or self.assetsSummary.numberOfFiles == 0
             ):
-                errors.append(
-                    "A Dandiset containing no files or zero bytes is not publishable"
+                err(
+                    "assetsSummary",
+                    "A Dandiset containing no files or zero bytes is not publishable",
                 )
             if self.doi is None:
                 if _INSTANCE_CONFIG.doi_prefix is None:
                     self.doi = ""
                 else:
-                    errors.append("doi is required for a published Dandiset")
-        if errors:
-            raise ValueError("; ".join(errors))
+                    err("doi", "doi is required for a published Dandiset")
+        if line_errors:
+            raise _CoreValidationError.from_exception_data(
+                self.__class__.__name__, line_errors
+            )
         return self
 
 
@@ -1973,9 +1986,8 @@ class BareAsset(CommonModel):
 class Asset(BareAsset):
     """Metadata used to describe an asset on the server."""
 
-    # A non-``None`` ``datePublished`` marks the asset as published, which
-    # additionally requires ``publishedBy``, a published-asset ``id``, and (for a
-    # non-zarr asset) a ``sha2_256`` digest; see ``check_publication_status``.
+    # A non-``None`` ``datePublished`` marks the asset as published; see
+    # ``check_publication_status`` for the rules that then apply.
 
     # all of the following are set by server
     id: str = Field(
@@ -2010,30 +2022,50 @@ class Asset(BareAsset):
 
         A non-``None`` ``datePublished`` marks the asset as published and
         triggers the published-asset requirements; otherwise ``publishedBy``
-        must be absent.
+        must be absent. Each violation is reported against its own field.
         """
-        errors = []
+        line_errors: list[InitErrorDetails] = []
+
+        def err(field: str, msg: str) -> None:
+            line_errors.append(
+                InitErrorDetails(
+                    type="value_error",
+                    loc=(field,),
+                    input=getattr(self, field),
+                    ctx={"error": msg},
+                )
+            )
+
         if self.datePublished is None:
             if self.publishedBy is not None:
-                errors.append("publishedBy is not allowed unless datePublished is set")
+                err(
+                    "publishedBy",
+                    "publishedBy is not allowed unless datePublished is set",
+                )
         else:
             if self.publishedBy is None:
-                errors.append("publishedBy is required for a published asset")
+                err("publishedBy", "publishedBy is required for a published asset")
             if not re.match(ASSET_UUID_PATTERN, self.id):
-                errors.append(
+                err(
+                    "id",
                     f'id "{self.id}" does not match the published-asset pattern '
-                    f'"{ASSET_UUID_PATTERN}"'
+                    f'"{ASSET_UUID_PATTERN}"',
                 )
             if self.encodingFormat != "application/x-zarr":
                 if DigestType.sha2_256 not in self.digest:
-                    errors.append("A non-zarr asset must have a sha2_256.")
-                elif not re.fullmatch(SHA256_PATTERN, self.digest[DigestType.sha2_256]):
-                    errors.append(
-                        "Digest must have an appropriate sha2_256 value. "
-                        f"Got {self.digest[DigestType.sha2_256]}"
-                    )
-        if errors:
-            raise ValueError("; ".join(errors))
+                    err("digest", "A non-zarr asset must have a sha2_256.")
+                else:
+                    sha2_256_digest = self.digest[DigestType.sha2_256]
+                    if not re.fullmatch(SHA256_PATTERN, sha2_256_digest):
+                        err(
+                            "digest",
+                            "Digest must have an appropriate sha2_256 value. "
+                            f"Got {sha2_256_digest}",
+                        )
+        if line_errors:
+            raise _CoreValidationError.from_exception_data(
+                self.__class__.__name__, line_errors
+            )
         return self
 
 
--- a/dandischema/consts.py
+++ b/dandischema/consts.py
@@ -1,6 +1,6 @@
 from packaging.version import Version as _Version
 
-DANDI_SCHEMA_VERSION = "0.7.0"
+DANDI_SCHEMA_VERSION = "0.8.0"
 ALLOWED_INPUT_SCHEMAS = [
     "0.4.4",
     "0.5.1",
@@ -16,6 +16,7 @@ ALLOWED_INPUT_SCHEMAS = [
     "0.6.8",
     "0.6.9",
     "0.6.10",
+    "0.7.0",
     DANDI_SCHEMA_VERSION,
 ]
 
--- a/dandischema/datacite/__init__.py
+++ b/dandischema/datacite/__init__.py
@@ -118,17 +118,23 @@ def to_datacite(
     validate: bool = False,
     publish: bool = False,
 ) -> dict:
-    """Convert published Dandiset metadata to Datacite"""
+    """Convert published Dandiset metadata to Datacite
+
+    Raises
+    ------
+    ValueError
+        If ``meta`` does not represent published Dandiset metadata, i.e. its
+        ``datePublished`` is not set.
+    """
 
     instance_config = get_instance_config()
 
     if not isinstance(meta, PublishedDandiset):
         meta = PublishedDandiset(**meta)
 
-    # ``to_datacite`` operates on *published* Dandiset metadata. Since the
-    # publication-only fields are now optional on ``Dandiset`` (gated on
-    # ``datePublished``), enforce the precondition the ``PublishedDandiset`` type
-    # used to guarantee.
+    # ``to_datacite`` operates on *published* Dandiset metadata, but the
+    # publication-only fields are optional on ``Dandiset`` (gated on
+    # ``datePublished``), so enforce that precondition explicitly here.
     if meta.datePublished is None:
         raise ValueError(
             "to_datacite requires published Dandiset metadata, but datePublished "
--- a/dandischema/datacite/__init__.py
+++ b/dandischema/datacite/__init__.py
@@ -125,6 +125,16 @@ def to_datacite(
     if not isinstance(meta, PublishedDandiset):
         meta = PublishedDandiset(**meta)
 
+    # ``to_datacite`` operates on *published* Dandiset metadata. Since the
+    # publication-only fields are now optional on ``Dandiset`` (gated on
+    # ``datePublished``), enforce the precondition the ``PublishedDandiset`` type
+    # used to guarantee.
+    if meta.datePublished is None:
+        raise ValueError(
+            "to_datacite requires published Dandiset metadata, but datePublished "
+            "is not set"
+        )
+
     attributes: Dict[str, Any] = {}
     if publish:
         attributes["event"] = "publish"
--- a/dandischema/datacite/tests/test_datacite.py
+++ b/dandischema/datacite/tests/test_datacite.py
@@ -260,6 +260,11 @@ def test_datacite(dandi_id: str, schema: Any) -> None:
 
     datacite = to_datacite(meta=meta, validate=True)
 
+    # ``doi`` is Optional on the consolidated ``Dandiset`` but always a ``str`` on a
+    # successfully validated published one (here via ``basic_publishmeta``); narrow it
+    # for the str-typed ``datacite_post`` argument.
+    assert meta.doi is not None
+
     # trying to post datacite
     datacite_post(datacite, meta.doi)
 
--- a/dandischema/tests/test_metadata.py
+++ b/dandischema/tests/test_metadata.py
@@ -13,7 +13,6 @@ from dandischema.utils import TransitionalGenerateJsonSchema
 
 from .utils import (
     DANDISET_METADATA_DIR,
-    DOI_PREFIX,
     INSTANCE_NAME,
     METADATA_DIR,
     skipif_instance_name_not_dandi,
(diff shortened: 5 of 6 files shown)


dandischema/models.py
```diff
@@ -1769,6 +1798,67 @@ def contributor_musthave_contact(
...
+                    loc=(field,),
+                    input=getattr(self, field),
+                    ctx={"error": msg},
+                )
+            )
+
+        if self.datePublished is None:
+            for name in ("doi", "publishedBy", "releaseNotes"):
+                if getattr(self, name) is not None:
+                    err(name, f"{name} is not allowed unless datePublished is set")
```
yarikoptic, line 1827: please express this model/check as linkml "rule" to validate `Dandiset` published and not, and also compare to how it would look like if we made it keyed on having ANY of the publish-related fields set.  Also how errors would look like in both cases.

candleindark: I worked this up as self-contained, runnable LinkML demos (linkml 1.11: `linkml-validate` / `gen-json-schema` / `gen-pydantic`), and the headline is reassuring: **LinkML `rules` can make one field's required-or-forbidden status depend on whether another field is set, so the publication-status dependency this PR enforces translates directly to LinkML.** By "publication-status dependency" I mean what the PR's `check_publication_status` implements: the publication fields must be set together, present when the record is published and absent when it is a draft. Nothing here suggests changing the model-validator in this PR.

Two ways to express it, and they turn out to be logically equivalent:

### Case 1: keyed on `datePublished` (matches this PR)

Both directions of the PR's `check_publication_status` become rules:

```yaml
rules:
  # published: datePublished set => publishedBy and doi required
  - preconditions:  { slot_conditions: { datePublished: { value_presence: PRESENT } } }
    postconditions: { slot_conditions: { publishedBy: { required: true }, doi: { required: true } } }
  # draft: datePublished absent => publishedBy and doi absent  (one rule per field, see gotcha below)
  - preconditions:  { slot_conditions: { datePublished: { value_presence: ABSENT } } }
    postconditions: { slot_conditions: { publishedBy: { value_presence: ABSENT } } }
  - preconditions:  { slot_conditions: { datePublished: { value_presence: ABSENT } } }
    postconditions: { slot_conditions: { doi: { value_presence: ABSENT } } }
```

### Case 2: keyed on ANY of the fields being set

```yaml
rules:
  - preconditions:
      any_of:
        - slot_conditions: { datePublished: { value_presence: PRESENT } }
        - slot_conditions: { publishedBy:   { value_presence: PRESENT } }
        - slot_conditions: { doi:           { value_presence: PRESENT } }
    postconditions:
      slot_conditions:
        datePublished: { required: true }
        publishedBy:   { required: true }
        doi:           { required: true }
```

### They are logically equivalent

Both admit exactly "all three present, or all three absent". Validating all 8 presence combinations against each gives identical pass/fail in every case:

| present set | Case 1 | Case 2 |
|---|---|---|
| `{}` | pass | pass |
| `{datePublished}` / `{publishedBy}` / `{doi}` | fail | fail |
| any two of the three | fail | fail |
| all three | pass | pass |

So the choice of keying is an implementation detail: in LinkML (Case 1 vs Case 2), and equally in the Python model-validator, which could key on `datePublished` or on "any publication field is set" to the same effect. **The design in this PR does not need to change on account of LinkML.** That said, keying on `datePublished` is conceptually cleaner than keying on "any publication field is set", regardless of the encoding:

- **Unambiguous intent.** `datePublished is not None` is the single, authoritative test for "this is a published version", a real domain invariant (the archive sets `datePublished` exactly when it builds a publishable version). Keying on "any publication field is set" instead treats an accidentally-populated field as a declaration that the record is published, when it may just be a mistake.
- **Cleaner docs and user code.** Downstream code and documentation can define published/draft off one field instead of "any of N".
- **More directional error messages.** Because the datePublished-keyed Python validator knows which way the dependency is violated, it can say either "doi is not allowed unless datePublished is set" (a stray field on a draft) or "doi is required for a published Dandiset" (a missing field on a published record). Keyed on "any publication field is set", the validator has only one move: demand the missing siblings. (This concerns the hand-written Python messages. The LinkML CLI messages shown below are unambiguous in both keyings, but being derived solely from JSON Schema validation errors, they can't carry the finer-grained, directional information a Pydantic validator can.)

### How the errors look

These are the messages from `linkml-validate`, the LinkML CLI validator, which by default validates an instance against the JSON Schema that `gen-json-schema` derives from the rules (they become `if`/`then`). Full matrix below.

Published-but-incomplete (`{datePublished: ...}`), identical under both keyings:

```
[ERROR] [.../0] 'publishedBy' is a required property in /
[ERROR] [.../0] 'doi' is a required property in /
```

The distinguishing input, a stray field with no `datePublished` (`{doi: ...}`):

- Case 1 (keyed on `datePublished`) flags the stray field:
  ```
  [ERROR] [.../0] {'doi': '...'} should not be valid under {'required': ['doi']} in /
  ```
- Case 2 (keyed on any) instead demands the missing siblings:
  ```
  [ERROR] [.../0] 'datePublished' is a required property in /
  [ERROR] [.../0] 'publishedBy' is a required property in /
  ```

Both are unambiguous messages; they just point at different fixes, matching each schema's semantics. One caveat: the structured `loc` is the object root (`in /`), not the offending field, because JSON Schema `required` attaches at the object level. The field name does still appear in the message text, it is just not in the machine-readable `loc`, which is coarser than the per-field `loc` the Pydantic `from_exception_data` validator in this PR produces.

### Note for the migration: `gen-pydantic` does NOT enforce these rules

The rules only produce enforcement through `gen-json-schema`. `gen-pydantic` emits them as inert `linkml_meta` metadata with no validator, so the generated Pydantic model happily constructs every one of the invalid instances above (verified for both keyings, matrix below). This is specific to cross-field `rules`: `gen-pydantic` does still enforce field-level constraints (types, `required`, `pattern`, enums) as ordinary Pydantic field validators.

This isn't a blocker, just a decision about where validation lives. Rather than keeping a hand-written `model_validator` forever, dandischema could delegate validation to LinkML (validating against the generated JSON Schema via the linkml validation path) and use the generated Pydantic models for loading and manipulating data. In that split, field-level constraints are still checked by Pydantic on load, and the cross-field publication dependency is enforced by the LinkML rules. In this PR, which predates the migration, that enforcement lives in the `model_validator` as before.

<details>
<summary>Aside: a gotcha writing the draft-direction rule (likely a LinkML bug; will file separately)</summary>

The natural way to forbid the fields on a draft, a single rule with two `value_presence: ABSENT` postconditions, is **wrong**:

```yaml
  - preconditions:  { slot_conditions: { datePublished: { value_presence: ABSENT } } }
    postconditions: { slot_conditions: { publishedBy: { value_presence: ABSENT }, doi: { value_presence: ABSENT } } }
```

`gen-json-schema` compiles that `then` to `{"not": {"required": ["publishedBy", "doi"]}}`, i.e. ¬(publishedBy ∧ doi) = "not *both* present", so `{publishedBy}` or `{doi}` alone slips through. Splitting into one rule per forbidden field yields two separate `{"not": {"required": ["publishedBy"]}}` / `["doi"]`, which correctly AND to "neither present". I'll file this against LinkML.
</details>

### Reproduce

Everything above was produced with linkml 1.11 (`linkml`/`linkml-runtime` 1.11.0). The Pydantic models were generated with:

```console
$ gen-pydantic case1_datePublished_keyed.yaml > case1_models.py
$ gen-pydantic case2_any_field_keyed.yaml     > case2_models.py
```

JSON Schema validation of a metadata file against a schema, and inspecting the generated `if`/`then`:

```console
$ linkml-validate -s case1_datePublished_keyed.yaml -C Dandiset metadata/<file>.yaml
$ gen-json-schema --top-class Dandiset case1_datePublished_keyed.yaml
```

The full set of files follows.

<details>
<summary><code>case1_datePublished_keyed.yaml</code> (correct Case 1: draft direction split into one rule per field)</summary>

```yaml
id: https://example.org/case1_datePublished_keyed
name: case1_datePublished_keyed
prefixes:
  linkml: https://w3id.org/linkml/
imports:
  - linkml:types
default_range: string
classes:
  Dandiset:
    slots: [datePublished, publishedBy, doi]
    rules:
      - preconditions:
          slot_conditions:
            datePublished: { value_presence: PRESENT }
        postconditions:
          slot_conditions:
            publishedBy: { required: true }
            doi: { required: true }
      - preconditions:
          slot_conditions:
            datePublished: { value_presence: ABSENT }
        postconditions:
          slot_conditions:
            publishedBy: { value_presence: ABSENT }
      - preconditions:
          slot_conditions:
            datePublished: { value_presence: ABSENT }
        postconditions:
          slot_conditions:
            doi: { value_presence: ABSENT }
slots:
  datePublished:
  publishedBy:
  doi:
```
</details>

<details>
<summary><code>case2_any_field_keyed.yaml</code></summary>

```yaml
id: https://example.org/case2_any_field_keyed
name: case2_any_field_keyed
prefixes:
  linkml: https://w3id.org/linkml/
imports:
  - linkml:types
default_range: string

# Case 2: the same fields, but required conditioned on ANY of them being set.
# If any of datePublished / publishedBy / doi is present, all three are required.

classes:
  Dandiset:
    slots:
      - datePublished
      - publishedBy
      - doi
    rules:
      - preconditions:
          any_of:
            - slot_conditions:
                datePublished:
                  value_presence: PRESENT
            - slot_conditions:
                publishedBy:
                  value_presence: PRESENT
            - slot_conditions:
                doi:
                  value_presence: PRESENT
        postconditions:
          slot_conditions:
            datePublished:
              required: true
            publishedBy:
              required: true
            doi:
              required: true

slots:
  datePublished:
  publishedBy:
  doi:
```
</details>

<details>
<summary><code>case1_naive_coherence_bug.yaml</code> (buggy single-rule draft direction, kept to demonstrate the gotcha)</summary>

```yaml
id: https://example.org/case1_naive_coherence_bug
name: case1_naive_coherence_bug
prefixes:
  linkml: https://w3id.org/linkml/
imports:
  - linkml:types
default_range: string

# Case 1: fields required conditioned on ONE designated field being set,
# plus the coherence direction: when datePublished is absent, the other two
# must be absent too.
# When datePublished is present, publishedBy and doi become required.
# When datePublished is absent, publishedBy and doi must be absent.

classes:
  Dandiset:
    slots:
      - datePublished
      - publishedBy
      - doi
    rules:
      - preconditions:
          slot_conditions:
            datePublished:
              value_presence: PRESENT
        postconditions:
          slot_conditions:
            publishedBy:
              required: true
            doi:
              required: true
      - preconditions:
          slot_conditions:
            datePublished:
              value_presence: ABSENT
        postconditions:
          slot_conditions:
            publishedBy:
              value_presence: ABSENT
            doi:
              value_presence: ABSENT

slots:
  datePublished:
  publishedBy:
  doi:
```
</details>

<details>
<summary>Test metadata (8 instances, <code>metadata/*.yaml</code>)</summary>

```yaml
# all_absent__valid_draft.yaml  -> valid under both keyings
{}

# all_present__valid_published.yaml  -> valid under both keyings
datePublished: "2021-01-01T00:00:00"
publishedBy: "urn:uuid:publisher"
doi: "10.48324/dandi.000001/0.210101.0000"

# only_datePublished__published_missing_publishedBy_and_doi.yaml  -> invalid under both
datePublished: "2021-01-01T00:00:00"

# datePublished_and_publishedBy__published_missing_doi.yaml  -> invalid under both
datePublished: "2021-01-01T00:00:00"
publishedBy: "urn:uuid:publisher"

# datePublished_and_doi__published_missing_publishedBy.yaml  -> invalid under both
datePublished: "2021-01-01T00:00:00"
doi: "10.48324/dandi.000001/0.210101.0000"

# only_publishedBy__stray_no_datePublished.yaml  -> the distinguishing case
publishedBy: "urn:uuid:publisher"

# only_doi__stray_no_datePublished.yaml  -> the distinguishing case
doi: "10.48324/dandi.000001/0.210101.0000"

# publishedBy_and_doi__stray_no_datePublished.yaml  -> invalid under both correct keyings
publishedBy: "urn:uuid:publisher"
doi: "10.48324/dandi.000001/0.210101.0000"
```
</details>

<details>
<summary><code>case1_models.py</code> (from <code>gen-pydantic case1_datePublished_keyed.yaml</code>)</summary>

Note the `rules` survive only as an inert `linkml_meta` ClassVar; nothing enforces them.

```python
from __future__ import annotations

import re
import sys
from datetime import (
    date,
    datetime,
    time
)
from decimal import Decimal
from enum import Enum
from typing import (
    Any,
    ClassVar,
    Literal,
    Optional,
    Union
)

from pydantic import (
    BaseModel,
    ConfigDict,
    Field,
    RootModel,
    SerializationInfo,
    SerializerFunctionWrapHandler,
    field_validator,
    model_serializer
)


metamodel_version = "1.11.0"
version = "None"


class ConfiguredBaseModel(BaseModel):
    model_config = ConfigDict(
        serialize_by_alias = True,
        validate_by_name = True,
        validate_assignment = True,
        validate_default = True,
        extra = "forbid",
        arbitrary_types_allowed = True,
        use_enum_values = True,
        strict = False,
    )


class LinkMLMeta(RootModel):
    root: dict[str, Any] = {}
    model_config = ConfigDict(frozen=True)

    def __getattr__(self, key:str):
        return getattr(self.root, key)

    def __getitem__(self, key:str):
        return self.root[key]

    def __setitem__(self, key:str, value):
        self.root[key] = value

    def __contains__(self, key:str) -> bool:
        return key in self.root


linkml_meta = LinkMLMeta({'default_prefix': 'https://example.org/case1_datePublished_keyed/',
     'default_range': 'string',
     'id': 'https://example.org/case1_datePublished_keyed',
     'imports': ['linkml:types'],
     'name': 'case1_datePublished_keyed',
     'prefixes': {'linkml': {'prefix_prefix': 'linkml',
                             'prefix_reference': 'https://w3id.org/linkml/'}},
     'source_file': 'case1_datePublished_keyed.yaml'} )


class Dandiset(ConfiguredBaseModel):
    linkml_meta: ClassVar[LinkMLMeta] = LinkMLMeta({'from_schema': 'https://example.org/case1_datePublished_keyed',
         'rules': [{'postconditions': {'slot_conditions': {'doi': {'name': 'doi',
                                                                   'required': True},
                                                           'publishedBy': {'name': 'publishedBy',
                                                                           'required': True}}},
                    'preconditions': {'slot_conditions': {'datePublished': {'name': 'datePublished',
                                                                            'value_presence': 'PRESENT'}}}},
                   {'postconditions': {'slot_conditions': {'publishedBy': {'name': 'publishedBy',
                                                                           'value_presence': 'ABSENT'}}},
                    'preconditions': {'slot_conditions': {'datePublished': {'name': 'datePublished',
                                                                            'value_presence': 'ABSENT'}}}},
                   {'postconditions': {'slot_conditions': {'doi': {'name': 'doi',
                                                                   'value_presence': 'ABSENT'}}},
                    'preconditions': {'slot_conditions': {'datePublished': {'name': 'datePublished',
                                                                            'value_presence': 'ABSENT'}}}}]})

    datePublished: Optional[str] = Field(default=None, json_schema_extra = { "linkml_meta": {'domain_of': ['Dandiset']} })
    publishedBy: Optional[str] = Field(default=None, json_schema_extra = { "linkml_meta": {'domain_of': ['Dandiset']} })
    doi: Optional[str] = Field(default=None, json_schema_extra = { "linkml_meta": {'domain_of': ['Dandiset']} })


# Model rebuild
# see https://pydantic-docs.helpmanual.io/usage/models/#rebuilding-a-model
Dandiset.model_rebuild()
```
</details>

<details>
<summary><code>case2_models.py</code> (from <code>gen-pydantic case2_any_field_keyed.yaml</code>) — only the <code>Dandiset</code> class differs</summary>

```python
class Dandiset(ConfiguredBaseModel):
    linkml_meta: ClassVar[LinkMLMeta] = LinkMLMeta({'from_schema': 'https://example.org/case2_any_field_keyed',
         'rules': [{'postconditions': {'slot_conditions': {'datePublished': {'name': 'datePublished',
                                                                             'required': True},
                                                           'doi': {'name': 'doi',
                                                                   'required': True},
                                                           'publishedBy': {'name': 'publishedBy',
                                                                           'required': True}}},
                    'preconditions': {'any_of': [{'slot_conditions': {'datePublished': {'name': 'datePublished',
                                                                                        'value_presence': 'PRESENT'}}},
                                                 {'slot_conditions': {'publishedBy': {'name': 'publishedBy',
                                                                                      'value_presence': 'PRESENT'}}},
                                                 {'slot_conditions': {'doi': {'name': 'doi',
                                                                              'value_presence': 'PRESENT'}}}]}}]})

    datePublished: Optional[str] = Field(default=None, json_schema_extra = { "linkml_meta": {'domain_of': ['Dandiset']} })
    publishedBy: Optional[str] = Field(default=None, json_schema_extra = { "linkml_meta": {'domain_of': ['Dandiset']} })
    doi: Optional[str] = Field(default=None, json_schema_extra = { "linkml_meta": {'domain_of': ['Dandiset']} })
```
</details>

<details>
<summary>Full JSON Schema validation matrix (<code>linkml-validate</code>)</summary>

```
# case1_datePublished_keyed.yaml  = keyed on datePublished, coherence split one-rule-per-field (CORRECT)
# case2_any_field_keyed.yaml      = keyed on ANY of the three fields being present
# case1_naive_coherence_bug.yaml  = keyed on datePublished, coherence as a SINGLE rule (buggy translation)

================================================================
SCHEMA: case1_datePublished_keyed.yaml
================================================================
--- all_absent__valid_draft.yaml
No issues found

--- all_present__valid_published.yaml
No issues found

--- datePublished_and_doi__published_missing_publishedBy.yaml
[ERROR] [.../0] 'publishedBy' is a required property in /

--- datePublished_and_publishedBy__published_missing_doi.yaml
[ERROR] [.../0] 'doi' is a required property in /

--- only_datePublished__published_missing_publishedBy_and_doi.yaml
[ERROR] [.../0] 'publishedBy' is a required property in /
[ERROR] [.../0] 'doi' is a required property in /

--- only_doi__stray_no_datePublished.yaml
[ERROR] [.../0] {'doi': '10.48324/dandi.000001/0.210101.0000'} should not be valid under {'required': ['doi']} in /

--- only_publishedBy__stray_no_datePublished.yaml
[ERROR] [.../0] {'publishedBy': 'urn:uuid:publisher'} should not be valid under {'required': ['publishedBy']} in /

--- publishedBy_and_doi__stray_no_datePublished.yaml
[ERROR] [.../0] {'publishedBy': 'urn:uuid:publisher', 'doi': '...'} should not be valid under {'required': ['publishedBy']} in /
[ERROR] [.../0] {'publishedBy': 'urn:uuid:publisher', 'doi': '...'} should not be valid under {'required': ['doi']} in /

================================================================
SCHEMA: case2_any_field_keyed.yaml
================================================================
--- all_absent__valid_draft.yaml
No issues found

--- all_present__valid_published.yaml
No issues found

--- datePublished_and_doi__published_missing_publishedBy.yaml
[ERROR] [.../0] 'publishedBy' is a required property in /

--- datePublished_and_publishedBy__published_missing_doi.yaml
[ERROR] [.../0] 'doi' is a required property in /

--- only_datePublished__published_missing_publishedBy_and_doi.yaml
[ERROR] [.../0] 'publishedBy' is a required property in /
[ERROR] [.../0] 'doi' is a required property in /

--- only_doi__stray_no_datePublished.yaml
[ERROR] [.../0] 'datePublished' is a required property in /
[ERROR] [.../0] 'publishedBy' is a required property in /

--- only_publishedBy__stray_no_datePublished.yaml
[ERROR] [.../0] 'datePublished' is a required property in /
[ERROR] [.../0] 'doi' is a required property in /

--- publishedBy_and_doi__stray_no_datePublished.yaml
[ERROR] [.../0] 'datePublished' is a required property in /

================================================================
SCHEMA: case1_naive_coherence_bug.yaml
================================================================
--- all_absent__valid_draft.yaml
No issues found

--- all_present__valid_published.yaml
No issues found

--- datePublished_and_doi__published_missing_publishedBy.yaml
[ERROR] [.../0] 'publishedBy' is a required property in /

--- datePublished_and_publishedBy__published_missing_doi.yaml
[ERROR] [.../0] 'doi' is a required property in /

--- only_datePublished__published_missing_publishedBy_and_doi.yaml
[ERROR] [.../0] 'publishedBy' is a required property in /
[ERROR] [.../0] 'doi' is a required property in /

--- only_doi__stray_no_datePublished.yaml
No issues found          <-- WRONG: stray doi slips through

--- only_publishedBy__stray_no_datePublished.yaml
No issues found          <-- WRONG: stray publishedBy slips through

--- publishedBy_and_doi__stray_no_datePublished.yaml
[ERROR] [.../0] {'publishedBy': 'urn:uuid:publisher', 'doi': '...'} should not be valid under {'required': ['publishedBy', 'doi']} in /
```
</details>

<details>
<summary>Pydantic construction matrix (both generated models) — nothing is rejected</summary>

```
# gen-pydantic does NOT emit validators for rules -> rule violations are NOT caught.

================================================================
MODEL: case1_datePublished_keyed
================================================================
all_absent__valid_draft.yaml                                  -> constructed OK
all_present__valid_published.yaml                             -> constructed OK
datePublished_and_doi__published_missing_publishedBy.yaml     -> constructed OK
datePublished_and_publishedBy__published_missing_doi.yaml     -> constructed OK
only_datePublished__published_missing_publishedBy_and_doi.yaml-> constructed OK
only_doi__stray_no_datePublished.yaml                         -> constructed OK
only_publishedBy__stray_no_datePublished.yaml                 -> constructed OK
publishedBy_and_doi__stray_no_datePublished.yaml              -> constructed OK

================================================================
MODEL: case2_any_field_keyed
================================================================
(identical: all 8 instances constructed OK)
```
</details>

candleindark: An [issue](https://github.com/linkml/linkml/issues/3719) has been filed related to the bug we discovered in LinkML in the last response.

candleindark: @yarikoptic prefers the "keyed on ANY of the fields being set".

yarikoptic:
note to myself, 
image
is just tests of dandi-cli whenever https://github.com/dandi/dandi-archive/pull/2866/changes is of the server itself

yarikoptic:
@satra we are thinking to proceed with this transition on how we test for publisher vs draft dandiset or asset, so that we do not breed the models and rather add "rules" (which could be expressed in linkml as well).  Side-effect -- jsonschema validation will be more relaxed for "PublishedDandiset" because they do not translate. After we migrate to linkml with `rules` -- they will be translated into jsonschema. But @candleindark that linkml `rules` do not translate at all or as fully to pydantic for which @candleindark yet to find or file an issue.

What do you @satra think about this transition?

candleindark:
I will add to @yarikoptic comment at https://github.com/dandi/dandi-schema/pull/419#issuecomment-5027114236 is that there actually will not be a JSON schema for `PublishedDandiset` though we will still have `published-dandiset.json` published at https://github.com/dandi/schema upon the next release of `dandi-schema`. In this next release, `dandiset.json` and `published-dandiset.json` will be identical for the Pydantic `PublishedDandiset` class is eliminated in this PR. The variable `PublishedDandiset` is now just an alias pointing to the consolidated `Dandiset` Pydantic model.

satra:
@candleindark and @yarikoptic -  this resonates with what we are doing this week in our group of creating a linkml based registry service of classes, properties, rules, and transforms. in particular, rules can represent validation logic and components allowing for merger of classes as in this PR.  so i'm on board with this transition. we haven't worked through the details of how rules are represented. (cc: @djarecka)

satra:
ps. we will be pulling in dandi's schemas through time into this registry as a use-case. so hopefully after the registry is live, we can use that then to define any future schema changes.

yarikoptic:
Great to see the convergence here @satra -- any pointers/examples of your WiP might be very relevant to align to etc! 

@candleindark could you collate the summary or even may be already set of rules we will need to mimic current validation in python custom validators + that published dandiset/assets validation?

yarikoptic (dismissed):

candleindark:
@yarikoptic Should I up the `DANDI_SCHEMA_VERSION` https://github.com/dandi/dandi-schema/blob/cd26d54d88cb66c21e166e3bd925992fbb49dfe4/dandischema/consts.py#L3

to just "0.7.1"? 

This PR will just move publish related validations into a Pydantic model validator so the corresponding JSON schemas for `Dandiset` and `Asset` will be just be more relaxed. However, `dandiset.json` and `published-dandiset.json` published to https://github.com/dandi/schema will be identical, so are `asset.json` and `published-asset.json`, as explained in https://github.com/dandi/dandi-schema/pull/419#issuecomment-5052169543.

In a followup PR, I will remove the names of `PublishedDandiset` and `PublishedAsset` completely from `dandischema.models`. With that, we will remove the publishing of `published-dandiset.json` and `published-asset.json` to https://github.com/dandi/schema. Should we bump `DANDI_SCHEMA_VERSION` to "0.8.0" then?

yarikoptic:
Not sure if we shouldn't make even this DANDI_SCHEMA_VERSION = "0.8.0" - the diff on schema difficult to grasp on the phone since many changes are moved but also seems some fields changes which might be taken as "breaking" change. Can you review/summarize the diffs ?

candleindark:
Yes, let's bump `DANDI_SCHEMA_VERSION` to `0.8.0`. I generated the four schemas the way `.github/workflows/release.yml` does and diffed them against the published `0.7.0` in `dandi/schema`. Two independent reasons for the bump, and to your other question: nothing in this PR is breaking for existing metadata.

**1. The bump is already due, independently of this PR.** `master` alone already generates schemas that differ from the published `0.7.0`: `sameAs` on `Dandiset` (#364, #370) and `readOnly` on `Asset.access` (#414). Since `0.7.0` is already published, the release workflow's "Test for unversioned changes" step fails today with `DANDI_SCHEMA_VERSION` left at `0.7.0`, with or without #419.

**2. #419 then changes all four files further,** so reusing `0.7.0` would make one version string mean three different things.

**Not breaking.** Every `required` change is a removal, every added property is optional and `readOnly`, and both pattern changes are relaxations. No stored Dandiset or Asset metadata becomes invalid, and the publication rules that leave the JSON Schema are re-enforced in Python by the `datePublished`-gated validator, which `metadata.validate()` always runs after the JSON Schema pass. A minor bump is the right size: `migrate()` has no version-specific transformation past `0.6.0`, so `0.7.0` -> `0.8.0` only rewrites `schemaVersion`.

<details>
<summary>Summary of the diff, published <code>0.7.0</code> -> this branch</summary>

Draft `dandiset.json` and `asset.json` gain schema text only, and `required` is unchanged in both:

| file | change | kind |
| --- | --- | --- |
| `dandiset.json` | `+ doi`, `+ publishedBy`, `+ datePublished`, `+ releaseNotes` | optional, `readOnly` |
| `dandiset.json` | `+ $defs.PublishActivity` | pulled in by `publishedBy`, its only `$ref` |
| `dandiset.json` | `url.description` reworded | doc only |
| `asset.json` | `+ publishedBy`, `+ datePublished` | optional, `readOnly` |
| `asset.json` | `+ $defs.PublishActivity` | as above |

Since all of these are `readOnly`, the Meditor (which hides `readOnly` properties) is unaffected.

`published-dandiset.json` and `published-asset.json` become byte-identical to their draft counterparts:

| file | change | effect |
| --- | --- | --- |
| `published-dandiset.json` | `- required: datePublished, publishedBy, url` | relaxation |
| `published-dandiset.json` | `id.pattern` widened to accept draft ids and lowercase instance prefixes | relaxation |
| `published-dandiset.json` | `doi.default` `""` -> `null`; `datePublished`/`publishedBy`/`url` gain `default: null` | annotation only, `default` is not a constraint |
| `published-dandiset.json` | `title` `"Published Dandiset"` -> `"Dandiset"`, `+ description` | cosmetic |
| `published-asset.json` | `- required: datePublished, publishedBy` | relaxation |
| `published-asset.json` | `id.pattern` (`^dandiasset:...$`) dropped | relaxation |
| `published-asset.json` | `title` `"Published Asset"` -> `"Asset"`, `+ description` | cosmetic |

`context.json` drops the `PublishedAsset` and `PublishedDandiset` term mappings, since `generate_context()` walks the model classes and those two names are now aliases. No stored metadata carries those `schemaKey` values, so this is inert in practice, but it is a vocabulary removal and another reason not to reuse `0.7.0`.

Also present in the diff but coming from `master`, not from this PR: `+ sameAs` on both Dandiset schemas, `access.readOnly` on both Asset schemas, and a reordering of the `LicenseType` enum, which is the non-determinism tracked by #433 rather than a real change.
</details>

<details>
<summary>The one thing worth flagging: <code>published-*.json</code> stops expressing publication requirements</summary>

This is the change most likely to be read as breaking, so to be explicit about what it does and does not cost.

Enforcement is not lost through `dandischema`. `validate()` always runs Pydantic validation after the optional JSON Schema pass, and `PublishedDandiset`/`PublishedAsset` are aliases of `Dandiset`/`Asset`, whose `check_publication_status` re-checks every dropped rule once `datePublished` is set: `publishedBy` presence, `url` presence and pattern, `doi` presence, the `id` patterns, `assetsSummary` file and byte counts, and the sha2-256 digest check. dandi-archive's publish-time call sites pass `schema_key='PublishedDandiset'`/`'PublishedAsset'`, so they keep the full check. Worth knowing too: `validate(..., json_validation=True)` uses the locally generated schema whenever `schema_version == DANDI_SCHEMA_VERSION`, and reads `dandi/schema` only for metadata stamped with an older version, fetching `releases/<that version>/`. Publishing `0.8.0` only adds `releases/0.8.0/` and leaves every existing release directory alone, `0.7.0` included, so a record stamped `0.7.0` still gets validated against the strict `releases/0.7.0/published-dandiset.json`. The relaxation applies to `0.8.0`-stamped metadata onward, not retroactively.

What is lost is enforcement for anyone consuming `published-*.json` standalone, outside `dandischema`: from `0.8.0` on, those files say nothing about publication. That follows from the constraints now being conditional on `datePublished`, which a flat `required` list cannot express. This is an acceptable gap for now, since the plan is to re-encode these requirements as LinkML rules in the LinkML translation, where `gen-json-schema` emits them as `if`/`then` and the published artifact gets the constraint back.
</details>

<details>
<summary>One direction in which the draft schemas get slightly stricter</summary>

"Additive" means added schema text, which is not quite the same as a pure relaxation. Neither asset schema sets `additionalProperties` at all (both Dandiset schemas set it to `true`), so unknown properties have always been accepted and never validated. Giving `publishedBy` and `datePublished` a declared subschema therefore type-checks documents that already carried those names:

| document | published `0.7.0` `asset.json` | this branch |
| --- | --- | --- |
| minimal valid asset | valid | valid |
| `+ "datePublished": 12345` | valid | invalid, `12345 is not of type 'string'` |
| `+ "publishedBy": 12345` | valid | invalid, `not valid under any of the given schemas` |
| `+ "someRandomKey": 12345` | valid | valid |

Nothing is at risk in practice: draft records do not carry these fields, and published records carry values that already had to satisfy the identical subschemas in `published-*.json`. Noting it for completeness, since it is the only place where this PR makes a schema stricter.
</details>

yarikoptic:
dandi-cli fails due to "expected"

> dandi.exceptions.SchemaVersionError: Server uses older incompatible schema version 0.7.0; client supports 0.8.0.

I now - re-wondering if "both" in our matrix should mean that server should have got it updated too, and thus potentially it should have all being peachy  @candleindark ?

edit, e.g. we do have one dandi-cli on linux (where we can test using docker) passing

image

candleindark:
> dandi-cli fails due to "expected"

The failures are expected due the way the tests are set up. However, the set up of those tests are buggy.

> I now - re-wondering if "both" in our matrix should mean that server should have got it updated too, and thus potentially it should have all being peachy  @candleindark ?

Particularly, "both", and "server", failed to set the run environment of the `dandi-archive` to use this version of `dandi-schema` because uv re-syncs the run environment using `uv.lock`

yarikoptic:
I see, so we kinda broke our testing setup when jump over to use `uv.lock` was introduced.  Do you see how to address it easily/quickly?

candleindark:
> I see, so we kinda broke our testing setup when jump over to use `uv.lock` was introduced. Do you see how to address it easily/quickly?

Yes. I will push it out and see how it works.

tools/api-with-schema.Dockerfile
```diff
@@ -1,3 +1,11 @@
...
-RUN uv pip install -e /opt/dandischema[dev]
+RUN uv pip install -e /opt/dandischema
+
+# The base image's `manage.py` runs via `#!/usr/bin/env -S uv run`, and `uv run`
+# syncs `/opt/django/.venv` to `uv.lock` before executing, which would reinstall
+# the `dandischema` release pinned by `dandi-archive` over the one installed
+# above. That sync must be skipped for this image to run the schema version
+# under test; skipping it is safe because the base image's `uv sync` has already
+# installed every other dependency.
+ENV UV_NO_SYNC=1
```
candleindark, line 11: @yarikoptic This eliminate two CI test failures.

candleindark:
@yarikoptic I eliminated two test failures in https://github.com/dandi/dandi-schema/pull/419#discussion_r3662027702. The remaining three test failures, not including the codecov ones, can't be eliminate in this PR and are expected (see "Expected `test-dandi-cli` failures" in PR description for details). After https://github.com/dandi/dandi-cli/pull/1894 is merged, one more CI failures can be eliminated.

This PR is good to merge. However, I want you to take a second look, particular at the changes I put in after your last approval. 

Note: The PR description has been updated with the latest changes.

yarikoptic:
oh -- what to add/do in https://github.com/dandi/dandi-schema/blob/master/dandischema/metadata.py#L303 `migrate` so  we could upgrade or downgrade for the changes in this PR? (or may be forgotten for prior PRs for this schema version boost)

candleindark:
@yarikoptic Nothing needs to be added to `migrate`, as far as I can tell, neither for this PR nor for the earlier PRs that this version bump carries.

**This PR** neither adds nor removes a restriction on existing metadata. A draft record carries no `doi` / `publishedBy` / `datePublished` / `releaseNotes`, which is exactly what the draft branch of the gated validator requires; a published record carries all of them and satisfies the same rules the `Published*` classes enforced. A valid `0.7.0` instance is therefore already a valid `0.8.0` instance, and the only thing `migrate` has to do is restamp `schemaVersion`, which it already does unconditionally.

**The earlier PRs** contribute exactly three deltas between the published `0.7.0` and `master`, and none of them needs migration either:

| change | why no migration step |
| --- | --- |
| `sameAs` added to `Dandiset` (#364, #370) | a new optional property, not in `required`, so its absence from older metadata is valid |
| `access.readOnly: true` on `Asset` (#414) | `readOnly` is a JSON Schema annotation rather than a constraint, and it affects no instance. `migrate` handles Dandiset metadata only, so this is outside its scope regardless |
| `LicenseType` enum members reordered | the same member set in a different order, an artifact of #433 rather than a schema change |

<details>
<summary>Checked rather than only reasoned about: <code>0.7.0</code> → <code>0.8.0</code> round trip</summary>

For both a draft and a published Dandiset, the script below asserts the instance is valid under the **published** `0.7.0` schema in `dandi/schema`, runs `migrate()`, reports every field the migration changed, and then validates the result against this branch's models and JSON Schema:

```
[draft] valid against published 0.7.0 dandiset.json
[draft] migrate() -> schemaVersion=0.8.0
[draft] fields changed by migrate(): ['schemaVersion']
[draft] validate() against 0.8.0 models + JSON Schema: OK

[published] valid against published 0.7.0 published-dandiset.json
[published] migrate() -> schemaVersion=0.8.0
[published] fields changed by migrate(): ['schemaVersion']
[published] validate() against 0.8.0 models + JSON Schema: OK
```

```python
"""Round-trip real 0.7.0-shaped Dandiset metadata through this branch's migrate()."""
import json
from copy import deepcopy
from pathlib import Path

import jsonschema

from dandischema.consts import DANDI_SCHEMA_VERSION
from dandischema.metadata import migrate, validate

PUB = Path("<clone of dandi/schema>/releases/0.7.0")
INSTANCE = "DANDI"

draft = {
    "schemaKey": "Dandiset",
    "schemaVersion": "0.7.0",
    "identifier": f"{INSTANCE}:999999",
    "id": f"{INSTANCE}:999999/draft",
    "version": "draft",
    "name": "testing dataset",
    "description": "testing",
    "contributor": [
        {
            "name": "last name, first name",
            "email": "someone@dandiarchive.org",
            "roleName": ["dcite:ContactPerson"],
            "schemaKey": "Person",
        }
    ],
    "license": ["spdx:CC-BY-4.0"],
    "citation": "Last, first (2021). Test citation.",
    "assetsSummary": {
        "schemaKey": "AssetsSummary",
        "numberOfBytes": 0,
        "numberOfFiles": 0,
        "dataStandard": [{"schemaKey": "StandardsType", "name": "NWB"}],
        "approach": [{"schemaKey": "ApproachType", "name": "electrophysiology"}],
        "measurementTechnique": [
            {"schemaKey": "MeasurementTechniqueType",
             "name": "two-photon microscopy technique"}
        ],
        "species": [{"schemaKey": "SpeciesType", "name": "Human"}],
    },
    "manifestLocation": [
        "https://api.dandiarchive.org/api/dandisets/999999/versions/draft/assets/"
    ],
    "url": "https://dandiarchive.org/dandiset/999999/draft",
}

published = deepcopy(draft)
published.update(
    {
        "id": f"{INSTANCE}:999999/0.0.0",
        "version": "0.0.0",
        "url": "https://dandiarchive.org/dandiset/999999/0.0.0",
        "doi": "10.80507/dandi.999999/0.0.0",
        "datePublished": "2021-01-01T00:00:00+00:00",
        "publishedBy": {
            "schemaKey": "PublishActivity",
            "id": "urn:uuid:08fffc59-9f1b-44d6-8e02-6729d266d1b6",
            "name": "Publish",
            "startDate": "2021-01-01T00:00:00+00:00",
            "endDate": "2021-01-01T00:00:00+00:00",
            "wasAssociatedWith": [
                {
                    "schemaKey": "Software",
                    "id": "urn:uuid:9267d2e1-4a37-463b-a2b6-3d1c1cd39c8f",
                    "identifier": "RRID:SCR_017571",
                    "name": "DANDI API",
                    "version": "0.1.0",
                }
            ],
        },
    }
)
published["assetsSummary"].update({"numberOfBytes": 1, "numberOfFiles": 1})

for label, obj, pub_file in [
    ("draft", draft, "dandiset.json"),
    ("published", published, "published-dandiset.json"),
]:
    # 1. Is it genuinely valid under the *published* 0.7.0 JSON Schema?
    schema = json.loads((PUB / pub_file).read_text())
    jsonschema.validate(obj, schema)
    print(f"[{label}] valid against published 0.7.0 {pub_file}")

    # 2. migrate() to this branch's version
    out = migrate(obj, skip_validation=True)
    print(f"[{label}] migrate() -> schemaVersion={out['schemaVersion']}")

    # 3. Does the migrated object differ from the input in anything but schemaVersion?
    delta = {k: (obj.get(k), out.get(k)) for k in set(obj) | set(out) if obj.get(k) != out.get(k)}
    print(f"[{label}] fields changed by migrate(): {sorted(delta)}")

    # 4. Does it validate under this branch (models + local JSON Schema)?
    validate(out, schema_key="Dandiset", json_validation=True)
    print(f"[{label}] validate() against {DANDI_SCHEMA_VERSION} models + JSON Schema: OK\n")
```
</details>

The one change `migrate` did need is already in the PR: `"0.7.0"` is added to `ALLOWED_INPUT_SCHEMAS`, without which `migrate` would reject the very metadata that is out in the wild today.

Downgrade is a separate matter. `migrate` rejects any target below the instance's own version, and `ALLOWED_TARGET_SCHEMAS` holds only `DANDI_SCHEMA_VERSION`, so there is no downgrade path to extend. Adding one is #343.

candleindark:
@yarikoptic Preferably, let's merge https://github.com/dandi/dandi-schema/pull/434 before this PR is merged and released so that we don't have to adjust the `DANDI_SCHEMA_VERSION` multiple times.

candleindark:
Updated the PR description to reflect that the two `tools/api-with-schema.Dockerfile` CI changes (adding `ENV UV_NO_SYNC=1` and dropping the undeclared `dev` extra) were split out into #435 and merged to `master`. They are no longer part of this PR's diff. This branch has been rebased onto the post-#435 `master` and inherits that Dockerfile fix from there, so the "Expected `test-dandi-cli` failures" explanation still holds.

PR #422, opened 2026-06-17
Remove the `missing_ok` parameter from `metadata.validate()`

### LINKED ISSUE #421: Remove the `missing_ok` parameter from `metadata.validate()`
## Summary

Remove the `missing_ok` parameter from `dandischema.metadata.validate()` (and the supporting filtering logic in `_validate_obj_json()` and `validate()`). The core reason is that `missing_ok` is a brittle feature that does not actually do what it claims, suppress "missing field" errors, and it is unused outside this repo's own tests. There is also no current justification for treating missing-field errors differently from any other validation error.

## Why `missing_ok` doesn't work

`missing_ok` tries to suppress "missing field" errors in two separate layers, and neither is sound.

**When a model enforces a required field through a custom validator, `missing_ok` fails outright.** This is the most important case, because dandischema's models do exactly this.

- At the Pydantic layer, `missing_ok` filters errors whose `type == "missing"`. But a required-field constraint implemented with a custom validator (for example a `model_validator`) surfaces the absent field as a `value_error`, not `missing`, so it is never filtered. Checking for `type == "missing"` therefore cannot detect all missing-value errors.
- At the JSON Schema layer there is nothing to suppress in the first place: nothing about a custom validator is translated into JSON Schema, so the constraint does not even exist there.

A missing field enforced this way is reported with `type == "value_error"`, not `"missing"`, so the Pydantic-layer filter never suppresses it.

<details>
<summary>Runnable example: a field required conditionally via a <code>model_validator</code></summary>

```python
from typing_extensions import Self

from pydantic import BaseModel, model_validator, ValidationError


class Foo(BaseModel):
    x: int | None = None
    y: int | None = None

    @model_validator(mode="after")
    def conditionally_require_y(self) -> Self:
        if self.x is not None and self.y is None:
            raise ValueError("y is required when x is provided")
        return self


try:
    # `y` is missing but not marked as `"missing"` because the validation is
    # done by a custom validator. Checking for the error type of `"missing"` in the
    # Pydantic layer cannot ensure all errors due to a missing value are detected.
    Foo(x=1)
except ValidationError as e:
    errs = e.errors()
    assert len(errs) == 1
    assert errs[0]["type"] != "missing"  # it is "value_error", so missing_ok would not filter it
```

</details>

**Even when a model uses no custom validators, the JSON Schema half is brittle.** To suppress missing-field errors there, it substring-matches the `jsonschema` message `"is a required property"`. That message text is not a stable, documented contract, so it can silently break on a dependency upgrade. (The Pydantic half's [`type == "missing"`](https://pydantic.dev/docs/validation/latest/api/pydantic-core/pydantic_core/#pydantic_core.ErrorDetails.type) check is at least stable and documented, so the two halves don't even rest on the same footing.)

So `missing_ok` cannot reliably "ignore missing fields": it misses custom-validator-enforced fields entirely, and for the cases it does handle it rests on brittle string matching.

## Why removing it is safe

No component in the core DANDI suite passes `missing_ok`; the only `missing_ok=True` call sites are this repo's own tests. A GitHub code search plus a survey of `dandischema` dependents found no external caller of `validate(..., missing_ok=...)`.

<details>
<summary>Usage survey</summary>

- `dandi-archive`'s `validate(...)` calls do not pass `missing_ok` (so it defaults to `False`):

  ```
  dandiapi/api/services/publish/__init__.py:191:  validate(new_version.metadata, schema_key='PublishedDandiset', json_validation=True)
  dandiapi/api/services/metadata/__init__.py:65:   validate(metadata, schema_key='PublishedAsset', json_validation=True)
  dandiapi/api/services/metadata/__init__.py:149:  validate(..., schema_key='PublishedDandiset', json_validation=True)
  ```

- `dandi-cli` does import and call `dandischema.metadata.validate` (in `dandi/tests/test_metadata.py`, e.g. `validate(data_as_dict)`), but never passes `missing_ok`.

- The only `missing_ok=True` call sites anywhere are in this repo's own tests (`dandischema/tests/test_metadata.py`).

- A GitHub code search scoped to `dandischema` returns a single hit for `missing_ok`, this repo's own test file; a scan of `dandischema` dependents (the `dandi-archive` fork `lincbrain` included) surfaced no other caller.

</details>

## No justification for treating missing-field errors specially

A missing required field is just one kind of validation failure. One can imagine a workflow that would justify suppressing them, but no such workflow exists in the current codebase, and the current implementation is haphazard and hard to maintain: two separate filtering mechanisms, in two layers, keyed off different signals (one of them brittle).

## Proposed change

Drop the `missing_ok` parameter from `validate()`, delete the now-trivial `_validate_obj_json()` helper (once `missing_ok` is gone it is a pure pass-through to `validate_json`, so its callers can call `validate_json` directly), and remove the associated tests.

`missing_ok` is part of the public `validate()` signature and `dandischema` is a published package, so this is a breaking API change. Given that the survey above found no caller passing `missing_ok` outside this repo's tests, the proposal is a hard removal rather than a deprecation period.

<details>
<summary>The two filtering branches (current implementation)</summary>

```python
# dandischema/metadata.py: JSON Schema layer (fragile substring match)
def _validate_obj_json(instance, validator, *, missing_ok=False):
    try:
        validate_json(instance, validator)
    except JsonschemaValidationError as e:
        if missing_ok:
            remaining_errs = [
                err for err in e.errors if "is a required property" not in err.message
            ]
            ...

# dandischema/metadata.py: Pydantic layer (stable type check, but does not catch
# missing fields enforced by custom validators)
for el in exc.errors():
    if not missing_ok or el["type"] != "missing":
        messages.append(el)
```
</details>

candleindark:
Closes #421.

Removes the `missing_ok` parameter from `dandischema.metadata.validate()`. As detailed in #421, `missing_ok` is a brittle feature that does not reliably suppress "missing field" errors (it misses required fields enforced by custom validators, and otherwise relies on substring-matching a `jsonschema` message) and is unused outside this repo's own tests.

## Changes

- Remove the `missing_ok` parameter from `validate()` and its error-filtering branch.
- Delete the `_validate_obj_json()` helper, which became a pure pass-through to `validate_json()` once `missing_ok` was gone; its four callers now call `validate_json()` directly.
- Remove the associated tests (`test_missing_ok`, `test_missing_ok_error`, and the `TestValidateObjJson` class, whose subject is gone; `validate_json` is already covered in `test_utils.py`).
- Correct the `Raises` section of `validate()`'s docstring: drop the stale clause describing the `ValueError` raised by the removed `missing_ok` error-filtering path, and name the exceptions actually raised on validation failure (`JsonschemaValidationError` and `PydanticValidationError`) instead of a generic `ValidationError`.

This is a breaking change to the public `validate()` signature. The usage survey in #421 found no caller passing `missing_ok` outside this repo's tests, so this is a hard removal rather than a deprecation.

## Background: why `missing_ok` existed at all

The parameter was introduced speculatively in #48 (June 2021) under the name `for_post` and was never adopted by any caller in dandi-schema, dandi-cli, or dandi-archive over its entire five-year lifetime; the draft-vs-published leniency it was meant to provide is served instead by the lenient `Dandiset`/`Asset` vs. strict `PublishedDandiset`/`PublishedAsset` model split.

<details>
<summary>Diffstat</summary>

```
 dandischema/metadata.py            |  67 +++----------
 dandischema/tests/test_metadata.py | 201 +------------------------------------
 2 files changed, 16 insertions(+), 252 deletions(-)
```

</details>

## Test plan

- [x] `tox -e typing,lint`
- [x] `python -m pytest dandischema` (260 passed, 17 skipped; skips are network/instance-name gated)

## Release Notes

The `missing_ok` parameter of the `validate()` function in the `dandischema.metadata` module has been removed. Any call passing `missing_ok` (regardless of its value) must be updated to no longer pass it.

### DIFF
--- a/dandischema/metadata.py
+++ b/dandischema/metadata.py
@@ -16,7 +16,7 @@ from .consts import (
     ALLOWED_VALIDATION_SCHEMAS,
     DANDI_SCHEMA_VERSION,
 )
-from .exceptions import JsonschemaValidationError, PydanticValidationError
+from .exceptions import PydanticValidationError
 from . import models
 from .utils import (
     TransitionalGenerateJsonSchema,
@@ -151,47 +151,16 @@ def publish_model_schemata(releasedir: Union[str, Path]) -> Path:
     return vdir
 
 
-def _validate_obj_json(
-    instance: Any, validator: JsonschemaValidator, *, missing_ok: bool = False
-) -> None:
-    """
-    Validate a data instance using a jsonschema validator with an option to filter out
-    errors related to missing required properties
-
-    :param instance: The data instance to validate
-    :param validator: The JSON schema validator to use
-    :param missing_ok: Indicates whether to filter out errors related to missing
-        required properties
-    :raises JsonschemaValidationError: If the metadata instance is invalid, and there
-        are errors detected in the validation, optionally discounting errors
-        related to missing required properties. An instance of this exception containing
-        a list of `jsonschema.exceptions.ValidationError` instances representing all the
-        (remaining) errors detected in the validation
-    """
-    try:
-        validate_json(instance, validator)
-    except JsonschemaValidationError as e:
-        if missing_ok:
-            remaining_errs = [
-                err for err in e.errors if "is a required property" not in err.message
-            ]
-            # Raise an exception only if there are errors left after filtering
-            if remaining_errs:
-                raise JsonschemaValidationError(remaining_errs) from e
-        else:
-            raise e
-
-
 def _validate_dandiset_json(data: dict, schema_dir: Union[str, Path]) -> None:
     with Path(schema_dir, "dandiset.json").open() as fp:
         schema = json.load(fp)
-    _validate_obj_json(data, dandi_jsonschema_validator(schema))
+    validate_json(data, dandi_jsonschema_validator(schema))
 
 
 def _validate_asset_json(data: dict, schema_dir: Union[str, Path]) -> None:
     with Path(schema_dir, "asset.json").open() as fp:
         schema = json.load(fp)
-    _validate_obj_json(data, dandi_jsonschema_validator(schema))
+    validate_json(data, dandi_jsonschema_validator(schema))
 
 
 @cache
@@ -273,7 +242,6 @@ def validate(
     obj: dict,
     schema_version: Optional[str] = None,
     schema_key: Optional[str] = None,
-    missing_ok: bool = False,
     json_validation: bool = False,
 ) -> None:
     """Validate object using pydantic
@@ -287,9 +255,6 @@ def validate(
     schema_key: str, optional
       Name of the schema key to be used, if not specified, `schemaKey` of the
       object will be consulted
-    missing_ok: bool, optional
-      This flag allows checking if all fields have appropriate values but ignores
-      missing fields. A `ValueError` is raised with the list of all errors.
     json_validation: bool, optional
       If set to True, `obj` is first validated against the corresponding jsonschema.
 
@@ -324,17 +289,12 @@ def validate(
                     "using json schema for older versions"
                 )
             jvalidator = _get_jsonschema_validator(schema_version, schema_key)
-        _validate_obj_json(obj, jvalidator, missing_ok=missing_ok)
+        validate_json(obj, jvalidator)
     klass = getattr(models, schema_key)
     try:
         klass(**obj)
     except pydantic.ValidationError as exc:
-        messages = []
-        for el in exc.errors():
-            if not missing_ok or el["type"] != "missing":
-                messages.append(el)
-        if messages:
-            raise PydanticValidationError(messages)  # type: ignore[arg-type]
+        raise PydanticValidationError(exc.errors())  # type: ignore[arg-type]
 
 
 def migrate(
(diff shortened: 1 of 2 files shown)


candleindark:
@yarikoptic @satra An analysis of why the `missing_ok` param is in existence at all is added to the description of this PR.

yarikoptic:
It changes API so definitely not `patch` level.  Adjusted the label.  Please ping @satra on slack and if no show stoppers discovered -- let's merge

candleindark:
@yarikoptic Are you sure you want to go with the `minor` label at this point? I thought `minor` is considered `major`, and `patch` is considered `minor` in this repo.

yarikoptic:
ideally PR description should have Release changelog for `auto` to note that it is breaking change but we do not expect to get broken

candleindark:
> ideally PR description should have Release changelog for `auto` to note that it is breaking change but we do not expect to get broken

A Release Notes section is added to the PR description for `auto` to pick up the needed info for changelog. However, section only supplies free-text content that gets surfaced at the top of the changelog — it plays no role in classifying a PR as a breaking change. That classification comes exclusively from the major [label](https://intuit.github.io/auto/docs/configuration/autorc#labels).

candleindark:
@satra has given an [OK](https://dandiarchive.slack.com/archives/GMRLT5RQ8/p1784609736346329?thread_ts=1784564759.119669&cid=GMRLT5RQ8) to move forward. @yarikoptic, let's merge this PR.

PR #428, opened 2026-07-14
test: fix bugs and tighten checks in `test_duplicate_classes`

candleindark:
## Summary
- Fix a latent bug in `test_duplicate_classes` where `name` was only assigned inside the `_ldmeta` "nskey" branch, so the `else` branch could reuse a stale (or unbound) `name` left over from a previous iteration.
- Replace the hardcoded `"dandi"` qname prefix with the existing `DANDI_NSKEY` constant, drop dead code (`Optional` on the `qnames` dict and its unreachable `None` check), and stop special-casing the iteration order of `CommonModel` and `BaseType`, since the conflict-resolution logic already handles either registration order.
- Tighten the conflicting-qname exception checks to compare both class names as a set, instead of only checking that the newly seen class's name was one of the two allowed names, so each exception verifies both classes involved are exactly the expected pair.

## Test plan
- [x] `tox -e py3 -- dandischema/tests/test_models.py -v`
- [x] `tox -e lint,typing`

### DIFF
--- a/dandischema/tests/test_models.py
+++ b/dandischema/tests/test_models.py
@@ -572,25 +572,25 @@ def test_duplicate_classes() -> None:
             elif issubclass(t, klass):
                 qnames[qname] = klass
                 return
-            if qname == "dandi:repository" and klass.__name__ in (
+            if qname == "dandi:repository" and {t.__name__, klass.__name__} == {
                 "Resource",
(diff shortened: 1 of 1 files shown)


yarikoptic:
thank you @candleindark !

PR #429, opened 2026-07-15
fix: guard against empty loc tuples; drop redundant list comprehensions

candleindark:
## Summary
- Fix a latent `IndexError`: `pydantic.ValidationError`'s `ErrorDetails.loc` is a tuple that can be empty for model-level (root) validator errors, so unconditionally indexing it at position `0` is unsafe; two test sites now guard against this.
- Drop redundant list comprehensions passed to `any()`, `all()`, and `set()`, which all consume an iterable directly, so wrapping a generator expression in a list comprehension first was unnecessary.

## Test plan
- [x] `tox -e py3` passes
- [x] `tox -e lint,typing` passes

### DIFF
--- a/dandischema/tests/test_metadata.py
+++ b/dandischema/tests/test_metadata.py
@@ -316,7 +316,7 @@ def test_requirements(
         validate(obj, schema_key=schema_key)
     with pytest.raises(PydanticValidationError) as exc:
         validate(obj, schema_key=schema_key, schema_version=DANDI_SCHEMA_VERSION)
-    assert set([el["loc"][0] for el in exc.value.errors]) == missingfields
+    assert set(el["loc"][0] for el in exc.value.errors) == missingfields
 
 
 @pytest.mark.parametrize(
--- a/dandischema/tests/test_metadata.py
+++ b/dandischema/tests/test_metadata.py
@@ -316,7 +316,10 @@ def test_requirements(
         validate(obj, schema_key=schema_key)
     with pytest.raises(PydanticValidationError) as exc:
         validate(obj, schema_key=schema_key, schema_version=DANDI_SCHEMA_VERSION)
-    assert set(el["loc"][0] for el in exc.value.errors) == missingfields
+    assert (
+        set(loc[0] if (loc := el["loc"]) else None for el in exc.value.errors)
+        == missingfields
+    )
 
 
 @pytest.mark.parametrize(
--- a/dandischema/tests/test_models.py
+++ b/dandischema/tests/test_models.py
@@ -150,10 +150,8 @@ def test_asset_digest() -> None:
             contentSize=100, encodingFormat="nwb", digest=digest_model, path="/"
         )
     assert any(
-        [
-            "Digest must have an appropriate dandi-etag value." in val
-            for val in set([el["msg"] for el in exc.value.errors()])
-        ]
+        "Digest must have an appropriate dandi-etag value." in val
+        for val in set(el["msg"] for el in exc.value.errors())
     )
     digest = 32 * "a" + "-1"
     digest_model = {models.DigestType.dandi_etag: digest}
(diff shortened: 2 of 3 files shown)


yarikoptic:
I do not see any fix to code, only to tests, but then the question how they were passing before and why we need guards in the tests -- may be they were testing exactly that there is a value there?!

dandischema/tests/test_metadata.py
```diff
@@ -316,7 +316,10 @@ def test_requirements(
         validate(obj, schema_key=schema_key)
     with pytest.raises(PydanticValidationError) as exc:
         validate(obj, schema_key=schema_key, schema_version=DANDI_SCHEMA_VERSION)
-    assert set([el["loc"][0] for el in exc.value.errors]) == missingfields
```
candleindark, line 319: This is the error. There is no guarantee that `el["loc"]` is not empty. This is the same bug in https://github.com/dandi/dandi-archive/pull/2853. We haven't encountered it because all of the errors we've encountered are field level errors and have a non-empty tuple as `loc`. 

Without the fix, the test can failed because the test code is accessing `loc` incorrectly not because the intended assertion is false.

dandischema/tests/test_models.py
```diff
@@ -493,7 +479,7 @@ def test_dandimeta_1(base_dandiset_metadata: dict[str, Any]) -> None:
         if expected_errors[err_loc].msg is not None:
             assert err["msg"] == expected_errors[err_loc].msg
 
-    assert set([el["loc"][0] for el in exc.value.errors()]) == {
```
candleindark, line 496: This is the same error again. There is no guarantee that `el["loc"]` is not empty. This is the same bug in https://github.com/dandi/dandi-archive/pull/2853. We haven't encountered it because all of the errors we've encountered are field level errors and have a non-empty tuple as `loc`. 

Without the fix, the test can failed because the test code is accessing `loc` incorrectly not because the intended assertion is false.

candleindark (commented):
@yarikoptic I marked the errors. The errors were in the tests not in the operational code. You may remove the bug label if you think that is not appropriate.

Other changes are more stylistic improvements.

dandischema/tests/test_metadata.py
yarikoptic: So the tests started to fail if we run on linkml'ified version, or when?

In effect, currently test "asserts" that there is `el["loc"]` .  With your changes it stops asserting that.

Don't we expect to have `"loc"`  field here, or when do we?

candleindark: > So the tests started to fail if we run on linkml'ified version, or when?

This is not LinkML specific. It has to do with the structure of the errors returned by Pydantic, specifically, the `ErrorDetails` objects, https://pydantic.dev/docs/validation/latest/errors/errors/#_top. You will get an `IndexError` for accessing the `0` index of `el["loc"]` once you encounter a `ValidationError` raised from a model validator in the standard way as in https://pydantic.dev/docs/validation/latest/concepts/validators/#model-after-validator since `el["loc"]` will be an empty tuple in such a case.

> In effect, currently test "asserts" that there is `el["loc"]` . With your changes it stops asserting that.

There is always `el["loc"]` per the contract at https://pydantic.dev/docs/validation/latest/errors/errors/#_top. There is just no guarantee that `el["loc"]` is non-empty. The fix handles the case the `el["loc"]` is empty and allowed the intended insertion to go forward.

> Don't we expect to have `"loc"` field here, or when do we?

We do, just not non-empty.

For details, checkout https://github.com/pydantic/pydantic-core/blob/383eb95a19433754c0cecf7025b50c26b6d97a36/python/pydantic_core/__init__.py#L74-L97

yarikoptic:
Ok, I agree to agree since in effect we will still get the same -- tests would not pass if there is an empty `el["loc"]`

PR #434, opened 2026-07-29
fix: sort `LicenseType` members by value for reproducible schemas

### LINKED ISSUE #433: LicenseType member order is non-deterministic, making generated JSON Schemas unreproducible
## Summary

The member order of `dandischema.models.LicenseType`, and hence the order of the `LicenseType.enum` array in every generated JSON Schema, varies from process to process, so two runs of `tools/pubschemata.py` on the same commit can emit different files.

`Config.licenses` is a `set` ([`conf.py:162`](https://github.com/dandi/dandi-schema/blob/0.12.1/dandischema/conf.py#L162)) and the enum is built by iterating it ([`models.py:122-125`](https://github.com/dandi/dandi-schema/blob/0.12.1/dandischema/models.py#L122-L125)). `License` is itself an `Enum`, and `Enum.__hash__` is `hash(self._name_)`, i.e. the hash of a `str`, so the iteration order follows `PYTHONHASHSEED`, which CPython randomizes per process. Affects 0.12.0 onward, where the `licenses` field first shipped (cf0293ef).

<details>
<summary>Reproducer</summary>

Pinning `PYTHONHASHSEED` turns the flakiness into a deterministic difference between runs. With released `dandischema` 0.12.1 on Python 3.10.15:

```bash
#!/bin/bash
set -eu
PYTHON=${1:-python}

echo "== LicenseType member order"
for seed in 0 1 2 3 4 5; do
    printf 'PYTHONHASHSEED=%s -> ' "$seed"
    PYTHONHASHSEED=$seed "$PYTHON" -c \
        'from dandischema.models import LicenseType; print([m.value for m in LicenseType])'
done

echo
echo "== the same ordering in the generated JSON Schema"
for seed in 0 3; do
    printf 'PYTHONHASHSEED=%s -> ' "$seed"
    PYTHONHASHSEED=$seed "$PYTHON" -c '
from dandischema.models import Dandiset
from dandischema.utils import TransitionalGenerateJsonSchema

schema = Dandiset.model_json_schema(schema_generator=TransitionalGenerateJsonSchema)
print(schema["$defs"]["LicenseType"]["enum"])
'
done
```

```
== LicenseType member order
PYTHONHASHSEED=0 -> ['spdx:CC-BY-4.0', 'spdx:CC0-1.0']
PYTHONHASHSEED=1 -> ['spdx:CC-BY-4.0', 'spdx:CC0-1.0']
PYTHONHASHSEED=2 -> ['spdx:CC-BY-4.0', 'spdx:CC0-1.0']
PYTHONHASHSEED=3 -> ['spdx:CC0-1.0', 'spdx:CC-BY-4.0']
PYTHONHASHSEED=4 -> ['spdx:CC0-1.0', 'spdx:CC-BY-4.0']
PYTHONHASHSEED=5 -> ['spdx:CC-BY-4.0', 'spdx:CC0-1.0']

== the same ordering in the generated JSON Schema
PYTHONHASHSEED=0 -> ['spdx:CC-BY-4.0', 'spdx:CC0-1.0']
PYTHONHASHSEED=3 -> ['spdx:CC0-1.0', 'spdx:CC-BY-4.0']
```

Without a pinned seed the outcome is simply random: six consecutive unseeded runs gave `['spdx:CC0-1.0', 'spdx:CC-BY-4.0']` five times and the reverse once.

</details>

## Impact

Validation is unaffected, since `enum` is order-insensitive and nothing in the code depends on the member order. The problem is that the build artifact is not reproducible:

- The `enum` array order in the files published to [dandi/schema](https://github.com/dandi/schema) is whatever the release run happened to roll. `releases/0.7.0` currently has `["spdx:CC0-1.0", "spdx:CC-BY-4.0"]` purely by chance.
- The "Test for unversioned changes" step of `.github/workflows/release.yml` regenerates the current version's directory and fails if it differs from what is already published. When a release does not bump `DANDI_SCHEMA_VERSION`, that gate can fail (or pass) on nothing but the hash seed.
- Any before/after schema comparison, such as the one done for #419, has to pin `PYTHONHASHSEED` to keep the noise out of the diff.

## Suggested fix

Sort at the point where the order becomes observable, which leaves `Config.licenses` a `set` (the right type for a configuration value that should be deduplicated and compared order-insensitively):

```python
LicenseType = Enum(
    "LicenseType",
    [
        (license_.name, license_.value)
        for license_ in sorted(_INSTANCE_CONFIG.licenses, key=lambda lic: lic.value)
    ],
)
```

This settles on `["spdx:CC-BY-4.0", "spdx:CC0-1.0"]`, which differs from what `releases/0.7.0` currently contains, so the resulting change to the published files should ride along with the next `DANDI_SCHEMA_VERSION` bump rather than silently rewriting an existing release directory.

A regression test is awkward within one process, since the enum is built at import time; running the reproducer under two pinned seeds in a subprocess is the straightforward way to cover it.

candleindark:
Fixes #433.

## Problem

`LicenseType` is built by iterating `Config.licenses`, which is a `set`. The order of iterating a `set` of `Enum` members depends on `PYTHONHASHSEED`, which CPython randomizes per process, so the member order of `LicenseType` varied from process to process.

That member order is what fixes the order of the elements in the `enum` array of the `LicenseType` definition under `$defs`, which is emitted into all four generated JSON Schemas (`dandiset.json`, `published-dandiset.json`, `asset.json`, `published-asset.json`). Two invocations of `tools/pubschemata.py` on the same commit could therefore emit different files.

## Fix

Sort the members by value when building the enum. `Config.licenses` stays a `set`, with the sort applied at the single point where the order becomes observable. Validation behavior is unchanged, since JSON Schema `enum` is order-insensitive.

## Merge order

Recommend merging to master before #419 is released. This PR carries no `DANDI_SCHEMA_VERSION` bump, since #419 already bumps it to `0.8.0`. The fix settles the order on `["spdx:CC-BY-4.0", "spdx:CC0-1.0"]`, which differs from what `dandi/schema` `releases/0.7.0/` holds, so merging ahead of #419's release lets `0.8.0` be published once with the sorted order.

## Test plan

- [x] `LicenseType` member order is identical across `PYTHONHASHSEED` in `{0, 1, 2, 3, 4, 5, 42, 1234}`
- [x] `tools/pubschemata.py` output is byte-identical between seeds 0 and 3; before the fix, all four schema files differed
- [x] `pytest dandischema`: 258 passed, 19 skipped
- [x] `pre-commit` and `mypy` clean

### DIFF
--- a/dandischema/models.py
+++ b/dandischema/models.py
@@ -120,9 +120,13 @@ if TYPE_CHECKING:
         ...  # fmt: skip
 
 else:
+    # `sorted()` makes the member order, and hence the JSON schema, consistent.
     LicenseType = Enum(
         "LicenseType",
-        [(license_.name, license_.value) for license_ in _INSTANCE_CONFIG.licenses],
+        [
+            (license_.name, license_.value)
+            for license_ in sorted(_INSTANCE_CONFIG.licenses, key=lambda lic: lic.value)
+        ],
     )
     r"""
     An enumeration of supported licenses


dandischema/models.py
```diff
@@ -120,9 +120,18 @@ class LicenseType(Enum):
         ...  # fmt: skip
 
 else:
+    # The `sorted()` call is required, not cosmetic. `Config.licenses` is a `set`,
+    # and the order of iterating a `set` of `Enum` members depends on
+    # `PYTHONHASHSEED`, which CPython randomizes per process. Without sorting, the
+    # member order, and hence the `enum` array of every JSON schema generated from
+    # these models, varies from run to run. Pydantic emits member values rather
+    # than names into the `enum` array, which is why `value` is the sort key.
```
yarikoptic:
```suggestion
    # `sorted()` since `Config.licenses` is a `set`.
```
is likely the absolute maximum needed here, the rest is too AIsloppy to my taste, or am I missing some detail which  might be crucial if omitted?

candleindark: I wanted to point out why ordering is needed here not that `Config.licenses` is unordered. I agree that the original one is too wordy. I have shorten it to just one line. Additionally, I have removed the change to the following docstring.

Please take another look.

